PR #10, opened 2025-04-03
Insert Quantize Op On Dtype Mismatch

jinevening:
This inserts Quantize Op into the place where dtype of input and node mismatch.

---

### DIFF
--- a/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
+++ b/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
@@ -1,0 +1,288 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torch.fx
+import copy
+
+import torch
+from torch.export import ExportedProgram
+
+from tico.serialize.quant_param import QPARAM_KEY, QuantParam
+from tico.utils import logging
+from tico.utils.errors import NotYetSupportedError
+from tico.utils.passes import PassBase, PassResult
+from tico.utils.trace_decorators import trace_graph_diff_on_pass
+from tico.utils.utils import set_new_meta_val
+from tico.utils.validate_args_kwargs import (
+    BmmArgs,
+    LinearArgs,
+    MulTensorArgs,
+    PermuteArgs,
+)
+
+
+def _dtype(node: torch.fx.Node) -> str:
+    assert QPARAM_KEY in node.meta
+    return node.meta[QPARAM_KEY].dtype
+
+
+# Convert i16 qparam to u8 qparam
+# scale and zero_point are inferred from i16 qparam
+def _i16_to_u8(qparam: QuantParam) -> QuantParam:
+    # Assume per-tensor quantization
+    assert qparam.scale is not None and len(qparam.scale) == 1
+    assert qparam.dtype == "int16"
+
+    s16_scale = qparam.scale[0]
+    max_ = s16_scale * 32767  # numeric_limits<int16>
+    min_ = -max_
+
+    u8_scale = (max_ - min_) / 255
+    u8_zerop = round(-min_ / u8_scale)
+
+    new_qparam = QuantParam()
+    new_qparam.scale = [u8_scale]
+    new_qparam.zero_point = [u8_zerop]
+    new_qparam.dtype = "uint8"
+
+    return new_qparam
+
+
+# Convert u8 qparam to i16 qparam
+# scale is inferred from u8 qparam
+def _u8_to_i16(qparam: QuantParam) -> QuantParam:
+    # Assume per-tensor quantization
+    assert qparam.scale is not None and len(qparam.scale) == 1
+    assert qparam.zero_point is not None and len(qparam.zero_point) == 1
+    assert qparam.dtype == "uint8"
+
+    u8_scale = qparam.scale[0]
+    u8_zerop = qparam.zero_point[0]
+    max_ = u8_scale * (255 - u8_zerop)
+    min_ = u8_scale * (-u8_zerop)
+
+    abs_max = max([max_, min_], key=abs)
+    s16_scale = abs_max / 32767
+    s16_zerop = 0
+
+    new_qparam = QuantParam()
+    new_qparam.scale = [s16_scale]
+    new_qparam.zero_point = [s16_zerop]
+    new_qparam.dtype = "int16"
+
+    return new_qparam
+
+
+@trace_graph_diff_on_pass
+class InsertQuantizeOnDtypeMismatch(PassBase):
+    """
+    Insert quantize Op in the operators where circle's type inference is violated.
+    Example. FullyConnected
+    [BEFORE]
+      Op (uint8) -  aten.linear.default (int16)
+    [AFTER]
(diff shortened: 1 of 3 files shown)


tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+@trace_graph_diff_on_pass
+class InsertQuantizeOnDtypeMismatch(PassBase):
+    """
+    Insert quantize Op in the operators where circle's type inference is violated.
+    Example. FullyConnected
+    [BEFORE]
+      Op (uint8) -  aten.linear.default (int16)
+    [AFTER]
+      Op (uint8) -  aten.linear.default (uint8) - quantized_decomposed.quantize_per_tensor.default (int16)
```
mhs4670go, line 98: @jinevening 

I'm not sure but I imaged like below change. Because linear output data type is changed otherwise. Is it okay?

```python
Op (uint8) - quantized_decomposed.quantize_per_tensor.default (int16) - aten.linear.default (int16) 
```

jinevening: It looks like output dtype is changed, but our backend compiler will fuse linear + quantize if possible. So, output dtype can be preserved.

Acutally Op's precision is mainly determined by input dtype. So, it woud be better to keep input dtype unmodified.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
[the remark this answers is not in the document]
jinevening: Hmm.. I didn't catch the point. How can we enable mixed-precision Ops by skipping folding?

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
mhs4670go: Thanks for the explanation. 

I'm not against the introducing pass. But, based on my understanding, I'll summarize the situation for others who will review it. Let me know if I got anything wrong.

The issue arises after applying fake quantization and folding the Q-DQ pairs on the model. In our current flow, each op is converted to a quantized dtype before passing the graph to the backend compiler, which requires all ops to be quantized. For instance, consider a module with two Linear ops:

#### Original
```
Linear(f32) - Linear(f32)
```

#### After fake quantization (FQ)
```
Linear(f32) - Q(uint8) - DQ(f32)  - Linear(f32) - Q(uint8) - DQ(f32)
```

#### After folding
```
Linear(uint8) - Linear(uint8)
```

However, the backend compiler also includes a Linear instruction with a uint8 input and an int16 output. To represent this during fake quantization, we end up with:

#### Initial case
```
Linear(f32) - Linear(f32)
```

#### After FQ
```
Linear(f32) - Q(uint8) - DQ(f32)  - Linear(f32) - Q(int16) - DQ(f32)
```

#### After folding
```
Linear(uint8) - Linear(int16)

(After applyting this pass: Linear(uint8) - Linear(uint8) - Q(int16))
```

This pattern lets the backend compiler convert the Linear(uint8)-Q(int16) segment into a linear_I8_I16 instruction. However, the problem is that after folding the Q-DQ pairs, **the dtype change from uint8 to int16 becomes implicit**, and the information about the type transition is lost.

One possible workaround is a this pass and the other one that I come up with is inserting a Q between the ops (i.e., Linear(uint8) - Q(int16) - Linear(int16)), but that approach fails to clearly distinguish between a genuine linear_I8_I16 instruction and a regular linear_I16 op. I mean current approach seems to be too a single target-specific soulution.

To resolve this, I suggest one potential approach for future.

#### Preserve quantization metadata by explicitly annotating ops with input and output dtype

Even after folding the Q-DQ pairs, retain the quantized dtype information as metadata or annotations on each op. For example, annotate the first Linear op as having a uint8 output and the second Linear op as expecting an int16 input. The backend compiler can then use this metadata to correctly identify the conversion pattern and transform it into a linear_I8_I16 instruction. I guess we need to extend a circle schema for this.

Solutions should aim to maintain explicit information about the quantization type transition between ops so that the backend compiler - including other compilers - can reliably detect and process the graph.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -0,0 +1,288 @@
...
+from tico.utils.utils import set_new_meta_val
+from tico.utils.validate_args_kwargs import (
+    BmmArgs,
+    LinearArgs,
+    MulTensorArgs,
+    PermuteArgs,
+)
+
+
+def _dtype(node: torch.fx.Node) -> str:
```
mhs4670go, line 38: (optional) The name of the current API could be misleading because it might be interpreted as returning the data type of the node's output tensor. Therefore, seems to me that, to make this clearer, a more descriptive name like `qparam_dtype()` or `get_qparam_dtype()` would be better.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -0,0 +1,288 @@
...
+
+# Convert i16 qparam to u8 qparam
+# scale and zero_point are inferred from i16 qparam
+def _i16_to_u8(qparam: QuantParam) -> QuantParam:
+    # Assume per-tensor quantization
+    assert qparam.scale is not None and len(qparam.scale) == 1
+    assert qparam.dtype == "int16"
+
+    s16_scale = qparam.scale[0]
+    max_ = s16_scale * 32767  # numeric_limits<int16>
```
mhs4670go, line 51: (Just for note) We can apply this change when we introduce `qmax` and `qmin` in the future.

```suggestion
    max_ = s16_scale * qparam.qmax  # numeric_limits<int16>
```

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -0,0 +1,288 @@
...
+    u8_scale = qparam.scale[0]
+    u8_zerop = qparam.zero_point[0]
+    max_ = u8_scale * (255 - u8_zerop)
+    min_ = u8_scale * (-u8_zerop)
+
+    abs_max = max([max_, min_], key=abs)
+    s16_scale = abs_max / 32767
+    s16_zerop = 0
+
+    new_qparam = QuantParam()
```
mhs4670go, line 82: `.min` and `.max` also should be saved.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -0,0 +1,288 @@
...
+                if QPARAM_KEY not in node.meta:
+                    continue
+
+                if _dtype(inp) == _dtype(node):
+                    continue
+
+                if _dtype(inp) == "uint8" and _dtype(node) == "int16":
+                    quantize = _insert_quantize_op_after(node)
+
+                    quantize.meta[QPARAM_KEY] = copy.deepcopy(node.meta[QPARAM_KEY])
```
mhs4670go, line 178:
```suggestion
                    quantize.meta[QPARAM_KEY] = node.meta[QPARAM_KEY]
```

mhs4670go (approved):
LGTM with some minor comments.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: Thanks for summarization :) As you pointed out, the problem stems from the lack of expression in circle schema. I think extending circle schema is also a feasible solution. One concern is that it would require significant changes across our toolchain.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: Is it necessary? Typically they are not used.

FYI one-quantize removes min/max by default to reduce model size.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: This can cause sharing of qparam. Copy would be safer.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: I renamed to `qparam_dtype`.

mhs4670go (approved):
LGTM

PR #11, opened 2025-04-03
Support more constant scales and zerop

jinevening:
This supports more constant scales and zeop other than named_buffers.

---

### DIFF
--- a/tico/experimental/quantization/passes/remove_weight_dequant_op.py
+++ b/tico/experimental/quantization/passes/remove_weight_dequant_op.py
@@ -30,11 +30,16 @@ from tico.utils.validate_args_kwargs import (
 )
 
 
-def get_data_from_buffers(exported_program: ExportedProgram, node: torch.fx.Node):
+def get_constant(exported_program: ExportedProgram, node: torch.fx.Node):
     assert isinstance(node, torch.fx.Node)
-    named_buffers = dict(exported_program.named_buffers())
-    buffer_name = exported_program.graph_signature.inputs_to_buffers[node.name]
-    return named_buffers[buffer_name]
+    if node.name in exported_program.constants:
+        return exported_program.constants[node.name]
+    elif node.name in exported_program.graph_signature.inputs_to_buffers:
+        buffer_name = exported_program.graph_signature.inputs_to_buffers[node.name]
+        named_buffer = dict(exported_program.named_buffers())
+        return named_buffer[buffer_name]
+    else:
+        raise RuntimeError("NYI constant")
 
 
 @trace_graph_diff_on_pass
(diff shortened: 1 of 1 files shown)


batcheu (commented):
LGTM

batcheu (approved):

tico/experimental/quantization/passes/remove_weight_dequant_op.py
```diff
@@ -30,11 +30,16 @@
...
-    buffer_name = exported_program.graph_signature.inputs_to_buffers[node.name]
-    return named_buffers[buffer_name]
+    if node.name in exported_program.constants:
+        return exported_program.constants[node.name]
+    elif node.name in exported_program.graph_signature.inputs_to_buffers:
+        buffer_name = exported_program.graph_signature.inputs_to_buffers[node.name]
+        named_buffer = dict(exported_program.named_buffers())
+        return named_buffer[buffer_name]
+    else:
+        raise RuntimeError("NYI constant")
```
miusic, line 42:
```suggestion
        raise RuntimeError(f"Cannot find following constant node: {node.name}")
```

(optional)
How about this?

miusic (approved):
LGTM w/ minor comment :) https://github.com/Samsung/TICO/pull/11#discussion_r2026494351

tico/experimental/quantization/passes/remove_weight_dequant_op.py
```diff
@@ -30,11 +30,16 @@
 )
 
 
-def get_data_from_buffers(exported_program: ExportedProgram, node: torch.fx.Node):
+def get_constant(exported_program: ExportedProgram, node: torch.fx.Node):
```
mhs4670go, line 33: For the reviewers, here’s a summary of where `scale` and `zero_point` typically end up in an `ExportedProgram` depending on how they are defined and used.

- If they are hardcoded in the code (e.g., written directly in a quantize function call), and not registered as parameters or buffers, they are usually treated as constants and placed in the `.constants` dictionary. I guess **current PR tries to cover this case**.
- If they are registered using `register_buffer()`, which is common in FX-based quantization workflows, they will appear in `inputs_to_buffers` and are treated as model state (non-trainable but persistent). Which is why I implemented this function the way I implemented in the first place.

```python
# torch/ao/quantization/fake_quantize.py
self.register_buffer("scale", torch.tensor([1.0], dtype=torch.float))
self.register_buffer("zero_point", torch.tensor([0], dtype=zero_point_dtype))
```
https://github.com/pytorch/pytorch/blob/9e106019f64d668f17f0b50dc46192cff7a37dce/torch/ao/quantization/fake_quantize.py#L204-L205

- If they are passed in as inputs to the model, they may appear in `inputs_to_lifted_tensor_constants`, especially when the export logic lifts them from the graph to expose them as inputs. Maybe we need to cover this in the future.

In practice, for quantized models generated via `prepare_fx` / `convert_fx`, `scale` and `zero_point` are most often stored as buffers, and therefore show up in `inputs_to_buffers`.

mhs4670go (approved):
LGTM

Seunghui98 (approved):
LGTM 👍

tico/experimental/quantization/passes/remove_weight_dequant_op.py
jinevening: Thanks for explaining the background! External quantization frameworks define scale/zero_point in various ways, so I had to extend this function (maybe updated again when we support another framework).

Issue #15, opened 2025-04-04
Support uint4 weights


jinevening:
## What

Let's support uint4 weights

## Why

They are widely used in LLMs.

jinevening:
Done.

PR #16, opened 2025-04-04
Support uint4 weights

jinevening:
This supports export of uint4 weights.

---
https://github.com/Samsung/TICO/issues/15

### DIFF
--- a/tico/experimental/quantization/passes/remove_weight_dequant_op.py
+++ b/tico/experimental/quantization/passes/remove_weight_dequant_op.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Optional, TYPE_CHECKING
+from typing import List, Optional, TYPE_CHECKING, Union
 
 if TYPE_CHECKING:
     import torch.fx
@@ -40,6 +40,32 @@ def get_constant(exported_program: ExportedProgram, node: torch.fx.Node):
         return named_buffer[buffer_name]
     else:
         raise RuntimeError("NYI constant")
+
+
+class ValRange:
+    def __init__(self, val: Union[torch.Tensor, List[int]]):
+        if isinstance(val, torch.Tensor):
+            self.max = torch.max(val).item()
+            self.min = torch.min(val).item()
+        elif type(val) == list:
+            self.max = max(val)
+            self.min = min(val)
+        else:
+            raise RuntimeError("Wrong dtype (val)")
+
+    def within(self, min_val, max_val):
+        return self.min >= min_val and self.max <= max_val
+
+
+# Infer dtype using weight, zero point, and dtype
+def infer_dtype(weight: torch.Tensor, zerop: List[int], dtype: torch.dtype) -> str:
+    weight_val = ValRange(weight)
+    zp_val = ValRange(zerop)
+
+    if weight_val.within(0, 15) and zp_val.within(0, 15) and dtype == torch.uint8:
+        return "uint4"
+    else:
+        return to_qparam_dtype(dtype)
 
 
 @trace_graph_diff_on_pass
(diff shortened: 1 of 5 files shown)


tico/serialize/pack.py
mhs4670go, line 1: License header should be added.

tico/experimental/quantization/passes/remove_weight_dequant_op.py
```diff
@@ -42,6 +42,32 @@ def get_constant(exported_program: ExportedProgram, node: torch.fx.Node):
...
+            self.min = min(val)
+        else:
+            raise RuntimeError("Wrong dtype (val)")
+
+    def within(self, min_val, max_val):
+        return self.min >= min_val and self.max <= max_val
+
+
+# Infer dtype using weight, zero point, and dtype
+def infer_dtype(weight: torch.Tensor, zerop: List[int], dtype: torch.dtype) -> str:
```
mhs4670go, line 61: (Just for note) For describing other approaches that we could do.

Range check alone may not be sufficient to determine whether a quantized tensor is truly intended for uint4. Here are a few additional checks and best practices that can help ensure correctness later.

1. Verify qmin and qmax

Explicitly confirm that the quantization parameters use qmin=0 and qmax=15. This is a strong indicator that the quantization was configured for 4-bit values. If the range is [0, 255], it’s clearly an 8-bit quantization, regardless of the actual values.

2. Attach explicit metadata after `torch.export.export`

After exporting a model using `torch.export.export`, we can directly insert a custom field like `quant_info` into the `ExportedProgram.metadata`. This allows us to retain quantization intent (e.g., bit-width) and avoids relying on heuristics downstream. But, this method should know the name of the exported names to map from an original tensor to exported one.

mhs4670go (approved):
LGTM

tico/serialize/pack.py
jinevening: Thanks for finding!

tico/experimental/quantization/passes/remove_weight_dequant_op.py
```diff
@@ -101,23 +127,32 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
             if QPARAM_KEY in q_weight.meta:
                 continue
 
-            q_weight_val = q_weight.meta["val"]
-            assert isinstance(q_weight_val, FakeTensor)
+            q_weight_meta = q_weight.meta["val"]
+            assert isinstance(q_weight_meta, FakeTensor)
```
jinevening, line 131: I renamed `q_weight_val` to `q_weight_meta`, because I use `q_weight_val` as a variable name for the actual value of weights.

mhs4670go (dismissed):

tico/serialize/pack.py
```diff
@@ -0,0 +1,34 @@
...
+        numel = flat_data.shape[0]
+        packed = np.zeros((numel + 1) // 2, dtype=np.uint8)
+        for i in range(numel):
+            if i % 2 == 0:
+                packed[i // 2] = flat_data[i] & 0x0F
+            else:
+                packed[i // 2] |= flat_data[i] << 4
+        return packed
+    else:
+        raise RuntimeError(f"NYI dtype: {dtype}")
```
miusic, line 34:
```suggestion
        raise NotImplementedError(f"NYI dtype: {dtype}")
```
It looks more proper to use `NotImplementedError` according to the error message

tico/serialize/pack.py
```diff
@@ -0,0 +1,34 @@
...
+            raise RuntimeError("uint4 data should be saved in uint8.")
+
+        numel = flat_data.shape[0]
+        packed = np.zeros((numel + 1) // 2, dtype=np.uint8)
+        for i in range(numel):
+            if i % 2 == 0:
+                packed[i // 2] = flat_data[i] & 0x0F
+            else:
+                packed[i // 2] |= flat_data[i] << 4
+        return packed
```
miusic, line 32:
```suggestion
        for i in range(numel):
            if not ((flat_data[i] >=0) or (flat_data[i] <=255)):
                raise RuntimeError("flat_data[i] is out of range uint4.")
            if i % 2 == 0:
                packed[i // 2] = flat_data[i] & 0x0F
            else:
                packed[i // 2] |= flat_data[i] << 4
        return packed
```
Is it guaranteed that `flat_data[i]` is bounded in the range of `uint4`?
Is there no possibility like that the `flat_data` is out of `uint4` range though the `dtype` is specified as `uint4`?
IMHO, it would be nice to check to ensure that the `flat_data` is in the range of `uint4`

jinevening: Range of weights values is checked in `infer_dtype`. I'll add an assert rather than exception.

tico/serialize/pack.py
jinevening: Updated.

miusic (approved):
LGTM 👍

mhs4670go (approved):
LGTM

PR #17, opened 2025-04-07
Revise quantization steps

jinevening:
This revises quantization steps
- Do quantization after optimization
- Remove FillMetaQuantVal

---

### DIFF
--- a/tico/utils/convert.py
+++ b/tico/utils/convert.py
@@ -20,7 +20,6 @@ import torch
 from torch.export import export, ExportedProgram
 
 from tico.config import CompileConfigBase, get_default_config
-from tico.experimental.quantization.passes.fill_meta_quant_val import FillMetaQuantVal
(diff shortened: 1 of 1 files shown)


jinevening:
To reviewers: Quantization process is changed as below.

As-is
- FoldQuantOps -> Optimize -> Finalize quantization

To-Be
- Optimize -> Quantize -> Post-quantization Optimize (NYI)

Previously, `FoldQuantOps` embdes qparams to nodes, so all passes in `Optimize` stage should have properly handled the qparams when node is replaced/created/modified (even for fp32-targeted optimizations). This makes it difficult to write passes.

Now optimization is done before quantization, thus pass writers do not have to consider qparams.

If we need some more optimizations after quantization, we may introduce a new post-quantization optimization stage in the future.

mhs4670go (approved):
LGTM

Seunghui98 (approved):
LGTM 👍

miusic (approved):
LGTM

PR #24, opened 2025-04-08
Use proper quantization parameters

jinevening:
This includes the following changes.
- quant_min and quant_max are properly set.
- use circle tensor's qparam in quantize_per_tensor.

### DIFF
--- a/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
+++ b/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
@@ -26,7 +26,7 @@ from tico.utils import logging
 from tico.utils.errors import NotYetSupportedError
 from tico.utils.passes import PassBase, PassResult
 from tico.utils.trace_decorators import trace_graph_diff_on_pass
-from tico.utils.utils import set_new_meta_val
+from tico.utils.utils import quant_min_max, set_new_meta_val
 from tico.utils.validate_args_kwargs import (
     BmmArgs,
     LinearArgs,
@@ -119,8 +119,7 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
             assert qparam.zero_point is not None
             scale = qparam.scale[0]
             zerop = qparam.zero_point[0]
-            min_ = qparam.min[0] if qparam.min is not None else 0
-            max_ = qparam.max[0] if qparam.max is not None else 0
+            min_, max_ = quant_min_max(qparam.dtype)
             dtype = getattr(torch, qparam.dtype)
 
             with graph.inserting_before(node):
@@ -129,6 +128,7 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
                     torch.ops.quantized_decomposed.quantize_per_tensor.default,
                     args=q_args,
                 )
+                quantize.meta[QPARAM_KEY] = copy.deepcopy(qparam)
                 set_new_meta_val(quantize)
 
             node.replace_input_with(inp, quantize)
@@ -141,8 +141,7 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
             assert qparam.zero_point is not None
             scale = qparam.scale[0]
             zerop = qparam.zero_point[0]
-            min_ = qparam.min[0] if qparam.min is not None else 0
-            max_ = qparam.max[0] if qparam.max is not None else 0
+            min_, max_ = quant_min_max(qparam.dtype)
             dtype = getattr(torch, qparam.dtype)
             with graph.inserting_after(node):
                 q_args = (node, scale, zerop, min_, max_, dtype)
@@ -153,6 +152,8 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
 
             node.replace_all_uses_with(quantize, propagate_meta=True)
             quantize.replace_input_with(quantize, node)
+
+            quantize.meta[QPARAM_KEY] = copy.deepcopy(qparam)
 
             return quantize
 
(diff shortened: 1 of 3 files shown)


tico/serialize/operators/op_quantize_per_tensor.py
```diff
@@ -48,13 +48,17 @@ def define_node(
...
+
+        # Tensor should have qparam when it's exported
+        # The qparam must match with the arguments of this Op
+        assert output_tensor.quantization.scale[0] == scale
+        assert output_tensor.quantization.zeroPoint[0] == zero_p
+
+        if output_tensor.type == circle.TensorType.TensorType.UINT8:
+            assert quant_min == 0 and quant_max == 255
+        elif output_tensor.type == circle.TensorType.TensorType.INT16:
+            assert quant_min == -32767 and quant_max == 32767
```
jinevening, line 61: I've changed the behavior. Setting the qparam of tensor is not the role of op visitor. Qparam has to be set when the tensor is exported.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -119,8 +119,7 @@ def _insert_quantize_op_before(node, inp):
             assert qparam.zero_point is not None
             scale = qparam.scale[0]
             zerop = qparam.zero_point[0]
-            min_ = qparam.min[0] if qparam.min is not None else 0
-            max_ = qparam.max[0] if qparam.max is not None else 0
+            min_, max_ = quant_min_max(qparam.dtype)
```
jinevening, line 122: Note that min/max of `torch.ops.quantized_decomposed.quantize_per_tensor.default` is quantized dtype's min/max, not the activation's min/max.

tico/utils/utils.py
```diff
@@ -302,3 +302,12 @@ def to_circle_qparam(qparam: QuantParam):
...
     return circle_qparam
+
+
+def quant_min_max(dtype: str):
+    if dtype == "uint8":
+        return (0, 255)
+    elif dtype == "int16":
+        return (-32767, 32767)
+    else:
+        raise NotImplementedError(f"NYI dtype: {dtype}")
```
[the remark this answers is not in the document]
jinevening: As mentioned in https://github.com/Samsung/TICO/pull/24#discussion_r2032258752, min/max of `torch.ops.quantized_decomposed.quantize_per_tensor.default` is quantized dtype's min/max (ex: 0, 255 for uint8 dtype), not the activation's min/max.



The previous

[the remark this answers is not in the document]
Seunghui98: Oh, I understood it!
Thanks for letting me know 👍

Seunghui98 (approved):
LGTM

tico/serialize/operators/op_quantize_per_tensor.py
mhs4670go: You intentionally skipped `else` statement? I just though of where the UINT4 was.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
mhs4670go: Isn't it natural to set min/max of `QuantParam` when it is created instead? And, these codes here wouldn't have to be changed.

tico/serialize/operators/op_quantize_per_tensor.py
jinevening: This is per-tensor quantize Op used for activation quantization. I've not seen u4 activation yet.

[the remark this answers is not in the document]
jinevening: -32767 is used by one of our internal frameworks, but it seems that -32768 is more general.

From onnx website: https://onnx.ai/onnx/operators/onnx__QuantizeLinear.html#summary

We're dealing with both ranges as int16 in `decompose_fake_quantize_tensor_qparams.py`.

```python
def get_quant_type(min: int, max: int) -> torch.dtype:
    if min == 0 and max == 255:
        return torch.uint8
    if min == -32768 and max == 32767:
        return torch.int16
    if min == -32767 and max == 32767:
        return torch.int16

    raise RuntimeError("Not supported min/max values")
```

Let's follow the same way. Thanks for catch!

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: `min_` and `max_` are used for `torch.ops.quantized_decomposed.quantize_per_tensor.default`, not QuantParam.

jinevening: QuantParam's min/max is activation min/max, which is different from quant_min and quant_max.

Seunghui98 (approved):

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
mhs4670go: Ah, I just thought about introducing `quant_min` and `quant_max`, which seems more flexible to me. Of course we've got no case where qmin/qmax and dtype are aligned though. 

Current codes looks good to me as well.

mhs4670go (approved):
LGTM

seockho-kim (approved):
LGTM

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: That could be another design choice :) Maybe we can save `quant_min` and `quant_max` instead of `dtype` in QuantParam. I just used `dtype` because `quant_min` and `quant_max` are not used in circle.

PR #27, opened 2025-04-08
Add Reshape to InsertQuantizeOnDtypeMismatch

jinevening:
This adds Reshape Op to InsertQuantizeOnDtypeMismatch.

### DIFF
--- a/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
+++ b/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
@@ -25,6 +25,7 @@ from test.modules.op.bmm import SimpleBatchMatMul
 from test.modules.op.linear import SimpleLinear
 from test.modules.op.mul import SimpleMulWithTensor
 from test.modules.op.permute import SimplePermute
+from test.modules.op.reshape import ReshapeTorchAPI
 
 
 class InsertQuantizeOnDtypeMismatchTest(unittest.TestCase):
@@ -141,4 +142,14 @@ class MulTest(InsertQuantizeOnDtypeMismatchTest):
 class BMMTest(InsertQuantizeOnDtypeMismatchTest):
     def test_i16o8(self):
         self.setup(SimpleBatchMatMul(), torch.ops.aten.bmm.default, dtype="int16")
+        self.run_test()
+
+
+class ReshapeTest(InsertQuantizeOnDtypeMismatchTest):
+    def test_i8o16(self):
+        self.setup(ReshapeTorchAPI(), torch.ops.aten.reshape.default, dtype="uint8")
+        self.run_test()
+
+    def test_i16o8(self):
+        self.setup(ReshapeTorchAPI(), torch.ops.aten.reshape.default, dtype="int16")
         self.run_test()
--- a/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
+++ b/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
@@ -117,7 +117,7 @@ class InsertQuantizeOnDtypeMismatchTest(unittest.TestCase):
 
     def run_test(self):
         # Before pass
-        # InsertQuantizeOnDtypeMismatch works for the case when input dtype does not match with output dtype
+        # InsertQuantizeOnDtypeMismatch inserts Quantize Op when input dtype does not match with output dtype
         self.assertNotEqual(self.target.meta[QPARAM_KEY].dtype, self.input_dtype)
 
         target_pass = InsertQuantizeOnDtypeMismatch()
@@ -145,6 +145,9 @@ class PermuteTest(InsertQuantizeOnDtypeMismatchTest):
         )
         self.run_test()
 
+    # Quantize Op is inserted before Permute, so desired_dtype of Reshape is uint8
+    # After conversion)
+    # Input (s16) -> Quantize (u8) -> Permute (u8)
     def test_i16o8(self):
         self.setup(
             SimplePermute(),
@@ -198,6 +201,9 @@ class ReshapeTest(InsertQuantizeOnDtypeMismatchTest):
         )
         self.run_test()
 
+    # Quantize Op is inserted before Reshape, so desired_dtype of Reshape is uint8
+    # After conversion)
+    # Input (s16) -> Quantize (u8) -> Reshape (u8)
     def test_i16o8(self):
         self.setup(
             ReshapeTorchAPI(),
--- a/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
+++ b/test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
@@ -33,7 +33,7 @@ class InsertQuantizeOnDtypeMismatchTest(unittest.TestCase):
     This class runs a test for InsertQuantizeOpOnDtypeMismatch pass
     - Set up a network with target_op and dtype of input. Target node's
       dtype is set differently from input dtype
-    - After the pass runs, node's dtype must be the same with its input
+    - After the pass runs, node's dtype must be the same with desired_dtype
 
     How to use?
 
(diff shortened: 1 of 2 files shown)


mhs4670go:
Style-check failed.

jinevening:
Updated the PR. I made https://github.com/Samsung/TICO/issues/36 for automatic formatting.

mhs4670go (approved):
LGTM

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -271,6 +272,39 @@ def _insert_quantize_op_after(node):
...
+
+                if QPARAM_KEY not in node.meta:
+                    continue
+
+                if qparam_dtype(inp) == qparam_dtype(node):
+                    continue
+
+                if qparam_dtype(inp) == "int16" and qparam_dtype(node) == "uint8":
+                    # A new Quantize Op (s16 to u8) is inserted before (not after)
+                    # reshape Op to reduce tensor size ealier
```
dayo09, line 297: I am new to this parts of codes, so could you please explain why int16->uint8 requires `insert quantize op before` and uint8->int16 requires `insert quantize op after` functions?

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -271,6 +272,39 @@ def _insert_quantize_op_after(node):
...
+                    quantize = _insert_quantize_op_before(node, inp)
+
+                    quantize.meta[QPARAM_KEY] = copy.deepcopy(node.meta[QPARAM_KEY])
+                    node.meta[QPARAM_KEY] = _u8_to_i16(node.meta[QPARAM_KEY])
```
miusic, line 301: May I ask some more details for the reason why the dtype of node was changed to `s16`?

IIUC, isn't it enough to insert `quantize` and not to change the dtype of the node

https://github.com/Samsung/TICO/blob/f0aeefcd34ae306392f37e8525070c792f8dddd2/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py#L99-L104

miusic: Based on my understanding, below is the graph before this pass and after this pass

### Before
```
Input(S16) - Reshape(U8)
```

### After
```
Input(S16) - Quantize(U8) - Reshape(S16)
```

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: Let me explain the rationale. For int16->uint8,

1. Insert Quantize before Reshape.

```
Predecessor (int16)-> Quantize (uint8) -> Reshape (uint8) -> ...
```

2. Insert Quantize after Reshape.

```
Predecessor (int16)-> Reshape (int16) -> Quantize (uint8) -> ...
```

Comparing 1) and 2), the difference is that Reshape operation is conducted in uint8 or int16.  I decided to go with 1), which does Reshape in uint8, for faster execution of Reshape. Note that Reshape Op does not change the value, so its dytpe does not affect accuracy.

uint8->int16 case is the opposite of the above example.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
jinevening: Nice catch! I've updated the code not to update Reshape's qparam. Permute was working incorrectly, so I updated it too.

test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
```diff
@@ -107,38 +117,92 @@ def setup(self, mod: torch.nn.Module, target_op, scale=1.0, zp=0, dtype="uint8")
 
     def run_test(self):
         # Before pass
-        self.assertNotEqual(self.target.meta[QPARAM_KEY].dtype, self.dtype)
+        # InsertQuantizeOnDtypeMismatch works for the case when input dtype does not match with output dtype
```
dayo09, line 120:
```suggestion
        # InsertQuantizeOnDtypeMismatchTest works for the case when input dtype does not match with output dtype
```

(optional) The pass `InsertQuantizeOnDtypeMismatch` works on all cases, it's `InsertQuantizeOnDtypeMismatchTest` that works for mismatch case. So to be the comment technically right.

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
dayo09: @jinevening Thank you for the clear explanation :-D!!

dayo09 (dismissed):
LGTM

miusic (dismissed):
LGTM! 👍

test/quantization/pass/test_insert_quantize_on_dtype_mismatch.py
```diff
@@ -107,38 +117,92 @@ def setup(self, mod: torch.nn.Module, target_op, scale=1.0, zp=0, dtype="uint8")
...
 
 class PermuteTest(InsertQuantizeOnDtypeMismatchTest):
     def test_i8o16(self):
-        self.setup(SimplePermute(), torch.ops.aten.permute.default, dtype="uint8")
+        self.setup(
+            SimplePermute(),
+            torch.ops.aten.permute.default,
+            input_dtype="uint8",
+            desired_dtype="uint8",
+        )
```
seockho-kim, line 145: I have a question.
What the meaning of test_i8o16? input 8bit, output 16bit.
Then, desired_dtype "uint8" looks a little weird to me.

dayo09: IIUC, it's `precedence(uint8) - permute(int16)` case. Thus, it must be converted as `precedence(uint8) - permute(uint8, which is changed from int16) - quantize(int16 -> uint8, thus its qparam is int16) `.

desired_dtype means its target node's desired dtype. In this example, target node is permute.
Therefore desired_dtype is uint8.

jinevening: @dayo09 Thanks for explanation :) It seems that the code is a bit difficult to understand. I added comments in the pass code

```python
    - NOTE For some cases, Quantize Op is inserted before the operators.

    Let's assume Reshape Op's input is int16 and output is uint8. There are two possible places to insert
    Quantize Op.

    1. Insert Quantize before Reshape.

    ```
    Predecessor (int16)-> Quantize (uint8) -> Reshape (uint8) -> ...
    ```

    2. Insert Quantize after Reshape.

    ```
    Predecessor (int16)-> Reshape (int16) -> Quantize (uint8) -> ...
    ```

    Comparing 1) and 2), the difference is that Reshape operation is conducted in uint8 or int16.
    We go with 1), which does Reshape in uint8, for faster execution. Note that Reshape Op does not
    change the value, so its dytpe does not affect accuracy.
```

and in the test code.

```python
    # Quantize Op is inserted before Permute, so desired_dtype of Reshape is uint8
    # After conversion)
    # Input (s16) -> Quantize (u8) -> Permute (u8)
    def test_i16o8(self):
        self.setup(
            SimplePermute(),
            torch.ops.aten.permute.default,
            input_dtype="int16",
            desired_dtype="uint8",
        )
        self.run_test()
```

dayo09 (approved):
LGTM

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -102,6 +103,26 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
...
+
+    2. Insert Quantize after Reshape.
+
+    ```
+    Predecessor (int16)-> Reshape (int16) -> Quantize (uint8) -> ...
+    ```
+
+    Comparing 1) and 2), the difference is that Reshape operation is conducted in uint8 or int16.
+    We go with 1), which does Reshape in uint8, for faster execution. Note that Reshape Op does not
+    change the value, so its dytpe does not affect accuracy.
```
dayo09, line 125:
```suggestion
    change the value, so its dtype does not affect accuracy.
```

miusic (approved):

mhs4670go (approved):
LGTM

Issue #28, opened 2025-04-08
Support dynamic shape


zetwhite:
`torch.onnx.export` supports exporting ONNX models with dynamic input/output shapes using `dynamic_axes` option. 

<details>

<summary> `torch.onnx.export` with `dynamic_axes` example code </summary> 

  ```python 
  import torch
  import torch.nn as nn
  
  # Define a simple model
  class SimpleModel(nn.Module):
      def __init__(self):
          super(SimpleModel, self).__init__()
          self.linear = nn.Linear(2, 2)
  
      def forward(self, x):
          return self.linear(x)
  
  # Instantiate and set model to evaluation mode
  model = SimpleModel()
  model.eval()
  
  # Create dummy input
  dummy_input = torch.randn(1, 2)
  
  # Export with dynamic batch size
  torch.onnx.export(
      model,
      dummy_input,
      "simple_model_dynamic.onnx",
      input_names=["input"],
      output_names=["output"],
      dynamic_axes={
          "input": {0: "batch_size"},   # THIS IS A dict, not a set
          "output": {0: "batch_size"}
      },
  )
  
  print("Dynamic ONNX model exported successfully!")
  ``` 

</details> 

As far as I know, **TICO** doesn't support exporting a model with dynamic signatures. 
Is it possible to export to circle model with dynamic shapes?  
I don't know the insides of TICO, so I'd like to know how hard to implement this feature. 

### background 

* When running LLM models on CPU, exporting with a dynamic sequence length would be very useful. ( ONERT also supports circle with dynamic shapes. )

mhs4670go:
I implemented [a simple draft](https://github.com/Samsung/TICO/pull/29) for this feature. Could you use this draft and let me know everything is okay?

One thing that is different from onnx's one is the interface. I followed [torch.export dynamism](https://pytorch.org/docs/stable/export.html#expressing-dynamism).

In short, you should use `torch.export.Dim()` for expressing dynamic feature and you don't have to set outputs because the shapes of the outputs are automatically inferred.

```python
import torch
import torch.nn as nn

import tico

# Define a simple model
class SimpleModel(nn.Module):
    def __init__(self):
        super().__init__()
        self.linear = nn.Linear(2, 3)

    def forward(self, x):
        return self.linear(x)

# Instantiate the model
model = SimpleModel()

# Set model to evaluation mode
model.eval()

# Create a dummy input tensor (shape must match model input)
[...]

PR #29, opened 2025-04-08
[DRAFT] Support dynamic shapes

mhs4670go:
This draft supports dynamic shapes.

Related: #28 

### DIFF
--- a/tico/serialize/circle_graph.py
+++ b/tico/serialize/circle_graph.py
@@ -144,7 +144,9 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         tensor.name = self._gen_unique_name_with_prefix(node.name)
         assert node.meta.get("val") is not None
         tensor.type = extract_circle_dtype(node)
-        tensor.shape = list(extract_shape(node))
+        tensor_shape = list(extract_shape(node))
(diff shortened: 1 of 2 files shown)


jinevening:
image

tico/passes/remove_redundant_reshape.py
```diff
@@ -348,7 +348,7 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
             assert isinstance(reshape1_input, torch.fx.Node), type(reshape1_input)
             assert isinstance(size, list), type(size)
             for s in size:
-                assert isinstance(s, int), type(s)
+                assert isinstance(s, (int, torch.SymInt)), type(s)
```
jinevening, line 351: Most of the current passes are written without consideration of dynamic shapes. We may need more tests to support dynamic shapes in existing passes.

mhs4670go: Ah, current draft is just a quick patch for checking a feasibility of #28. I have a plan that introduces a dynamic shape tests using onert backend internally.

jinevening: Just curiosity. Does onert run well on x86?

zetwhite: > Just curiosity. Does onert run well on x86?

Yes, onert runs well.

Issue #40, opened 2025-04-10
Revise the WIKI page for DCO


miusic:
## What

Let's rename the title of WIKI page about DCO
(https://github.com/Samsung/TICO/wiki/TICO-Developer's-Certificate-of-Origin)

+) The above link is changed to below one now 😄
(https://github.com/Samsung/TICO/wiki/TICO-Developer-Certificate-of-Origin)

#### 1.

When I click the link which points the WIKI page for DCO in following code, it failed to redirect to the proper web page

https://github.com/Samsung/TICO/blob/f0aeefcd34ae306392f37e8525070c792f8dddd2/.github/workflows/check-pr.yaml#L35

In github page or code editor (like VSCode), the link was splitted by the quotation mark in the full link,
so it was perceived as a wrong address (https://github.com/Samsung/TICO/wiki/TICO-Developer)

**Ref**
![Image](https://github.com/user-attachments/assets/21072ab5-d715-45e9-af56-b2f97d2aa076)

#### 2.

It is not that critical, but when I search `Developer's Certificate of Origin`, the most of the results says about `Developer Certificate of Origin` 😅

## How

- [x] Change the name of WIKI page to `Developer Certificate of Origin`
- [x] Submit a PR for change the link
#43

miusic:
Done with #43 :)

PR #42, opened 2025-04-10
[CI] Fix broken link in comment

miusic:
This commit fixes broken link which redirects to the WIKI page about DCO from the `check-pr` workflow

TICO-DCO-1.0-Signed-off-by: miusic <jeyjey6@naver.com>

---
related to: #40


miusic:
Close this PR because of wrong Signed-off 😓

Issue #46, opened 2025-04-11
Formatting only diff


jinevening:
## What

Let's introduce --diff-only option to `./ccex format` command.

## Why

`./ccex format` always runs format check for all files, taking a long time. This will be worsend as our repo grows.

Let's make an option that only checks the files changed from `main` branch.

jinevening:
Done

PR #47, opened 2025-04-11
Format only diff from main

jinevening:
This adds --diff-only option to ./ccex format.

---
Related to: https://github.com/Samsung/TICO/issues/46

### DIFF
--- a/infra/style/format.sh
+++ b/infra/style/format.sh
@@ -27,8 +27,10 @@ if [[ "${CHECK_DIFF_ONLY}" = "1" ]]; then
   DIFF_COMMITS=`git log --graph --oneline main..HEAD | wc -l`
   if [[ -z "${MAIN_EXIST}" ]]; then
     echo "Cannot find main branch"
+    exit 1
   elif [[ "${CURRENT_BRANCH}" = "main" ]]; then
     echo "Current branch is main"
+    exit 1
   else
     # Gather diff from HEAD
     FILES_TO_CHECK=$(git diff --name-only --diff-filter=d HEAD~${DIFF_COMMITS})
--- a/infra/style/format.sh
+++ b/infra/style/format.sh
@@ -6,13 +6,42 @@ set -e
 APPLY_PATCH_OPTION="-a"
 while [[ "$#" -gt 0 ]]; do
   case $1 in
-    --no-apply-patches) APPLY_PATCH_OPTION="";;
+    --no-apply-patches)
+      APPLY_PATCH_OPTION=""
+      shift
+      ;;
+    --diff-only)
+      CHECK_DIFF_ONLY="1"
+      shift
+      ;;
     *) echo "[ERROR] Unknown parameter passed: $1"; exit 255 ;;
   esac
-  shift
 done
 
 GIT_ROOT="$(git rev-parse --show-toplevel)"
 STYLE_DIR="${GIT_ROOT}/infra/style"
 
-lintrunner --force-color --all-files $APPLY_PATCH_OPTION --config "${GIT_ROOT}/.lintrunner.toml"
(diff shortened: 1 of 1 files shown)


infra/style/format.sh
```diff
@@ -6,13 +6,42 @@ set -e
...
 GIT_ROOT="$(git rev-parse --show-toplevel)"
 STYLE_DIR="${GIT_ROOT}/infra/style"
 
-lintrunner --force-color --all-files $APPLY_PATCH_OPTION --config "${GIT_ROOT}/.lintrunner.toml"
+if [[ "${CHECK_DIFF_ONLY}" = "1" ]]; then
+  MAIN_EXIST=$(git rev-parse --verify main)
+  CURRENT_BRANCH=$(git branch | grep \* | cut -d ' ' -f2-)
+  DIFF_COMMITS=`git log --graph --oneline main..HEAD | wc -l`
+  if [[ -z "${MAIN_EXIST}" ]]; then
+    echo "Cannot find main branch"
```
shs-park, line 29: Question, 
is it okay to not exit when the `main` branch does not exist?

```suggestion
    echo "Cannot find main branch"
    exit 1
```

shs-park (commented):
I left a question,
PTAL
=)

infra/style/format.sh
```diff
@@ -6,13 +6,42 @@ set -e
...
 done
 
 GIT_ROOT="$(git rev-parse --show-toplevel)"
 STYLE_DIR="${GIT_ROOT}/infra/style"
 
-lintrunner --force-color --all-files $APPLY_PATCH_OPTION --config "${GIT_ROOT}/.lintrunner.toml"
+if [[ "${CHECK_DIFF_ONLY}" = "1" ]]; then
+  MAIN_EXIST=$(git rev-parse --verify main)
+  CURRENT_BRANCH=$(git branch | grep \* | cut -d ' ' -f2-)
+  DIFF_COMMITS=`git log --graph --oneline main..HEAD | wc -l`
```
dayo09, line 27: When I run this with `./ccex format --diff-only`, I get this error.

```sh
fatal: ambiguous argument 'HEAD~22': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

When I run `git log --graph --oneline main..HEAD`, it generates very long logs (wc -l results in 22), when I only checked out this PR.

infra/style/format.sh
```diff
@@ -6,13 +6,42 @@ set -e
 APPLY_PATCH_OPTION="-a"
 while [[ "$#" -gt 0 ]]; do
   case $1 in
-    --no-apply-patches) APPLY_PATCH_OPTION="";;
+    --no-apply-patches)
+      APPLY_PATCH_OPTION=""
+      shift
+      ;;
+    --diff-only)
```
dayo09, line 13: How about providing short option together, like `-d`?

infra/style/format.sh
jinevening: Can you show your git log output?

infra/style/format.sh
jinevening: > Or you can make diff-only as default.

I've thought about this idea, but it has some side effects to other codes. We may need some discussion before doing that.

infra/style/format.sh
jinevening: I've updated the code. Thanks!

infra/style/format.sh
dayo09: My main branch and the upstream/main branch had conflict. After I resolve the conflict, the problem's gone. ;-D

dayo09 (approved):
LGTM with some comments

shs-park (approved):
LGTM!
=)

Issue #34, opened 2025-04-09
[CI] Publish TICO to pypi


dayo09:
### What?

Let's publish tico to public pypi artifactory.

### How?
1. Do we need an open source process?
About internal open source process: I posted an issue on opensource hub. Let me write here if I get replied.

2. Name?
There's no `TICO` in pypi yet. We can maintain the project name.

3. From which account?
`nnfw@samsung.com` can be used.

dayo09:
> 1. Do we need an open source process?

I got an reply that it's redundant :-D

dayo09:
### Testpypi

DRAFT: https://github.com/Samsung/TICO/pull/50

https://github.com/Samsung/TICO/actions/runs/14398387552/job/40378542944

dayo09:
### pypi release done!

It will be published daily.

https://pypi.org/project/tico/0.1.0.dev250411/#description

dayo09:
Let's close this after it's merged..

shs-park:
I got below mail from PyPI.
Please note
=)

---

Sender : PyPI <noreply@pypi.org>
Date : 2025-04-11 18:55 (GMT+9)
Title : [PyPI] Trusted Publisher for project tico can be made more secure

A Trusted Publisher for project [tico](https://pypi.org/project/tico/) was just used from a CI/CD job configured with a GitHub environment. The environment used was: pypi.

Since the Trusted Publisher is configured to allow any environment, for security reasons we recommend constraining it to only one.

If you are an owner of this project, you can automatically constrain this Trusted Publisher to 'pypi' by following this link: [constrain publisher](https://pypi.org/manage/project/tico/settings/publishing/?constrained_publisher_id=5c9d648c-3135-44c8-bb14-6dc596bcabac&constrained_environment_name=pypi#constrain_environment-modal).

Alternatively, you can do this manually by going to the project's [publishing settings](https://pypi.org/manage/project/tico/settings/publishing/), deleting the existing Trusted Publisher and creating a new one with the environment set to 'pypi'.

If you have questions, you can email [admin@pypi.org](mailto:admin@pypi.org) to communicate with the PyPI administrators.

dayo09:
@shs-park I changed it by constraining environment to `pypi` just now. Let me know if additional email comes. Thanks!  :-D

PR #50, opened 2025-04-11
[CI] Publish to pypi

dayo09:
Let's publish nightly packages to pypi.

#34


dayo09:
FOR REVIEWERS, the last second deployement was successful. And the last deployment was failed due to the same file name, so please don't care.

Issue #52, opened 2025-04-14
[test] Add backend-specific optimization support before compilation


mhs4670go:
## What

In our current quantization test pipeline, we follow these steps:

1. Convert a PyTorch module to a Circle model.
2. If the module provides `get_calibration_data()`, quantize it.
3. Compile it using the backend compiler.
4. Run the compiled binary to validate functionality.

However, before step 3, we want to insert an optimization step that can apply transformations like a NCHW-to-NHWC pass on the Circle model. This step should be customizable per backend and per module for extensibility, since:
- Different backends support different optimization passes.
- Different modules may benefit from or require different optimization strategies.

## Proposed Solution

I propose the following extensible design:

1. Modules implement `get_optimization_config(backend: BACKEND) -> dict`

This method returns a backend-specific optimization configuration dictionary.

### Example

```python
class MyModule(torch.nn.Module):
  # ..
  def get_optimization_config(self, backend: BACKEND):
      if backend == BACKEND.TRIV24:
          return {"fuse_avgpool": False, "representative_dataset_size": 256}
      elif backend == BACKEND.QNN:
          return {"fuse_avgpool": True, "quantize_type": "per_channel"}
      return {}
```

2. Introduce a `BackendOptimizer` interface and implement backend-specific optimizer.

```python
class BackendOptimizer:
    def optimize(self, circle_model_path: str, config: dict) -> str:
        raise NotImplementedError()

class Triv2Optimizer(BackendOptimizer):
  def optimize(self, circle_model_path: str, config: dict) -> str:
    onecc_optimize(config)
    # ..
    return final_path (or circle_model)
```

3. Use a registry to associate optimizers with backends

```python
OPTIMIZER_REGISTRY = {
    BACKEND.TRIV24: TRIV24Optimizer(),
    BACKEND.QNN: QNNOptimizer(),
}
```

4. Update the test flow to call the optimizer before compilation

```python
config = module.get_optimization_config(backend)
optimizer = get_optimizer_for_backend(backend)
optimized_model = optimizer.optimize(circle_model_path, config)
```

## Future Work (optional)

We may later formalize the optimization configuration using a structured data format such as a Python dataclass, like this:

```python
@dataclass
class OptimizationConfig:
    fuse_avgpool: bool = False
    quantize_type: str = "per_tensor"
    representative_dataset_size: int = 100
```

This would improve:
- Type safety and validation
- IDE support (auto-completion, documentation)
- Maintainability and consistency across modules

[...]

PR #53, opened 2025-04-14
[test] Temporarily turn off quantization test for AdaptiveAvgPool

mhs4670go:
This commit temporarily turns off a quantization test.

It will be turned on after #52 is resolved.

### DIFF
--- a/test/modules/op/avg_pool2d.py
+++ b/test/modules/op/avg_pool2d.py
@@ -41,9 +41,9 @@ class AdaptiveAvgPool(torch.nn.Module):
         return (torch.randn(1, 128, 7, 7),)
 
     # This test should be done after applying NCHW_to_NHWC with onecc.
-    def get_calibration_data(self):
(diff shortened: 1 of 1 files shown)


shs-park (approved):
LGTM!
=)

jinevening (approved):
LGTM

test/modules/op/avg_pool2d.py
```diff
@@ -41,9 +41,9 @@ def get_example_inputs(self):
         return (torch.randn(1, 128, 7, 7),)
 
     # This test should be done after applying NCHW_to_NHWC with onecc.
```
jinevening, line 43: Why does this test fail Without NCHW_to_NHWC ? It might be inefficient due to Transpose Ops, but shouldn't the value test pass?

mhs4670go: @jinevening I guess the axis values of Mean op are not supported in our backend compiler.

PR #54, opened 2025-04-15
[draft] Use `uv` for check-pr workflow

miusic:
This draft checks how fast using `uv` is

TICO-DCO-1.0-Signed-off-by: Jihun-Kim <jeyjey6@naver.com>


mhs4670go:
It's good to have any descriptions about the background of this PR:)

miusic:
@mhs4670go 
In our backend compiler repo, `uv` was applied to some workflows and it speeds up incredibly
So I just want to check if it works in `TICO`, too
This draft is for checking the effect of using `uv` :)

miusic:
Ah, this workflow cannot be tested in this way.. 😕

PR #55, opened 2025-04-15
[draft] Check effect of uv

miusic:
This draft checks effect of the `uv`

TICO-DCO-1.0-Signed-off-by: Jihun-Kim <jeyjey6@naver.com>

---
#### Background (https://github.com/Samsung/TICO/pull/54#issuecomment-2803807444)
In our backend compiler repo, `uv` was applied to some workflows and it speeds up incredibly
So I just want to check if it works in `TICO`, too
This draft is for checking the effect of using `uv` :)

+) This draft is just for testing
I have no plan to apply it, yet

### DIFF
--- a/.github/workflows/check-uv.yaml
+++ b/.github/workflows/check-uv.yaml
@@ -1,0 +1,80 @@
+name: Check UV effect
+
+on:
+  pull_request:
+
+jobs:
+  build-and-test:
+    runs-on: tico-linux
+    container:
+      image: ubuntu:22.04
+    strategy:
+      matrix:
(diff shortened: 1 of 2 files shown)


miusic:
It looks like that there is no dramatic change in the time 🤔

**With uv (1m 44s):** https://github.com/Samsung/TICO/actions/runs/14462237959/job/40556822348?pr=55#step:12:1
**Without uv (2m 12s):** https://github.com/Samsung/TICO/actions/runs/14438666927/job/40484081540?pr=53#step:12:1

miusic:
IMHO, it reduces the spending time, but not enough to consider to adopt

PR #58, opened 2025-04-16
[test] Remove nightly dependency

mhs4670go:
This commit removes nightly dependencies.

We won't have to worry about the version updates as much if we change it like this.


seanshpark (approved):
LGTM

dayo09 (approved):
LGTM

Issue #59, opened 2025-04-17
[CI] Remove redundant python installs in Check PR workflow


periannath:
# What

Check PR workflow installs python twice.

#### 1. Installs latest python using `actions/setup-python@v5`

- Installs python 3.13

https://github.com/Samsung/TICO/blob/b433d3388be423aedf50f260b774e3054db017d7/.github/workflows/check-pr.yaml#L71-L73

#### 2. Installs python using `apt`

- Installs python 3.10

https://github.com/Samsung/TICO/blob/b433d3388be423aedf50f260b774e3054db017d7/.github/workflows/check-pr.yaml#L75-L77

It would be better to install python once since it is hard to know which python version is used.

mhs4670go:
Ah, thanks for the works. I remembered I added the L77 because I encountered there's no pip or venv. But, I'm not sure.. 

Anyway, it would be good to resolve the issue.

PR #60, opened 2025-04-17
[Draft][CI] Remove duplicated python installs

periannath:
This draft removes duplicated python installs

---

For issue #59


periannath:
>  one-compiler depends on python3-venv; however:
>   Package python3-venv is not installed.

From https://github.com/Samsung/TICO/actions/runs/14508339362/job/40701781800?pr=60#step:8:28

Looks like `python3-venv` should be installed to use one-compiler. I'll remove `setup-python` step and restore `apt install` step.

PR #70, opened 2025-04-30
[ccex] Add --no-deps for dist install

### LINKED ISSUE #69: [ccex] Remove `--force-reinstall` from `--dist` install
# What

Let's remove `--force-reinstall` from `--dist` install.

# Why

Here is current `infra/scripts/install.sh` behavior when `--dist` is given.

1. Install requirements including torch
   - `torch==2.6.0` is installed as default
1. Install tico from wheel
   - `torch==2.7.0` is re-installed

That script installs latest torch regardless of given torch version. It would be better to use installed torch version during `tico` installation from wheel.

periannath:
This commit adds --no-deps for dist install

---

To resolve #69 
For issue #74

### DIFF
--- a/infra/scripts/install_requirements_2_5.txt
+++ b/infra/scripts/install_requirements_2_5.txt
@@ -4,3 +4,6 @@ wheel==0.43.0
 circle_schema
 
 torch==2.5.0
+cffi==1.17.1
+packaging==25.0
+pyyaml==6.0.2
--- a/infra/scripts/install.sh
+++ b/infra/scripts/install.sh
@@ -75,13 +75,13 @@ SCRIPTS_DIR="${CCEX_PROJECT_PATH}/infra/scripts"
 
 if [ "$_TORCH_VER" == "nightly" ]; then
   echo "Install package dependencies from torch nightly version"
-  python3 -m pip install --force-reinstall -r "${SCRIPTS_DIR}/install_requirements_dev.txt"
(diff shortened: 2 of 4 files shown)


periannath:
I met this error in my local environment.

```bash
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
tico 0.1.0 requires cffi, which is not installed.
tico 0.1.0 requires packaging, which is not installed.
tico 0.1.0 requires pyyaml, which is not installed.
```

Looks like we should add those packages in `requirements.txt`.

infra/scripts/install_requirements_2_5.txt
```diff
@@ -4,3 +4,6 @@ wheel==0.43.0
 circle_schema
 
 torch==2.5.0
+cffi==1.17.1
+packaging==25.0
+pyyaml==6.0.2
```
periannath, line 9: Using latest release versions.

mhs4670go: @dayo09 Not related this PR but on second thought, `pyyaml` doesn't need to be inside setup.py's `install_requires` and install requirement. Isn't it more like a test requirement?

mhs4670go (approved):
LGTM

infra/scripts/install_requirements_2_5.txt
dayo09: @mhs4670go I agree 👍

seockho-kim (approved):
LGTM

Issue #72, opened 2025-04-30
[CI] Failed to pull docker image due to pull rate limit


periannath:
# What

```bash
  Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
```
From https://github.com/Samsung/TICO/actions/runs/14747261148/job/41396869764?pr=70#step:2:19

seanshpark:
Wait for 30 minutes or so will pass the limit... =_=

periannath:
# Options

## Option 1 : Use Google Artifact Registry Mirror

- Google provides docker image mirror
  - https://cloud.google.com/artifact-registry/docs/pull-cached-dockerhub-images#cli
- URL: `mirror.gcr.io/ubuntu:20.04`

## Option 2 : Use GitHub Packages

- GitHub provides Package which can upload docker images
  - https://docs.github.com/en/packages/learn-github-packages/introduction-to-github-packages
- Some packages were already uploaded at Samsung organization
  - https://github.com/orgs/Samsung/packages
- URL: `ghcr.io/<USER>/ubuntu:20.04`

mhs4670go:
Option 1 looks good.

### Pros:

- Avoids Docker Hub rate limits (`toomanyrequests`)
- No need to set up Docker Hub credentials or GHCR
- Operated by Google, usually stable and fast

### Cons:

- Not officially documented or guaranteed for public use (originally intended for GKE)
- Might lag behind upstream `ubuntu:22.04` updates
- Could be blocked in restricted network environments

If stability and ease are top priorities and you're okay with potentially slower updates, this is a very practical workaround.

Also, I found some other options.

## Option 3: Add Docker Hub login credentials

### Pros:

- Easiest and fastest fix
- Allows using the official ubuntu:22.04 image directly
- Requires only adding secrets (DOCKER_USERNAME, DOCKER_PASSWORD) to GitHub

### Cons:

- Logged-in users still have a limit (200 pulls per 6 hours for free accounts)
- Still dependent on Docker Hub availability

## Option 4: Mirror the image to GitHub Container Registry (GHCR)

### Pros:

- Effectively no rate limits when using GITHUB_TOKEN in GitHub Actions
- More stable and controlled environment (no reliance on Docker Hub)
- Can version or customize the image if needed

### Cons:

- Requires initial setup: pull → tag → push to GHCR
- Will not automatically track upstream updates unless automated

## Recommendation:

[...]

PR #75, opened 2025-04-30
[CI] Add docker credentials in Check PR workflow

periannath:
This commit adds docker credentials in Check PR workflow

---

For issue #72

### DIFF
--- a/.github/workflows/check-pr.yaml
+++ b/.github/workflows/check-pr.yaml
@@ -59,6 +59,9 @@ jobs:
     runs-on: tico-linux
     container:
       image: ubuntu:22.04
+      credentials:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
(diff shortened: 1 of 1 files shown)


.github/workflows/check-pr.yaml
```diff
@@ -59,6 +59,9 @@ jobs:
     runs-on: tico-linux
     container:
       image: ubuntu:22.04
+      credentials:
+        username: ${{ secrets.DOCKERHUB_USERNAME }}
+        password: ${{ secrets.DOCKERHUB_TOKEN }}
```
periannath, line 64: I've checked that this grammar is working correctly on GHES.

@mhs4670go Would you please check that registered secret name is same as this code?

periannath: Looks like `secrets` can't be tested on forked repo.

periannath:
Let's use #76 instead.

PR #90, opened 2025-05-14
[DRAFT] Preserve SDPA

mhs4670go:
This commit preserves SDPA for runtime optimization.

### DIFF
--- a/tico/serialize/operators/op_sdpa.py
+++ b/tico/serialize/operators/op_sdpa.py
@@ -1,0 +1,58 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, List, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torch._ops
+    import torch.fx
+import torch
+from circle_schema import circle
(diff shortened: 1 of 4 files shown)


mhs4670go:
@glistening You can refer to this draft when you want to see overall graphs.

mhs4670go:
@glistening In the offline discussion, I heard that you want to abstract Attentions in a more general way. First of all, I will find the way of exporting `Florence2SdpaAttention` class itself as a single node. 

If you let me know about the spec overview of CircleAttention(?) (inputs/outputs or attributes) would be helpful to do the work.

tico/serialize/operators/op_sdpa.py
```diff
@@ -0,0 +1,58 @@
...
+        node: torch.fx.Node,
+    ) -> circle.Operator.OperatorT:
+        args = SDPAArgs(*node.args, **node.kwargs)  # type: ignore[arg-type]
+        outputs = [node]
+
+        op_index = get_op_index(
+            circle.BuiltinOperator.BuiltinOperator.CUSTOM, self._op_codes, "SDPA"
+        )
+
+        inputs = list(node.args) + [kw.value for kw in node.kwargs.values()]  # type: ignore[union-attr]
```
dayo09, line 50: Parsing kwargs here seems a bit dangerous, because some optional kwargs may be ignored and the arg order cannot be maintained. How about using parsed **args** from `args = SDPAArgs(*node.args, **node.kwargs)`?

```py
# Adding this iterate function to 'class SDPAArgs' will help you
class SDPAArgs:
...
    def __iter__(self):
      return (getattr(self, field.name) for field in dataclasses.fields(self))


...

inputs = list(args)
```

glistening:
I told @mhs4670go this morning:

- I wanted higher level sdpa layer (e.g. `Florence2SPDALayer`, `LlamaAttention`, ...), which is not the part of aten IR.
- I want high-level spda layer to cover serveral different types of attentions. For example, 
  - LlamaAttention has additional operator in attention like RoPE, and its relevant parameters. (RoPE type, ...).
  - Attention may work differently whether it is in encoder or decoder
  - ...
- It may be impossible or difficult to preserve high-level spda layer in current `TICO`'s way, which depends on the aten IR of trace.
- I will check and think more.

Issue #100, opened 2025-05-20
Propagate quantization parameters across 1:N mapping ops during Aten-to-Circle lowering


mhs4670go:
## What

Currently, we have [a pass](https://github.com/Samsung/TICO/blob/962669a4b6c9e2ba92193de75e4adca48f0f5f2f/tico/experimental/quantization/passes/propagate_qparam_forward.py) at the Aten level that propagates quantization parameters (qparams) forward and backward through the graph. However, when converting from Aten to Circle, some ops are lowered with a 1:N mapping (i.e., one Aten op becomes multiple Circle ops). Even if the resulting Circle ops are capable of carrying qparams, the propagation doesn’t happen across these boundaries.

This gap causes qparams to be lost in the lowering process, leading to incorrect quantized behavior in the final Circle model.

## Proposed fix

- Extend the qparam propagation logic to support 1:N mappings during Aten-to-Circle lowering.

jinevening:
> Extend the qparam propagation logic to support 1:N mappings during Aten-to-Circle lowering.

Could you explain a little bit more? Do you plan to extend the existing pass? Or do you want to add a new pass?

> This gap causes qparams to be lost in the lowering process, leading to incorrect quantized behavior in the final Circle model.

Could you give an example?

mhs4670go:
@jinevening 

> Could you give an example?

When we convert a [torch.ops.circle_custom.maxpool2d](https://github.com/Samsung/TICO/blob/324b472a3b43aa8bd17f6b1e7de61b18058bb9f3/tico/serialize/operators/op_max_pool2d_with_indices.py) op, a pad op can be inserted according to the arguments.

As of now, a pad op doesn't have qparams when we get the quantized torch modules from external PTQ tools because qparams are exists only in `torch.ops.circle_custom.maxpool2d`.

> Do you want to add a new pass?

No. I'll just add a logic that propagates maxpool's qparam to such ops in op_xxx.py.

jinevening:
IIUC, your plan is to set pad Op's qparam when the pad Op is created in`op_max_pool2d_with_indices.py`. That looks good to me.

I asked the question because I was worried about adding a general propagation logic to `op_xxx.py`. That will give multiple roles to `op_xxx.py`, which is not desirable.

mhs4670go:
> That will give multiple roles to op_xxx.py, which is not desirable.

I agree on this. Well, I guess the way I described about what I would do was too broad.

[...]

PR #102, opened 2025-05-20
[serialize] Set qparams for Conv and MaxPool padding

mhs4670go:
This commit propagates input qparam across 1:N mapping ops during Aten-to-Circle lowering.

Related: #100 

### DIFF
--- a/tico/serialize/operators/op_conv2d.py
+++ b/tico/serialize/operators/op_conv2d.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict, List, TYPE_CHECKING
+from typing import Dict, List, Optional, TYPE_CHECKING
 
 if TYPE_CHECKING:
     import torch._ops
@@ -24,6 +24,7 @@ from tico.serialize.circle_mapping import extract_circle_dtype, extract_shape
 from tico.serialize.operators.hashable_opcode import OpCode
 from tico.serialize.operators.node_visitor import NodeVisitor, register_node_visitor
 from tico.serialize.operators.utils import create_builtin_operator, get_op_index
+from tico.serialize.quant_param import QPARAM_KEY, QuantParam
 from tico.utils.define import define_pad_node
 from tico.utils.padding import is_same_padding, is_valid_padding, SAME, VALID
 from tico.utils.validate_args_kwargs import Conv2DArgs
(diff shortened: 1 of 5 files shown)


tico/serialize/operators/op_conv2d.py
```diff
@@ -23,7 +23,12 @@
 from tico.serialize.circle_mapping import extract_circle_dtype, extract_shape
 from tico.serialize.operators.hashable_opcode import OpCode
 from tico.serialize.operators.node_visitor import NodeVisitor, register_node_visitor
-from tico.serialize.operators.utils import create_builtin_operator, get_op_index
+from tico.serialize.operators.utils import (
+    create_builtin_operator,
+    get_integer_dtype_min,
```
miusic, line 28:
```suggestion
```
It seems fine to remove :)

miusic (dismissed):
LGTM

tico/serialize/operators/utils.py
```diff
@@ -49,3 +50,42 @@ def create_builtin_operator(
     operator.inputs = [graph.get_tid(input) for input in inputs]
     operator.outputs = [graph.get_tid(output) for output in outputs]
     return operator
+
+
+def get_integer_dtype_min(dtype_str: str):
```
jinevening, line 55:
```suggestion
def get_integer_dtype_min(dtype_str: str) -> int:
```

Please add unit tests for this function. There are some cases that produce wrong outputs.

"int1" -> -1
"int0" -> -0.5

Better to use if statement rather than re to avoid such corner cases.

jinevening:
The commit title does not match with the content.

It would be better to be something like "Set qparams for Conv and MaxPool padding".

tico/serialize/operators/utils.py
miusic: Good point 👍
However, IMHO, is there any possibility for such cases like `int1` or `int0`?
If we use signed type, at leaset 2 bit is needed I guess.. 🤔

jinevening: They are "corner cases" as I mentioned :) They may rarely happen, but we shouldn't just ignore them.

miusic: Thank you for clarification! 😄 As you said, it would better to prepare for such cases

jinevening (approved):
LGTM

seockho-kim (approved):
LGTM

miusic (approved):
LGTM 😄

Issue #97, opened 2025-05-19
Remove Redundant Reshape when exporting GroupNorm


mhs4670go:
## What

While exporting a model using `torch.export.export`, I noticed that the resulting graph includes inefficient or redundant reshape operations that are not present in the equivalent ONNX graph exported via `torch.onnx.export`.

These reshape ops appear to be unnecessary and can negatively impact performance on certain inference backends, especially when they are inserted repeatedly or between ops that preserve shape.

![Image](https://github.com/user-attachments/assets/99552b1e-a3ae-4a35-9083-d576abddb797)

mhs4670go:
This pattern is made from [DecomposeGroupNorm](https://github.com/Samsung/TICO/blob/962669a4b6c9e2ba92193de75e4adca48f0f5f2f/tico/passes/decompose_group_norm.py). We may need to change the way they are decomposed.

PR #103, opened 2025-05-20
Remove Redundant Reshape when exporting GroupNorm

mhs4670go:
This commit removes redundant reshapes when exporting GroupNorm.

Related: #97 

### DIFF
--- a/tico/passes/decompose_group_norm.py
+++ b/tico/passes/decompose_group_norm.py
@@ -88,6 +88,25 @@ class DecomposeGroupNorm(PassBase):
     def __init__(self):
         super().__init__()
 
+    def _insert_norm(self, graph, tensor, eps):
+        """
+        Insert (tensor - mean) / sqrt(var + eps)) into the graph
+          and return the normalized tensor node.
+        """
+        mean = graph.call_function(
+            torch.ops.aten.mean.dim, (tensor, [-1]), {"keepdim": True}
+        )
+        dev = graph.call_function(torch.ops.aten.sub.Tensor, (tensor, mean))
+        sqr = graph.call_function(torch.ops.aten.pow.Tensor_Scalar, (dev, 2))
+        var = graph.call_function(
+            torch.ops.aten.mean.dim, (sqr, [-1]), {"keepdim": True}
+        )
+        inv_std = graph.call_function(
+            torch.ops.aten.rsqrt.default,
+            (graph.call_function(torch.ops.aten.add.Tensor, (var, eps)),),
+        )
+        return graph.call_function(torch.ops.aten.mul.Tensor, (dev, inv_std))
+
     def call(self, exported_program: ExportedProgram) -> PassResult:
         logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


tico/passes/decompose_group_norm.py
```diff
@@ -155,52 +155,85 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
...
+                        torch.ops.aten.mul.Tensor,
+                        (layer_deviation, rsqrt),
+                    )
+                else:
+                    layer = graph.call_function(
+                        # Sometimes, `x` has a stride for NHWC, which can't be reshaped with `aten.view.default`.
+                        # TODO Find out how to process such case properly.
+                        torch.ops.aten.reshape.default,
+                        (x, pack_shape),
+                    )
```
dayo09, line 196: (optional) I think this if-else statement could shrink to include reshape only

tico/passes/decompose_group_norm.py
```diff
@@ -155,52 +174,20 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
             pack_shape = [layer_size, norm_size]
 
             with gm.graph.inserting_before(node):
-                layer = graph.call_function(
-                    # Sometimes, `x` has a stride for NHWC, which can't be reshaped with `aten.view.default`.
-                    # TODO Find out how to process such case properly.
```
dayo09, line 160: (optional) How about keeping this comment?

dayo09 (approved):
LGTM

tico/passes/decompose_group_norm.py
mhs4670go: Good point. Actually, I removed it intentionally as we can just process them when we exactly meet the case.

tico/passes/decompose_group_norm.py
```diff
@@ -88,6 +88,25 @@ class DecomposeGroupNorm(PassBase):
...
 
+    def _insert_norm(self, graph, tensor, eps):
+        """
+        Insert (tensor - mean) / sqrt(var + eps)) into the graph
+          and return the normalized tensor node.
+        """
+        mean = graph.call_function(
+            torch.ops.aten.mean.dim, (tensor, [-1]), {"keepdim": True}
+        )
+        dev = graph.call_function(torch.ops.aten.sub.Tensor, (tensor, mean))
```
miusic, line 99: (just out of curiosity)
`dev` here denotes `deviation`?

mhs4670go: Right.

tico/passes/decompose_group_norm.py
```diff
@@ -88,6 +88,25 @@ class DecomposeGroupNorm(PassBase):
     def __init__(self):
         super().__init__()
 
+    def _insert_norm(self, graph, tensor, eps):
+        """
+        Insert (tensor - mean) / sqrt(var + eps)) into the graph
+          and return the normalized tensor node.
+        """
```
miusic, line 95:
```suggestion
        """
        Insert below arithmetic into the graph
        (tensor - mean) / sqrt(var + eps))
        ~~~~~~~~~~~~~~~  ~~~~~~~~~~~~~~~~~
        ^                 ^
        denoted as `dev`  denoted as `inv_std`
          and return the normalized tensor node.
        """
```
(optional)
How about to enrich description..? I suggest this because it was bit confused to me what the `dev` means at the first glance

miusic (approved):
LGTM w/ minor comment https://github.com/Samsung/TICO/pull/103#discussion_r2099448051 👍

tico/passes/decompose_group_norm.py
mhs4670go: Then, I just rename the variable next PR:) Thanks.

PR #110, opened 2025-05-26
Support aten.op.dim

mhs4670go:
This commit supports aten.op.dim.


seockho-kim (dismissed):
LGTM

Seunghui98:
Why don't you change the PR title and commit messgae to `Support aten.max.dim`?

PR #111, opened 2025-05-26
Support aten.abs

seockho-kim:
This commit support aten.abs operator.

ONE-DCO-1.0-Signed-off-by: Seockho Kim seockho.kim@samsung.com

### DIFF
--- a/tico/utils/validate_args_kwargs.py
+++ b/tico/utils/validate_args_kwargs.py
@@ -36,7 +36,7 @@ This file includes OpArgs classes that provide arguments with type annotations.
 @dataclass
 class AbsArgs:
     """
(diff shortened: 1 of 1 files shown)


tico/utils/validate_args_kwargs.py
```diff
@@ -32,6 +32,16 @@
 """
 
 
+@enforce_type
+@dataclass
+class AbsArgs:
+    """
+    cos(Tensor self) -> Tensor
```
jinevening, line 39:
```suggestion
    abs(Tensor self) -> Tensor
```

Seunghui98 (approved):
LGTM 👍

PR #115, opened 2025-05-27
[serialize] Ignore None output

dayo09:
Let's ignore None output from graph signature.


jinevening:
When does this happen?

mhs4670go:
It would be better to have test modules for this.

jinevening (approved):
LGTM

PR #119, opened 2025-05-28
[passes] Introduce LegalizeAvgpool2d

dayo09:
This commit is to accept torch.nn.AvgPool2d(count_include_pad = False). count_include_pad must be true for Circle Avgpool2D. Therefore, generate mask to make equivalent operation.

---

From #93

### DIFF
--- a/tico/passes/legalize_avgpool2d.py
+++ b/tico/passes/legalize_avgpool2d.py
@@ -1,0 +1,135 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


mhs4670go:
I think current codes from main branch has a bit wrong assumption.

In short, just setting SAME padding option to the circle node and not creating `CirclePad` when the SAME padding condition is met will do the trick.

`CircleAveragePool2D` op has no explicit `count_include_pad` parameter;it just uses a fixed behavior. Circle's "SAME" padding mode for average pooling excludes the padded area from the average count. It behaves like `count_include_pad=False`. But, PyTorch uses `count_include_pad=True` by default.

Therefore, we need to explicitly introduce the padding as part of the computation so that it gets counted in the average. That is why current codes work well.

---

The example module from this PR can be simply converted like this.

tico/passes/legalize_avgpool2d.py
```diff
@@ -0,0 +1,135 @@
...
+from torch.export import ExportedProgram
+
+from tico.utils import logging
+from tico.utils.passes import PassBase, PassResult
+from tico.utils.trace_decorators import trace_graph_diff_on_pass
+from tico.utils.validate_args_kwargs import AvgPool2dArgs
+
+
+@trace_graph_diff_on_pass
+class LegalizeAvgpool2D(PassBase):
```
jinevening, line 29: Better to rename to `LegalizeAvgpool2DCountIncludePad`.

And please add PassTest.

PR #127, opened 2025-06-02
[serialize] Support Avgpool2D with count_include_pad=False

dayo09:
Let's support AvgPool2D with count_include_pad=False partially.

### DIFF
--- a/test/modules/op/avg_pool2d.py
+++ b/test/modules/op/avg_pool2d.py
@@ -14,6 +14,8 @@
 
 import torch
 
+from test.utils.tag import skip
+
 
 class SimpleAvgPool(torch.nn.Module):
     def __init__(self):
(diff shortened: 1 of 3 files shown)


tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -26,16 +27,79 @@
...
 
 @register_node_visitor
 class AvgPool2DVisitor(NodeVisitor):
+    """
+    This class defines how to serialize AvgPool2D operation into Circle IR.
+
+    Torch                           | Circle
+
+    count_include_pad: True/False   | (count_include_pad): Always False
+    (padding): Always "valid"       | padding: "valid"/"same"
```
jinevening, line 42: Torch avgpool2d Op can have padding. It's not always "valid".

https://docs.pytorch.org/docs/stable/generated/torch.nn.functional.avg_pool2d.html

tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -26,16 +27,79 @@
...
+            return False
+        else:
+            return True
+
+    def has_same_padding(self, args: AvgPool2dArgs) -> bool:
+        input_shape = list(extract_shape(args.input))
+        kernel_size = args.kernel_size
+        stride = args.stride
+        padding = args.padding
+
```
jinevening, line 65: The below equation only works when ceil_mode == False. Better to add a guard here

```suggestion
        # TODO Update this function when supporting ceil_mode = True
        assert args.ceil_mode is False
```

tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -26,16 +27,79 @@
...
+            (input_shape[1] + padding[0] * 2 - kernel_size[0]) / stride[0] + 1
+        )
+        output_width = math.floor(
+            (input_shape[2] + padding[1] * 2 - kernel_size[1]) / stride[1] + 1
+        )
+
+        if input_shape[1] == output_height and input_shape[2] == output_width:
+            return True
+        else:
+            return False
```
jinevening, line 76: (optional) For simplicity

```suggestion
        return input_shape[1] == output_height and input_shape[2] == output_width
```

test/modules/op/avg_pool2d.py
```diff
@@ -96,3 +98,49 @@ def forward(self, tensor):
 
     def get_example_inputs(self):
         return (torch.randn(2, 4, 8, 16),)
+
+
+class AvgPoolWithNoCountIncludePad(torch.nn.Module):
```
jinevening, line 103:
```suggestion
class AvgPoolWithSamePaddingNoCountIncludePad(torch.nn.Module):
```

tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -81,32 +144,30 @@ def define_padding_node():
...
+            # Don't care count_include_pad
+            result = self.define_avgpool_node(
+                [avgpool_input], [node], "VALID", stride, kernel_size
+            )
+        elif count_include_pad is True:
+            # Add padding before avgpool2d
+            # Circle's avgpool2d does not support count_include_pad=True, so we need to add padding manually
             avgpool_input = define_padding_node()
+            result = self.define_avgpool_node(
+                [avgpool_input], [node], "VALID", stride, kernel_size
```
Seunghui98, line 157: May I ask a question? 🙂
if count_include_pad is True, Why we set "VALID"...?

dayo09: Let me share this psuedo code for better understanding!


```py
if count_include_pad == True:
    (Circle cannot represent count_include_pad=True in AvgPool2D. Therefore we manually add zero padding node.)
    DEFINE zero padding node
    DEFINE avgpool node with no padding (valid)
if count_include_pad == False:
    (Lucky! Circle can represent count_include_pad=False)
    define avgpool node with same/valid padding.
    
    (However, it cannot represent all paddings. So, if the padding is not same or valid, we throw an error.)
    if the paddding is neither same nor valid, throw an error.
```

dayo09: Plus, let me revise the statements.

jinevening (approved):
LGTM

tico/serialize/operators/op_avg_pool2d.py
Seunghui98: @dayo09 
Ah! I understood it 😀
Thanks for letting me know 👍

Issue #126, opened 2025-05-29
Remove the Reshape surrounding Add → Softmax in the encoder


mhs4670go:
## What

Right before the attention soft-max, the exported graph still performs an **extra reshape round-trip**.

Eliminating those reshapes would let us feed `Softmax` directly with the `(B × H, T, S)` tensor.

Related: #124

## Why

* **Less data movement** per layer → lower latency on bandwidth-bound NPUs.
* **Cleaner graph** → makes subsequent pattern-matching / fusion passes simpler.

![Image](https://github.com/user-attachments/assets/0eb67bac-b9b7-4925-98b4-791a0f127bac)

Issue #113, opened 2025-05-27
Optimize NMT


dayo09:
### What?

Let's optimize fairseq model.

#### Tasks

- [x] ~~Remove redundant transpose https://github.com/Samsung/TICO/issues/113#issuecomment-2911385355~~ It's optimized at one-optimize.

(ADDED)

It look TICO is well-optimized than onnx2circle for now.

I just eye-checked that **decoder model was exactly the same.**
And for encoder model, **onnx2circle-converted model had a little bit redundancy in attention.**

  | TICO (Rose-P) | TICO, q8 (Rose-L) | onnx2circle, q8 (Rose-L)
-- | -- | -- | --
encoder | 31.83ms | ~~50.11~~ 52.600 ms, 39976522 cycle | 52.684 ms, 40040067 cycle
decoder | 3.529ms | ~~5.207~~ 5.482ms, 4166578 cycle | 5.482 ms, 4166618 cycle

dayo09:
### Remove redundant transpose

These pattern is found in TICO-exported nmt encoder.

![Image](https://github.com/user-attachments/assets/8ba55a98-ecd0-4653-9528-aaebcf195f65)

![Image](https://github.com/user-attachments/assets/143a646b-6578-4254-96b2-e821dcb2a11b)

Whoever interested in this task, please name your self and try this :-D

dayo09:
### Difference in Attention

After optimization, encoder has this difference in attention. (TICO is better optimized better than onnx2circle)

#### onnx2circle

![Image](https://github.com/user-attachments/assets/3423e74a-34c3-41b2-9135-52334a54b5d2)

#### TICO

![Image](https://github.com/user-attachments/assets/e9f3e2ce-135a-4eaf-bf76-f58e0b6243bc)

PR #130, opened 2025-06-05
[passes] Adds another case for removing Reshape

mhs4670go:
This commit adds more case for removing Reshape surrounding Add -> Softmax.

image

Related: #126, #113

### DIFF
--- a/test/unit_test/pass_test/test_remove_redundant_reshape.py
+++ b/test/unit_test/pass_test/test_remove_redundant_reshape.py
@@ -126,8 +126,6 @@ class RemoveRedundantReshapePattern3Test(SinglePassValueTest):
     def test_pass(self):
         self.setup(RedundantReshapePattern3())
 
-        self.assertEqual(num_of_ops(self.exported_program(), ops.aten.reshape), 3)
-
         from tico.utils.utils import SuppressWarning
 
         # torch.ops.aten.softmax.int -> torch.ops.aten._softmax.default
@@ -138,6 +136,7 @@ class RemoveRedundantReshapePattern3Test(SinglePassValueTest):
             #     CompositeImplicitAutograd and have functional schema are safe to not decompose.
             self.ep = self.exported_program().run_decompositions()
         self.run_value_test(ConvertLayoutOpToReshape())
+        self.assertEqual(num_of_ops(self.exported_program(), ops.aten.reshape), 3)
 
         self.run_value_test(RemoveRedundantReshapePattern3())
         self.assertEqual(num_of_ops(self.exported_program(), ops.aten.reshape), 0)
@@ -167,8 +166,6 @@ class RemoveRedundantReshapePattern3BroadcastedTest(SinglePassValueTest):
     def test_pass(self):
         self.setup(RedundantReshapePattern3Broadcasted())
 
-        self.assertEqual(num_of_ops(self.exported_program(), ops.aten.reshape), 3)
-
         from tico.utils.utils import SuppressWarning
 
         # torch.ops.aten.softmax.int -> torch.ops.aten._softmax.default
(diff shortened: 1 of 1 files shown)


tico/passes/remove_redundant_reshape.py
```diff
@@ -295,10 +295,13 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
             # Check condition
             reshape_2_input_shape = extract_shape(reshape_2_input)
             reshape_3_input_shape = extract_shape(reshape_3_input)
-            if reshape_2_input_shape != reshape_3_input_shape:
+            if not broadcastable(reshape_2_input_shape, reshape_3_input_shape):
                 continue
```
jinevening, line 299: I think this has to be true. If not, the model is somehow wrong.

```suggestion
            assert broadcastable(reshape_2_input_shape, reshape_3_input_shape)
```

tico/passes/remove_redundant_reshape.py
```diff
@@ -295,10 +295,13 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
             # Check condition
```
jinevening, line 295: Could you add this test case?

reshape_2_input_shape = [1,1,6]
reshape_3_input_shape = [1,1,6]
reshape_2_shape = [1,2,3]
reshape_3_shape = [1,2,3]
softmax_shape = [1,2,3]
reshape_1_shape = [1,1,6]

I'm not sure the current code can block the case.

(optional) Code is very difficult to read. It would be better to separate the function to two parts.
1) Get arguments.
2) Check conditions (add comment for each condition).

mhs4670go: @jinevening 

> I'm not sure the current code can block the case.

Is it the case that should be blocked? 

> Code is a little bit difficult to read

I guess it would be better to do this in refactoring PR.

mhs4670go: Ah, now I got it! I'll add a condition for this. Thanks:)

test/unit_test/pass_test/test_remove_redundant_reshape.py
```diff
@@ -126,14 +126,48 @@ class RemoveRedundantReshapePattern3Test(SinglePassValueTest):
     def test_pass(self):
         self.setup(RedundantReshapePattern3())
 
-        if Version(torch.__version__) <= Version("2.6.0.dev20241015"):
-            self.run_value_test(ConvertLayoutOpToReshape())
-
```
mhs4670go, line 131: This pass should be run after `run_decompositions()`.

Seunghui98:
@mhs4670go 

CI for 2.5 runs fails 😥

test/unit_test/pass_test/test_remove_redundant_reshape.py
mhs4670go: I added a `run_decompositions()` line because `aten.softmax.int` is decomposed to `aten.softmax.default`.

But, at the same time, `run_decompositions()` may replace some reshape ops to views. Therefore, I just pulled down the position of `ConvertLayoutOpToReshape` call. Which would replace all view ops to reshape ops.

test/unit_test/pass_test/test_remove_redundant_reshape.py
```diff
@@ -144,11 +178,51 @@ def test_pass(self):
...
+
+        # torch.ops.aten.softmax.int -> torch.ops.aten._softmax.default
+        with SuppressWarning(UserWarning, ".*quantize_per_tensor"):
+            # Warning details:
+            #   ...site-packages/torch/_subclasses/functional_tensor.py:364
+            #   UserWarning: At pre-dispatch tracing, we assume that any custom op marked with
+            #     CompositeImplicitAutograd and have functional schema are safe to not decompose.
+            self.ep = self.exported_program().run_decompositions()
+        self.run_value_test(ConvertLayoutOpToReshape())
+
```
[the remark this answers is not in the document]
mhs4670go: Umm, I'm not sure I understood the question. Now, you can see how many reshape ops are with L217.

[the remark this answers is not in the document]
Seunghui98: @mhs4670go 
Ah, isn't there a redundant reshape op here?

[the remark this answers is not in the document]
mhs4670go: You're right. This test is from https://github.com/Samsung/TICO/pull/130#discussion_r2134959637 where reshape ops shouldn't be removed.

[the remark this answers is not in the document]
Seunghui98: Ah, I got it! Thanks!!

jinevening (approved):
LGTM

Issue #129, opened 2025-06-05
Internal state of original model changes during `torch.export.export`


dayo09:
### What?

There is a strange phenomenon where the internal state of the original (nn.Module) model changes before and after torch.export.export. This has been observed in EfficientFormer as well (where the forward result appears as FakeTensor - referred to as FakeTensor Leakage), and in some environments (suspected to be torch 2.6.0 or lower), inference does not work properly after exporting in Fairseq (the shape of the output tuple changes).

This has been reported in pytorch/pytorch#155114.

dayo09:
For those who must be curious,

This simple script can reproduce it.

```py
import timm, torch

model: timm.models.efficientformer.EfficientFormer = (
    timm.create_model("efficientformer_l1", pretrained=True).to("cpu")
)

example_inputs = (torch.randn(1, 3, 224, 224),)

# Running before export: GOOD!
res = model.forward(*example_inputs)
print(res) # Expected Tensor Output

exported = torch.export.export(
    model.eval(), args=example_inputs,
)

# Running after export: WEIRD!!
res = model.forward(*example_inputs)
print(res) # FakeTensor
```

PR #131, opened 2025-06-05
[test] Infer before export

dayo09:
Let's infer before export, to avoid the model's internal state change due to 'export leakage'.

---

#129

### DIFF
--- a/test/pt2_to_circle_test/test_pt2_to_circle.py
+++ b/test/pt2_to_circle_test/test_pt2_to_circle.py
@@ -140,6 +140,10 @@ def verify_circle(circle_model_path: str, opt_circle_model_str: str):
 @print_name_on_exception
 def infer_nnmodule(model: torch.nn.Module, example_inputs: tuple):
     with torch.no_grad():
+        # Model should be frozen to compare the result with others.
+        # e.g. BatchNorm running_mean/running_var will be updated during training mode, thus changing the model behavior.
+        model.eval()
+
         _args, _kwargs = helper.get_args_kwargs(example_inputs)
         torch_result = model.forward(*_args, **_kwargs)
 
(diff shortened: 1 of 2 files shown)


test/pt2_to_circle_test/test_pt2_to_circle.py
```diff
@@ -140,6 +140,10 @@ def verify_circle(circle_model_path: str, opt_circle_model_str: str):
 @print_name_on_exception
 def infer_nnmodule(model: torch.nn.Module, example_inputs: tuple):
     with torch.no_grad():
+        # Model should be frozen to compare the result with others.
+        # e.g. BatchNorm running_mean/running_var will be updated during training mode, thus changing the model behavior.
+        model.eval()
```
dayo09, line 145: FOR OTHER REVIEWERS

[Seunghui98](https://github.com/Seunghui98) [1 hour ago](https://github.com/Samsung/TICO/pull/120#discussion_r2128140617)
it's just out of curiousity. 😀
If this line is added, does it not affect other tests?


@[dayo09](https://github.com/dayo09) dayo09 [34 minutes ago](https://github.com/Samsung/TICO/pull/120#discussion_r2128214059)
@Seunghui98 That's a good question.

It seems only BatchNorm family within our test cases are affected - only they fails the test.
WHY? - BatchNorm changes its internal values during inference if it's not set as eval() mode (training mode)
WHY It didn't failed before then? - Before, eval() was set outside this function (in convert_nnmodule_to_pt2), but this PR includes the change that convert_nnmodule_to_pt2 is run on copied model.

mhs4670go (approved):
LGTM

seockho-kim (approved):
LGTM

Issue #132, opened 2025-06-09
Eliminate redundant aten.permute chains even when strides differ


mhs4670go:
## What

Our current “Remove Redundant Permute” pass deleted two consecutive `aten.permute` nodes only when both shape and stride are identical between the original input and the final output.

However, PyTorch may optimize view-in-view striding after a second permute, so **the final stride can differ even though the overall permutation is mathematically the identity.** 

In such cases:

- Every logical index (i, j, k, …) points to the same storage location as the original tensor – the two permutes truly cancel out.
- The differing stride is just metadata describing a non-contiguous view; it does **not** change the values.

Because of this, we were missing valid optimization opportunities whenever PyTorch collapsed strides (e.g. input.stride = (10240, 1280, 64, 1) → output.stride = (512, 64, 512, 1)).

## New optimization rule

The chain is removed regardless of stride.

```
permute(dims1) - permute(dims2)
  - if  dims2 == identity ➜  delete both permutes
  - else                  ➜  fuse into a single permute(fused_dims)
```

Related: #113

jinevening:
Could you show the part of NMT model that this optimization targets? It would be helpful to understand the impact of this issue.

PR #133, opened 2025-06-09
Revise condition of RemoveRedundantPermute pass

mhs4670go:
This commit revises the condition of RemoveRedundantPermute pass.

- eliminate permute ops even when strides differ.
- fuse them when output shape is not identity.

Related: #113, #132

### DIFF
--- a/tico/passes/remove_redundant_permute.py
+++ b/tico/passes/remove_redundant_permute.py
@@ -24,6 +24,22 @@ from tico.serialize.circle_mapping import extract_shape, extract_stride
 from tico.utils import logging
 from tico.utils.passes import PassBase, PassResult
 from tico.utils.trace_decorators import trace_graph_diff_on_pass
+from tico.utils.validate_args_kwargs import PermuteArgs
+
+
+def _compose_permutation(dims1: list[int], dims2: list[int]):
+    """
+    Compose two permutation vectors.
+
+    Given y = x.permute(dims1) and z = y.permute(dims2),
+    the overall permutation p = dims2 ∘ dims1 is
+
+        p[i] = dims1[dims2[i]]
+    """
+    assert len(dims1) == len(
+        dims2
+    ), f"len(dims1): {len(dims1)}, len(dims2): {len(dims2)}"
+    return [dims1[i] for i in dims2]
 
 
 def passes():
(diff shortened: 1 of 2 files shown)


tico/passes/remove_redundant_permute.py
```diff
@@ -61,39 +81,37 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
...
+
+            fused_dims = _compose_permutation(permute1_dims, permute2_dims)
+            identity = list(range(len(fused_dims)))
+
+            if fused_dims == identity:
+                # shape
+                permute1_input_shape = extract_shape(permute1_input)
+                permute2_shape = extract_shape(permute2)
+                if permute1_input_shape != permute2_shape:
+                    continue
```
jinevening, line 102: Is this case possible?

mhs4670go: Ah, it is from old condition. Seems that it can be just an assertion.

jinevening (approved):
LGTM

tico/passes/remove_redundant_permute.py
```diff
@@ -61,39 +81,36 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
...
-
-            permute2.replace_all_uses_with(permute1_input, propagate_meta=False)
-
+            permute1_args = PermuteArgs(*permute1.args, **permute1.kwargs)  # type: ignore[arg-type]
+            permute1_input, permute1_dims = permute1_args.input, permute1_args.dims
+
+            fused_dims = _compose_permutation(permute1_dims, permute2_dims)
+            identity = list(range(len(fused_dims)))
+
+            if fused_dims == identity:
```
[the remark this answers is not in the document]
mhs4670go: If two consecutive permute ops result in identity tensor, they can be removed. Could you elaborate which part is hard to understand?

tico/passes/remove_redundant_permute.py
```diff
@@ -45,9 +61,13 @@ def __init__(self):
...
         """
         [BEFORE]
-            (AxBxC) - aten.permute - aten.permute - (AxBxC)
+            (AxBxC) - aten.permute_1 - aten.permute_2 - (OUT_SHAPE)
         [AFTER]
-            (AxBxC)
+            if OUT_SHAPE == (AxBxC):
+                (AxBxC)
+            else:
+                (AxBxC) - aten.permute (fused dims) - (OUT_SHAPE)
```
Seunghui98, line 69: Why don't we add the case where `permute_1` and `permute_2` are removed to the explanation? 
I think it can be misunderstood as if it exists only in the case of fused.

Issue #134, opened 2025-06-09
[pass] Fuse leading unsqueeze reshape


mhs4670go:
## What

Let's fuse “reshape → permute → reshape (leading unsqueeze)” into a single “reshape → permute”.

In many exported graphs we find the pattern
```
x                                 # e.g. (20, 1, 512)
 └─ reshape  -> (20,  8, 64)      # split last dim
     └─ permute -> ( 8, 20, 64)   # reorder
         └─ reshape -> (1, 8, 20, 64)   # **only adds leading 1-dims**
```
The final reshape does nothing but prepend one-sized (broadcast) dimensions, i.e. an unsqueeze that could have been introduced before the permute.

Eliminating it reduces kernel launches and makes later pattern-matching cleaner.

### Scope & Definition of “leading unsqueeze”

- 2nd reshape’s target shape must equal `[1, *shape_after_permute]`.
- No other size changes are allowed.
- Strides must remain valid (all involved ops are view-like).

![Image](https://github.com/user-attachments/assets/9e9fc8cd-d570-4300-b695-4ce44a7db74c)

PR #135, opened 2025-06-10
[pass] Introduce FuseLeadingUnsqueezeReshape pass

mhs4670go:
This commit introduces FuseLeadingUnsqueezeReshape pass.

Related: #113, #134


tico/passes/fuse_leading_unsqueeze_reshape.py
```diff
@@ -0,0 +1,107 @@
...
+                continue
+            permute_args = PermuteArgs(*permute.args, **permute.kwargs)  # type: ignore[arg-type]
+            reshape_front, permute_dims = permute_args.input, permute_args.dims
+
+            if not is_single_use_target_node(reshape_front, ops.aten.reshape):
+                continue
+            reshape_front_args = ReshapeArgs(*reshape_front.args, **reshape_front.kwargs)  # type: ignore[arg-type]
+            reshape_front_input, reshape_front_size = (
+                reshape_front_args.input,
+                reshape_front_args.size,
```
dayo09, line 73: `size` means `shape` here. I don't know why `ReshapeArgs` renamed its `shape` to `size`

mhs4670go: Good point. It should be aligned with the spec:)

tico/passes/fuse_leading_unsqueeze_reshape.py
```diff
@@ -0,0 +1,107 @@
...
+
+            # ---- condition: only leading unsqueeze ------------------------
+            back_shape = extract_shape(reshape_back)
+            permute_shape = extract_shape(permute)
+
+            if not _is_leading_unsqueeze(back_shape, permute_shape):
+                continue
+
+            # ---- create new reshape & new permute -------------------------
+            k = len(back_shape) - len(permute_shape)
```
dayo09, line 84: How about adding a test for k >= 2 case?

tico/passes/fuse_leading_unsqueeze_reshape.py
```diff
@@ -0,0 +1,107 @@
...
+    """
+
+    def call(self, ep: ExportedProgram) -> PassResult:
+        logger = logging.getLogger(__name__)
+
+        gm = ep.graph_module
+        graph = gm.graph
+        modified = False
+        for reshape_back in graph.nodes:
+            if not is_single_use_target_node(reshape_back, ops.aten.reshape):
```
jinevening, line 58: Why should this node have a single use?

tico/utils/utils.py
```diff
@@ -378,3 +378,31 @@ def broadcastable(
         if dim_a != 1 and dim_b != 1 and dim_a != dim_b:
             return False
     return True
+
+
+def is_single_use_target_node(
```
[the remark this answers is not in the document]
mhs4670go: Well, as you can see in the other passes, these conditions are very common. I guess we can check them in the single function.

tico/passes/fuse_leading_unsqueeze_reshape.py
```diff
@@ -0,0 +1,107 @@
...
+@trace_graph_diff_on_pass
+class FuseLeadingUnsqueezeReshape(PassBase):
+    """
+    Fuse reshape → permute → reshape where the second reshape only
+    prepends one-sized dims (unsqueeze) to the permuted tensor.
+
+    [BEFORE]
+        x - aten.reshape(s1) - aten.permute(p) - aten.reshape([1]*k + p(s1))
+    [AFTER]
+        x - aten.reshape([1]*k + s1) - aten.permute([d+k for d in p])
```
jinevening, line 48: Maybe?
```suggestion
        x - aten.reshape([1]*k + s1) - aten.permute(list(range(k)) + [d+k for d in p])
```

tico/passes/fuse_leading_unsqueeze_reshape.py
jinevening: I think other nodes don't have to have a single use.

dayo09 (approved):
LGTM

PR #139, opened 2025-06-11
Add (-32767,32767) qmin/qmax for int16

jinevening:
This adds (-32767,32767) qmin/qmax for int16.

---
To support quest framework


mhs4670go (approved):
LGTM

Issue #145, opened 2025-06-13
Add create_node helper for proper meta propagation & delete FillMetaVal pass


mhs4670go:
## Motivation

We now embed the originating torch module information inside every node via `meta["nn_module_stack"]` so that exported Circle nodes carry meaningful names. However, when optimization / decomposition passes insert fresh nodes (e.g. splitting `aten.layer_norm` into primitive ops) metadata is lost, leaving "unknown" in the final IR. We can propagate each meta every time we create the node. But, it would make duplicate codes and could be missed sometimes.

And, when it comes to the meta propagation, the `FillMetaVal` pass exists solely to patch `meta["val"]` after the node creations.

But keeping a separate pass + ad-hoc meta propagation in each optimization is brittle. Therefore, we'd better to have this wrapper to help node creations with expected meta propagation.

### Spec

```python
def create_node(
    graph: torch.fx.Graph,
    target: torch._ops.OpOverload,
    args: Tuple[Any, ...] = (),
    kwargs: Dict[str, Any] | None = None,
    *,
    origin: torch.fx.Node | None = None,
) -> torch.fx.Node:
  # Create node
  # Propagate origin's meta (except meta["val"])
  # Fill meta values (what `FillMetaVal` does)
```

### Before

```python
with gm.graph.inserting_before(node):
    packed = graph.call_function(
        torch.ops.aten.reshape.default, (x, pack_shape)
    )
```

### After

```python
with gm.graph.inserting_before(node):
    packed = create_node(
        graph, torch.ops.aten.reshape.default, (x, pack_shape), origin=node
    )
```

![Image](https://github.com/user-attachments/assets/be9ed4db-a868-4620-85f0-b57fb835376b)

PR #147, opened 2025-06-16
Introduce create_node helper

mhs4670go:
This commit introduces create_node helper for proper meta propagation.

Related: #145 

### DIFF
--- a/tico/passes/convert_to_relu6.py
+++ b/tico/passes/convert_to_relu6.py
@@ -20,6 +20,7 @@ import torch
 from torch.export import ExportedProgram
 
 from tico.utils import logging
+from tico.utils.graph import create_node
 from tico.utils.passes import PassBase, PassResult
 from tico.utils.trace_decorators import trace_graph_diff_on_pass
 from tico.utils.validate_args_kwargs import ClampArgs, HardTanhArgs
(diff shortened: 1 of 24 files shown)


tico/passes/convert_to_relu6.py
```diff
@@ -58,7 +59,7 @@ def convert(self, exported_program, node):
         input = args.input
 
         with graph.inserting_after(node):
-            relu_node = graph.call_function(torch.ops.aten.relu6.default, args=(input,))
+            relu_node = create_node(graph, torch.ops.aten.relu6.default, args=(input,))
```
dayo09, line 62: For some nodes, origins are missing. Do you mean it?

mhs4670go: Ah, right. I intentionally didn't set `origin` because those are propagated with `replace_all_uses_with`.

tico/passes/lower_to_slice.py
```diff
@@ -103,15 +103,21 @@ def call(self, exported_program: ExportedProgram) -> PassResult:
...
                 reshape_args = (slice_node, list(node_shape))
-                reshape_node = graph.call_function(
-                    torch.ops.aten.reshape.default, args=reshape_args
+                reshape_node = create_node(
+                    graph,
+                    torch.ops.aten.reshape.default,
+                    args=reshape_args,
+                    origin=node,
                 )
                 node.replace_all_uses_with(reshape_node, propagate_meta=False)
```
dayo09, line 122: I don't remember why it had to be `propagate_meta=False`...

PR #155, opened 2025-06-17
Fix mising condition of RemoveRedundantToCopy pass

mhs4670go:
This commit considers memory_format in RemoveRedundantToCopy pass and fix related test.

We are supporting only contiguous format now.

### DIFF
--- a/test/unit_test/pass_test/test_remove_redundant_to_copy.py
+++ b/test/unit_test/pass_test/test_remove_redundant_to_copy.py
@@ -54,9 +54,9 @@ class NonRedundantToCopyNet(torch.nn.Module):
 
 
 class NonRedundantToCopyTest(SinglePassValueTest):
-    def test_pass(self):
+    def test_pass_neg(self):
         self.setup(NonRedundantToCopyNet())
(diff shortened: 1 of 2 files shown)


test/unit_test/pass_test/test_remove_redundant_to_copy.py
```diff
@@ -54,9 +54,9 @@ def get_example_inputs(self):
...
 
 class NonRedundantToCopyTest(SinglePassValueTest):
-    def test_pass(self):
+    def test_pass_neg(self):
         self.setup(NonRedundantToCopyNet())
         self.assertEqual(num_of_ops(self.exported_program(), ops.aten.to_copy), 1)
 
         self.run_value_test(RemoveRedundantToCopy())
-        self.assertEqual(num_of_ops(self.exported_program(), ops.aten.to_copy), 0)
+        self.assertEqual(num_of_ops(self.exported_program(), ops.aten.to_copy), 1)
```
Seunghui98, line 62: After adding this condition for the pass, I see that this test was changed to a negative test.
so, I guess the number of `ops.aten.to_copy` operations remained unchanged (1 → 1).
Is my understanding correct? 🤔

Issue #165, opened 2025-06-20
Support fake-quantized NMT


jinevening:
## What

Let's convert fake-quantized NMT (PyTorch module) to quantized circle.

PR #166, opened 2025-06-20
Support Add, Relu in InsertQuantizeOpOnDtypeMismatch

jinevening:
This adds Add, Relu in InsertQuantizeOpOnDtypeMismatch.

---
Related to: #165

### DIFF
--- a/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
+++ b/tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
@@ -237,6 +237,9 @@ class InsertQuantizeOnDtypeMismatch(PassBase):
                 if qparam_dtype(x) == qparam_dtype(node):
                     continue
 
+                if qparam_dtype(x) != qparam_dtype(y):
+                    continue
+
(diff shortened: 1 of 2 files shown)


mhs4670go (approved):
LGTM

tico/experimental/quantization/passes/insert_quantize_on_dtype_mismatch.py
```diff
@@ -210,6 +212,39 @@ def _insert_quantize_op_after(node):
...
+                    continue
+                if QPARAM_KEY not in node.meta:
+                    continue
+
+                if qparam_dtype(x) == qparam_dtype(node):
+                    continue
+
+                if qparam_dtype(x) == "int16" and qparam_dtype(node) == "uint8":
+                    quantize = _insert_quantize_op_after(node)
+
```
dayo09, line 242: Is it okay without checking qparam_dtype(y)? What if one of x and y get folded in another pass and dtype gets differ?

jinevening: Sounds reasonable. I will add a check for that case.

PR #177, opened 2025-06-27
Fix conv with SAME padding

mhs4670go:
This commit fixes the logic when it has SAME padding.

### DIFF
--- a/tico/serialize/operators/op_conv2d.py
+++ b/tico/serialize/operators/op_conv2d.py
@@ -120,10 +120,10 @@ class Conv2dVisitor(NodeVisitor):
         conv_input: torch.fx.node.Node | circle.Tensor.TensorT = input_
         weight_shape = list(extract_shape(weight))
 
-        if is_same_padding(padding, input_shape, output_shape) and stride == [1, 1]:
-            conv2d_padding_type = SAME
-        elif is_valid_padding(padding):
+        if is_valid_padding(padding):
             conv2d_padding_type = VALID
+        elif is_same_padding(padding, input_shape, output_shape) and stride == [1, 1]:
+            conv2d_padding_type = SAME
         else:
             assert isinstance(padding, list) and len(padding) == 2
 
--- a/tico/serialize/operators/op_conv2d.py
+++ b/tico/serialize/operators/op_conv2d.py
@@ -120,7 +120,7 @@ class Conv2dVisitor(NodeVisitor):
         conv_input: torch.fx.node.Node | circle.Tensor.TensorT = input_
         weight_shape = list(extract_shape(weight))
 
-        if is_same_padding(padding, input_shape, output_shape):
+        if is_same_padding(padding, input_shape, output_shape) and stride == [1, 1]:
             conv2d_padding_type = SAME
         elif is_valid_padding(padding):
             conv2d_padding_type = VALID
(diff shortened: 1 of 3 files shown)


tico/serialize/operators/op_conv2d.py
```diff
@@ -120,10 +120,10 @@ def define_node(self, node: torch.fx.Node) -> circle.Operator.OperatorT:
         conv_input: torch.fx.node.Node | circle.Tensor.TensorT = input_
         weight_shape = list(extract_shape(weight))
 
-        if is_valid_padding(padding):
-            conv2d_padding_type = VALID
-        elif is_same_padding(padding, input_shape, output_shape):
+        if is_same_padding(padding, input_shape, output_shape):
```
mhs4670go, line 123: The same padding check should be done first.

glistening:
Please fix commit title typo.

tico/serialize/operators/op_conv2d.py
glistening: (Out of this PR's scope)

I wonder why `is_same_padding` needs `input_shape` and `output_shape` while `is_valid_padding` accepts only padding, and check whether padding type is "valid". I took a look the definition `is_same_padding`, but I don't understand why it checks input_HW == output_HW in `is_same_padding`.

mhs4670go: As I said in the offline, the condition must check if `stride` is 1. But, current function spec is hard to apply this condition. I'll refactor this function in the next PR.

tico/serialize/operators/op_conv2d.py
```diff
@@ -120,7 +120,7 @@ def define_node(self, node: torch.fx.Node) -> circle.Operator.OperatorT:
         conv_input: torch.fx.node.Node | circle.Tensor.TensorT = input_
         weight_shape = list(extract_shape(weight))
 
-        if is_same_padding(padding, input_shape, output_shape):
+        if is_same_padding(padding, input_shape, output_shape) and stride == [1, 1]:
```
glistening, line 123: How about changing the order?
```py
if is_valid_padding(...):
  ...
elif is_same_padding(...):
```
If padding is specified as "valid", it should be "valid".
But if we check `is_same_padding` first, "valid" may happen to be guessed as "same".

tico/serialize/operators/op_conv2d.py
seockho-kim: > The same padding check should be done first.

Why same padding check should be done first?

tico/utils/padding.py
```diff
@@ -40,8 +40,8 @@ def is_same_padding(
     if isinstance(padding, list):
         assert len(padding) == 2, "Padding should be a list of length 2."
 
-        input_HW = input_shape[1:2]  # N H W C
-        output_HW = output_shape[1:2]  # N H W C
+        input_HW = tuple(input_shape[1:3])  # N H W C
+        output_HW = tuple(output_shape[1:3])  # N H W C
```
mhs4670go, line 44: Without this patch, `ConvWithSamePadding2` is exported like below because list == torch.Size comparison doesn't work well. They should be casted into same data types.

image

jinevening (approved):
LGTM

seockho-kim (approved):

PR #183, opened 2025-06-30
[test] Use copy of example_inputs for immutablity

dayo09:
During `forward` or `export`, example_inputs could change. To prevent it, let's use the copy of them.

---

From #173

### DIFF
--- a/test/pt2_to_circle_test/builder.py
+++ b/test/pt2_to_circle_test/builder.py
@@ -16,6 +16,7 @@ import contextlib
 import io
 import os
 import unittest
+from copy import deepcopy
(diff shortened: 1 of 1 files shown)


dayo09:
> When does this happen?

@mhs4670go 

While running a model with KVCache inputs.
After export/forward, it changes by appending more values.

mhs4670go:
Just for the reviewers, unlike typical input tensros like `input_ids`, `DynamicCache` in transformers is an in-place updated object.

mhs4670go (approved):
LGTM

seockho-kim (approved):
LGTM

PR #185, opened 2025-06-30
Print warning for training model

dayo09:
Let's print a warning when training model is given


tico/utils/convert.py
```diff
@@ -287,6 +287,12 @@ def convert(
     strict: bool = True,
     config: CompileConfigBase = get_default_config(),
 ) -> CircleModel:
+    if mod.training:
```
mhs4670go, line 290: Hmm.. the other day I met this model that didn't have `.eval()` method. Maybe it could happen with `.training` method. How about check if it has `.training` first?

mhs4670go (approved):
LGTM

mhs4670go (approved):

Issue #159, opened 2025-06-18
Reduce or Eliminate Warnings When Running ./ccex test


mhs4670go:
## What

Currently, running `./ccex test` results in a large number of warnings being printed to the console. While the tests themselves pass, the volume of warnings makes it difficult to identify actual issues and leads to noise in CI logs.

- Improve readability and clarity of test output
- Avoid masking real issues or regressions
- Ensure a clean and professional CI pipeline

## Expected Behavior

Running `./ccex test` should ideally produce no warnings, or only necessary and meaningful ones. If some warnings are expected and unavoidable, they should be documented or explicitly suppressed.

PR #189, opened 2025-07-01
[test] Use buffer

mhs4670go:
This commit uses buffer instead of in-place tensors.

Related: #159

This change turns off below warnings.
```
/home/seongwoo/TICO/.venv/lib/python3.10/site-packages/torch/fx/graph.py:1801: UserWarning: Node segm_index2_1 target segm_index2 segm_index2 of  does not reference an nn.Module, nn.Parameter, or buffer, which is what 'get_attr' Nodes typically target
  warnings.warn(
ok

```


dayo09 (approved):
LGTM

PR #200, opened 2025-07-08
[DRAFT] Install onert when running `ccex configure test`

seockho-kim:
This commit installs onert when you run `ccex configure test`. It allows to use `use_onert` tag.

ONE-DCO-1.0-Signed-off-by: Seockho Kim seockho.kim@samsung.com


seockho-kim:
https://github.com/Samsung/TICO/pull/201

Issue #179, opened 2025-06-27
Support bitnet-b1.58 model


seockho-kim:
## What

Let's support bitnet-b1.58 model

## Why

1bit LLM is known as fast and light model without big performance degradation

seockho-kim:
```
import torch
from transformers import AutoModelForCausalLM, AutoTokenizer
import tico

model_id = "microsoft/bitnet-b1.58-2B-4T"

def prepare_input():
    kwargs = {
        "input_ids": torch.randint(0, 10000, (1, 21), dtype=torch.int64),
        "attention_mask": torch.ones((1, 21), dtype=torch.int64),
        "position_ids": torch.arange(0, 21, dtype=torch.int64).unsqueeze(0),
        "use_cache": False,
        "logits_to_keep": 1,
    }
    return kwargs

# Load tokenizer and model
tokenizer = AutoTokenizer.from_pretrained(model_id)
model = AutoModelForCausalLM.from_pretrained(model_id)

tico.convert(model, (), prepare_input()).save("bitnet.circle")

```

seockho-kim:
With this draft https://github.com/Samsung/TICO/pull/186

```
flatbuffers.builder.BuilderSizeError: flatbuffers: cannot grow buffer beyond 2 gigabytes
```

seockho-kim:
Circles can be exported for each modules.

```
-rw-rw-r-- 1 seockho seockho 1.3G  7월  4 19:04 lm_head.circle
-rw-rw-r-- 1 seockho seockho 1.3G  7월  4 19:04 model.embed_tokens.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:04 model.layers.0.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.10.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.11.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.12.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.13.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.14.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.15.circle
-rw-rw-r-- 1 seockho seockho 266M  7월  4 19:05 model.layers.16.circle
[...]

PR #203, opened 2025-07-09
Fix the dtype of min and max of clamp

seockho-kim:
This commit fixes dtype of min and max value of clamp op. The dtype should be matched with the input dtype.

for https://github.com/Samsung/TICO/issues/179#issuecomment-3047972123

### DIFF
--- a/tico/serialize/operators/op_clamp.py
+++ b/tico/serialize/operators/op_clamp.py
@@ -27,6 +27,7 @@ from tico.serialize.circle_graph import (
     extract_circle_dtype,
     extract_shape,
 )
+from tico.serialize.circle_mapping import extract_torch_dtype
 from tico.serialize.operators.hashable_opcode import OpCode
 from tico.serialize.operators.node_visitor import NodeVisitor, register_node_visitor
 from tico.serialize.operators.utils import create_builtin_operator, get_op_index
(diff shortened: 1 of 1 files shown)


tico/serialize/operators/op_clamp.py
```diff
@@ -92,6 +93,13 @@ def define_node(
         min_val = args.min
         max_val = args.max
 
+        input_dtype = extract_torch_dtype(input)
```
mhs4670go, line 96: Seems that `clamp()` doesn't always follow the input dtype. How about casting them to the output dtype instead?

Plus, please add related clamp tests, too.

dayo09, line 96: Could you add a test for this, too?

mhs4670go: As we discussed in the offline, we should insert casting ops like `aten._to_copy` to cast the input node because it's not a constant tensor. Let's make a pass for this.

seockho-kim: As we talked in offline,
we need a pass to cast input,min,max to output's dtype when it's different one another.
so, let me submit another PR.

seockho-kim:
https://github.com/Samsung/TICO/pull/209

PR #209, opened 2025-07-11
Cast mixed type args of clamp

seockho-kim:
This pass ensures consistent dtypes for clamp operations by:
    1. Converting min/max arguments to match output dtype when provided
    2. Inserting cast operations when input dtype differs from output dtype

for https://github.com/Samsung/TICO/issues/179#issuecomment-3047972123

### DIFF
--- a/tico/passes/cast_clamp_mixed_type_args.py
+++ b/tico/passes/cast_clamp_mixed_type_args.py
@@ -19,6 +19,9 @@ if TYPE_CHECKING:
 import torch
 from torch.export import ExportedProgram
 
+from tico.passes import ops
+
+from tico.serialize.circle_mapping import extract_torch_dtype
 from tico.utils import logging
 from tico.utils.graph import create_node
 from tico.utils.passes import PassBase, PassResult
@@ -81,12 +84,8 @@ class CastClampMixedTypeArgs(PassBase):
         min = args.min
         max = args.max
 
-        if "val" not in input.meta or "val" not in node.meta:
-            logger.warning(f"Missing meta['val'] for node {node.name}")
-            return False
-
-        input_dtype = input.meta["val"].dtype
-        output_dtype = node.meta["val"].dtype
+        input_dtype = extract_torch_dtype(input)
+        output_dtype = extract_torch_dtype(node)
 
         def _convert_arg(arg, arg_name: str):
             if arg is None:
@@ -95,13 +94,14 @@ class CastClampMixedTypeArgs(PassBase):
             arg_dtype = torch.tensor(arg).dtype
             arg_idx = node.args.index(arg)
             if arg_dtype != output_dtype:
-                if output_dtype == torch.float32 or output_dtype == torch.float:
+                assert output_dtype in [torch.float, torch.int]
+                if output_dtype == torch.float:
                     arg = float(arg)
                 else:
                     arg = int(arg)
                 node.update_arg(arg_idx, arg)
                 logger.debug(
-                    f"Converted {arg_name} value from {arg_dtype} to {output_dtype} for clamp operation at {node.name}"
+                    f"Casting {arg_name} value from {arg_dtype} to {output_dtype} for clamp operation at {node.name}"
                 )
                 return True
             return False
@@ -129,7 +129,7 @@ class CastClampMixedTypeArgs(PassBase):
         return modified
 
     def call(self, exported_program: ExportedProgram) -> PassResult:
-        target_op = [torch.ops.aten.clamp.default, torch.ops.aten.clamp.Tensor]
+        target_op = ops.aten.clamp
 
         graph_module = exported_program.graph_module
         graph = graph_module.graph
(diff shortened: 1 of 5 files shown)


tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+        # clamp(Tensor self, Scalar? min=None, Scalar? max=None) -> Tensor
+        args = ClampArgs(*node.args, **node.kwargs)  # type: ignore[arg-type]
+
+        input = args.input
+        min = args.min
+        max = args.max
+
+        if "val" not in input.meta or "val" not in node.meta:
+            logger.warning(f"Missing meta['val'] for node {node.name}")
+            return False
```
mhs4670go, line 86: Just curiosity, is there any case that meets this condition? AFAIK, all fx nodes have `meta["val"]` at this point.

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+
+        input_dtype = input.meta["val"].dtype
+        output_dtype = node.meta["val"].dtype
```
mhs4670go, line 89: You can use `extract_torch_dtype` function instead. (tico/serialize/circle_mapping.py)

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+                )
+                set_new_meta_val(to_copy)
+                node.update_arg(node.args.index(input), to_copy)
+
+            modified = True
+
+        return modified
+
+    def call(self, exported_program: ExportedProgram) -> PassResult:
+        target_op = [torch.ops.aten.clamp.default, torch.ops.aten.clamp.Tensor]
```
mhs4670go, line 132: This can be from `tico/passes/ops.py` module.

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+            arg_dtype = torch.tensor(arg).dtype
+            arg_idx = node.args.index(arg)
+            if arg_dtype != output_dtype:
+                if output_dtype == torch.float32 or output_dtype == torch.float:
+                    arg = float(arg)
+                else:
+                    arg = int(arg)
+                node.update_arg(arg_idx, arg)
+                logger.debug(
+                    f"Converted {arg_name} value from {arg_dtype} to {output_dtype} for clamp operation at {node.name}"
```
mhs4670go, line 104: (Optional) For debugging purpose, `Cast` seems more appropriate.

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+        output_dtype = node.meta["val"].dtype
+
+        def _convert_arg(arg, arg_name: str):
+            if arg is None:
+                return False
+
+            arg_dtype = torch.tensor(arg).dtype
+            arg_idx = node.args.index(arg)
+            if arg_dtype != output_dtype:
+                if output_dtype == torch.float32 or output_dtype == torch.float:
```
mhs4670go, line 98: Both are same. Just it is an alias.
```suggestion
                if output_dtype == torch.float:
```

mhs4670go, line 98: I think adding an assertion that checks supported output dtypes would be helpful for later extension. Because below logic considers only two dtypes for now.
```suggestion
            if arg_dtype != output_dtype:
                assert output_dtype in [torch.float, torch.int]
                if output_dtype == torch.float32 or output_dtype == torch.float:
```

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+                    arg = float(arg)
+                else:
+                    arg = int(arg)
```
mhs4670go, line 101: Actually, this line isn't necessary because. But, seems okay for safety.

tico/passes/cast_clamp_mixed_type_args.py
seockho-kim: Thanks, I didn't know it's an alias. :)

seockho-kim: Okay, I agree.

tico/passes/cast_clamp_mixed_type_args.py
seockho-kim: I'll keep it, for safety and clarification.

mhs4670go (approved):
LGTM

tico/passes/cast_clamp_mixed_type_args.py
```diff
@@ -0,0 +1,147 @@
...
+    Behavior Examples:
+    - When input dtype differs from output:
+        Inserts _to_copy operation to convert input
+    - When min/max dtype differs from output:
+        Converts min/max values to output dtype
+
+    [before, if input dtype is different from output dtype]
+
+            input               min(or max)
+           (dtype=int)         (dtype=float)
```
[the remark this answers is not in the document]
seockho-kim: Case 1 and 2 are covered in the code.
The description makes some misunderstandings, I'll update it.

dayo09 (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #89, opened 2025-05-13
Introduce PTQWrapper class for calibration + export with multiple PTQ algorithm


mhs4670go:
### Motivation

Quantizing a pretrained model in PyTorch typically involves hand-crafting per-layer calibration loops, duplicating code for each observer algorithm, and juggling separate state for min/max, histograms, or reconstruction parameters. A unified `PTQWrapper` solves these pain points by:

- Encapsulating each layer’s calibration logic, fake-quant calls, and parameter storage in a single module  
- Allowing multiple observer-based algorithms to collect statistics in one pass and be switched at will for head-to-head comparison  
- Centralizing `state_dict` management so you can checkpoint, resume, and export a fully instrumented model with minimal boilerplate  

By introducing `PTQWrapper`, we streamline end-to-end PTQ workflows, reduce code duplication, and ensure reproducible, exportable quantization pipelines.  

Below is a suggested design for a flexible, custom PTQ wrapper module in PyTorch—complete with considerations, a sketch of the API/architecture, and a concrete code example.

### Design Overview

```
       +-------------------+
       |   FP32 Model      |
       +-------------------+
                 |
                 | prepare(model, configs)
                 v
       +-------------------+
       | Wrapped Model     |
       | (PTQWrapper)      |
       +-------------------+
                 |
                 | run_fn(calibration data)
                 v
       +-------------------+
       | Calibration Pass: |
       | • minmax observer |
       | • percentile obs. |
       +-------------------+
                 |
                 | compute_qparams()
                 v
       +-------------------+
       | QParams Stored    |
       | (scale / zero‐pt) |
       +-------------------+
                 |
                 | set_active_algorithm('minmax')  
                 |   or ('percentile'), etc.
                 v
       +-------------------+
       | Quantized Forward |
       | • fake‐quantize   |
       +-------------------+
                 |
                 | convert(..., algorithm='minmax')
                 v
       +-------------------+
       | Exportable Graph  |
       | (torch.export)    |
       +-------------------+
```

#### PTQWrapper

- **Wraps any** `nn.Module` (e.g. `Conv2d`, `Linear`, etc.)  
- **Holds one or more** Observer/Algorithm objects responsible for:
  - Collecting statistics (min/max, histograms, percentiles, …)  
  - Computing and holding quantization parameters (scale, zero-point, bitwidth)

#### Observer/Algorithm Interface

- **`collect(x)`**  
  - Called during calibration mode to update running stats.  
[...]

PR #212, opened 2025-07-15
[DRAFT] Introduce PTQWrapper

mhs4670go:
This draft introduces PTQWrapper.

Related: #89

### DIFF
--- a/test/quantization/ptq/observers/test_histogram.py
+++ b/test/quantization/ptq/observers/test_histogram.py
@@ -1,0 +1,110 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+
+from tico.experimental.quantization.ptq.dtypes import UINT8
+from tico.experimental.quantization.ptq.observers.histogram import HistogramObserver
+
+
+class TestHistogramObserver(unittest.TestCase):
+    """Sanity-checks for the patched HistogramObserver."""
+
+    # ------------------------------------------------------------------ #
+    # Helper
+    # ------------------------------------------------------------------ #
+    @staticmethod
+    def _make_obs(**kwargs):
+        """Factory for a fresh observer with optional ctor kwargs."""
+        return HistogramObserver(name="hist", dtype=UINT8, **kwargs)
+
+    # ------------------------------------------------------------------ #
+    # 1. Histogram accumulation
+    # ------------------------------------------------------------------ #
+    def test_histogram_accumulation(self):
+        obs = self._make_obs(bins=256)
+        x = torch.randn(4, 16)
+        obs.collect(x)
+
+        self.assertAlmostEqual(
+            obs.hist.sum().item(),
+            x.numel(),
+            delta=1e-3,
+            msg="Histogram counts must equal number of collected samples",
+        )
+        self.assertFalse(torch.isinf(obs.min_val))
+        self.assertFalse(torch.isinf(obs.max_val))
+
+    # ------------------------------------------------------------------ #
+    # 2. Q-param computation
+    # ------------------------------------------------------------------ #
+    def test_compute_qparams_sets_scale_and_zp(self):
+        obs = self._make_obs(bins=512)
+        obs.collect(torch.randn(8, 32))
+        scale, zp = obs.compute_qparams()
+
+        self.assertTrue(obs.has_qparams)
+        self.assertGreater(scale.item(), 0.0)
+        self.assertIsInstance(zp.item(), int)
+
+    # ------------------------------------------------------------------ #
+    # 3. fake_quant clamps properly
+    # ------------------------------------------------------------------ #
+    def test_fake_quant_within_clip(self):
+        obs = self._make_obs(bins=512)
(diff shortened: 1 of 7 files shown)


tico/experimental/quantization/algorithm/ptq/examples/quantize_lienar.py
```diff
@@ -0,0 +1,108 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
```
dayo09, line 1: Typo in the filename: quantize_lienar -> quantize_linear

tico/experimental/quantization/algorithm/ptq/quant_config.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+
+    default_dtype: DType = DType.uint(8)
+    default_observer: Type[ObserverBase] = MinMaxObserver
+    default_qscheme: QScheme = QScheme.PER_TENSOR_AFFINE
+    overrides: Mapping[str, Mapping[str, Any]] = field(default_factory=dict)
+
+    def get_kwargs(self, obs_name: str) -> Dict[str, Any]:
+        """
+        Return user-specified kwargs for *obs_name* inside **this** wrapper.
```
dayo09, line 80: Q. do you use specific editor to parse this markdown-style enhancements? Actually it's hard to read in my plain vscode editor...

tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
```diff
@@ -0,0 +1,161 @@
...
+# -------------------------------------------------------------------------
+print("Applying GPTQ …")
+dataset = load_dataset("wikiText", "wikitext-2-raw-v1", split="test")
+q_m = prepare(model, GPTQConfig(), inplace=True)
+
+for d in tqdm.tqdm(dataset, desc="GPTQ calibration"):
+    ids = tokenizer(d["text"], return_tensors="pt").input_ids.to(DEVICE)
+    q_m(ids)  # observers gather weight stats
+
+q_m = convert(q_m, inplace=True)  # materialize INT-weight tensors
```
dayo09, line 110: It's great use of tqdm above but when running this script, `q_m = convert(q_m, inplace=True)` takes more time.
Could you add tqdm to quantizer.py too?

```py
        quantizers: Dict[str, Any] = {}
        for l_idx, layer in enumerate(tqdm.tqdm(target_layers)): # <-- to track the conversion in terminal.
```

tico/experimental/quantization/algorithm/ptq/examples/compare_ppl.py
```diff
@@ -0,0 +1,124 @@
...
+
+# ── local quantization helpers ────────────────────────────────────────────────
+from tico.experimental.quantization.algorithm.ptq.quant_config import QuantConfig
+from tico.experimental.quantization.algorithm.ptq.utils import perplexity
+from tico.experimental.quantization.algorithm.ptq.wrappers.ptq_wrapper import PTQWrapper
+
+# -------------------------------------------------------------------------
+# 0. Global configuration
+# -------------------------------------------------------------------------
+MODEL_NAME = "meta-llama/Meta-Llama-3-1B"
```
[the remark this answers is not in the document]
mhs4670go: I just wrote it down for the example purpose because it is gated model. You should replace that string with valid ones.

tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
mhs4670go: Yeah, I think so. I kept thinking about it and ketp putting it off, lol. Thanks for the reivew:)

tico/experimental/quantization/algorithm/ptq/quant_config.py
mhs4670go: But, let's have some more discussion about preferences in the near future.

tico/experimental/quantization/ptq/wrappers/quant_elementwise.py
```diff
@@ -0,0 +1,83 @@
...
+class QuantElementwise(QuantModuleBase):
+    """
+    Generic wrapper for *any* 1-to-1 element-wise op `y = f(x)`.
+
+    Sub-classes only need to implement:
+        • `FUNC`: a Callable that maps tensor→tensor
+    """
+
+    # subclass must set this
+    FUNC: Callable[[torch.Tensor], torch.Tensor] = staticmethod(lambda x: x)
```
dayo09, line 36: If it's must be set, why not "throw Exception"?

mhs4670go:
> In case of masking, some channel io could all be set as -inf. In the case, will this ema collect logic work properly?

@dayo09 Good catch. Therefore, the mask is created with -120 value.

tico/experimental/quantization/ptq/wrappers/quant_elementwise.py
mhs4670go: I've added `__init_subclass__` for this. Thanks.

tico/experimental/quantization/ptq/wrappers/quant_module_base.py
```diff
@@ -0,0 +1,153 @@
...
+            obs.collect(x.detach())
+            return x
+        if self._mode is Mode.QUANT:
+            return obs.fake_quant(x)
+        return x  # NO_QUANT
+
+    @abstractmethod
+    def _all_observers(self) -> Iterable[ObserverBase]:
+        """Return every observer owned by this module."""
+        ...
```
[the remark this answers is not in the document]
mhs4670go: Technically speaking, observers are attached to operations rather than modules. Wrapped modules run operations, which means they has multiple observers.

`.children()` only iterates over sub-modules. `_all_observers` iterates not only such modules' observers but also own operations' observers. Therefore, only calling child's observers wouldn't be enough. 

Below is the exmaple.
```python
# QuantLlamaMLP
def _all_observers(self):
        # local first
        yield self.act_in_obs
        yield self.mul_obs
        # recurse into children that are QuantModuleBase
        for m in (self.gate_proj, self.up_proj, self.down_proj, self.act_fn):
            yield from m._all_observers()
```

PR #214, opened 2025-07-16
Support dynamic shape

dayo09:
Let's add dynamic shape example and test using onert.

### DIFF
--- a/test/modules/op/depthwise_conv2d.py
+++ b/test/modules/op/depthwise_conv2d.py
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 import torch
+from torch.export import Dim
+
+from test.utils import tag
 
 
 class SimpleDepthwiseConv(torch.nn.Module):
@@ -27,6 +30,28 @@ class SimpleDepthwiseConv(torch.nn.Module):
 
     def get_example_inputs(self):
         return (torch.randn(1, 8, 64, 64),)
+
+
+@tag.use_onert
+class SimpleDepthwiseConvDynamicShape(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.conv2d = torch.nn.Conv2d(
+            in_channels=8, out_channels=8, kernel_size=3, groups=8, padding=(2, 2)
+        )
+
+    def forward(self, input):
+        return self.conv2d(input)
+
+    def get_example_inputs(self):
+        return (torch.randn(4, 8, 64, 64),)
+
+    def get_dynamic_shapes(self):
+        batch = Dim("batch", min=1, max=128)
+        dynamic_shapes = {
+            "input": {0: batch},
+        }
+        return dynamic_shapes
 
 
 class SimpleDepthwiseConvWithValidPaddingInStr(torch.nn.Module):
--- a/tico/serialize/circle_graph.py
+++ b/tico/serialize/circle_graph.py
@@ -250,6 +250,15 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         if source_node is not None:
             self.name_to_node[tensor.name] = source_node
         tensor.shape = shape
+
+        # Handle dynamic shape
+        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
+            tensor.shapeSignature = tensor.shape.copy()
+            for idx, s in enumerate(tensor.shape):
+                if isinstance(s, torch.SymInt):
+                    tensor.shape[idx] = 1
+                    tensor.shapeSignature[idx] = -1
+
         if qparam is not None:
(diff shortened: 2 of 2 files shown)


test/modules/net/mlp_dyn.py
```diff
@@ -0,0 +1,45 @@
...
+        )
+        return down_proj
+
+    def get_example_inputs(self):
+        # (B, Seq_Len, Dim)
+        return (torch.randn(B, SEQ_LEN, DIM),)
+
+    def get_dynamic_shapes(self):
+        batch = Dim("batch", min = 1, max = 128)
+        dynamic_shapes = {"x": {0: batch},}
```
dayo09, line 43: ### Exported Circle Model ✔️ 

image

test/pt2_to_circle_test/builder.py
```diff
@@ -150,7 +150,7 @@ def _run(
         if without_inference:
             return
 
-        USE_ONERT = os.environ.get("CCEX_RUNTIME") == "onert" or dynamic
+        USE_ONERT = os.environ.get("CCEX_RUNTIME") == "onert"
```
dayo09, line 153: Let's set `use_onert` explicitly

tico/utils/dtype.py
```diff
@@ -0,0 +1,20 @@
...
+    np.dtype("int64"): torch.int64,
+    np.dtype("int32"): torch.int32,
+    np.dtype("int16"): torch.int16,
+    np.dtype("int8"): torch.int8,
+    np.dtype("uint8"): torch.uint8,
+    np.dtype("bool"): torch.bool,
+}
+
+
+def numpy_dtype_to_torch_dtype(np_dtype):
```
jinevening, line 19: Could you add type hints?

```suggestion
def numpy_dtype_to_torch_dtype(np_dtype):
```

tico/utils/dtype.py
```diff
@@ -0,0 +1,20 @@
...
+    return NUMPY_TO_TORCH_DTYPE_DICT[np.dtype(np_dtype)]
```
jinevening, line 20: Isn't it possible to just pass np_dtype?

```suggestion
    return NUMPY_TO_TORCH_DTYPE_DICT[np_dtype]
```

test/modules/op/depthwise_conv2d.py
```diff
@@ -29,6 +32,28 @@ def get_example_inputs(self):
         return (torch.randn(1, 8, 64, 64),)
 
 
+@tag.use_onert
+class SimpleDepthwiseConvDynamicShape(torch.nn.Module):
```
dayo09, line 36: @jinevening Thanks for the comment! I added the logic to add_tensor_from_scratch.

image

This test ensures that any intermediate tensor made by `add_tensor_from_scratch` has a dynamic shape - padding in this case.

test/pt2_to_circle_test/builder.py
```diff
@@ -187,15 +187,6 @@ def build(self, submodule):
...
-            testdict[base_name] = NNModuleTest(base_name, module_instance).make(
-                dynamic=False
-            )
-
-            if hasattr(module_instance, "get_dynamic_shapes"):
-                dyn_name = base_name + "_dynamic"
-                testdict[dyn_name] = NNModuleTest(dyn_name, module_instance).make(
-                    dynamic=True
-                )
+            testdict[base_name] = NNModuleTest(base_name, module_instance).make()
```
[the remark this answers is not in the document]
mhs4670go: @dayo09 The reason why I branched was because I wanted to reuse the existing tests. For example, if I introduce a `get_dynamic_shapes` to existing tests, the test would be run twice: one with `get_example_inputs` and the other with `get_dynamic_shapes`. 

But, with current approach, `SimpleDepthwiseConvDynamicShape` seems to be run only once — with `get_dynamic_shapes`. Is it what you intended?

tico/serialize/circle_graph.py
```diff
@@ -241,6 +250,15 @@ def add_tensor_from_scratch(
         if source_node is not None:
             self.name_to_node[tensor.name] = source_node
         tensor.shape = shape
+
+        # Handle dynamic shape
+        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
```
jinevening, line 255: The type hints and comments of this function have to be updated as well.

```python
        shape: List[int],
```

```
        shape : List[int]
            The shape of the tensor.
```

jinevening (approved):
LGTM w/ minor comment

tico/serialize/circle_graph.py
dayo09: Let me revisit there in the next PR!

PR #208, opened 2025-07-10
[test] Add LlamaPrefill using captured input

glistening:
It captures the target model input from user input (e.g. prompt). It helps us to prepare example input in high level.

TICO-DCO-1.0-Signed-off-by: Sanggyu Lee <sg5.lee@samsung.com>

```
./ccex test -m LlamaPrefill
open llama.prefill.circle
```

PR #215, opened 2025-07-17
[test] Add LlamaDecoderLayer using captured input

glistening:
It adds LlamaDecoderLayer test, which uses captured input.


glistening:
It is a few line modified version from #208.
It generates circle model for `LlamaDecoderLayer` while #208 generates the whole `LlamaModel` for prefill pass with padded user tokens.

glistening:
See #217.

PR #217, opened 2025-07-17
[test+operators] Fuse attention to circle attention

glistening:
It introduces circle_attention op, and add tests which fuse attention from LlamaDecoderLayers.

It is another fuser. It does not use pattern match, but it adds an operator using decorator.
You can run the code using

```
python test/modules/model/LlamaDecoderLayerWithKVCache/model.py
```

### DIFF
--- a/test/modules/model/LlamaDecoderLayerWithKVCacheAndFusedAttention/model.py
+++ b/test/modules/model/LlamaDecoderLayerWithKVCacheAndFusedAttention/model.py
@@ -1,0 +1,156 @@
+# User input
+prompt = "Lily picked up a flower."
+model_name = "Maykeye/TinyLLama-v0"
+
+captured_input = ()
+
+import copy, inspect, types
+
+from transformers.models.llama.modeling_llama import LlamaDecoderLayer
+
+forward_org = LlamaDecoderLayer.forward
+
+
+def capture_and_forward(self, *args, **kwargs):
+    global captured_input
+
+    # Prepare args tuple for TICO.convert()
+    # Get arg_names in positional args order using inspect
+    sig = inspect.signature(forward_org)
+    args_names = [
+        # signature includes `self`` and `kwargs``.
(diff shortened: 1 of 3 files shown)


glistening:
```
python test/modules/model/LlamaDecoderLayerWithKVCache/layer.py 
```

glistening:
## 7/28 Offline Code Review

- Assumption of 1 continuous k_cache instead of k_cache_0, k_cache_1, ... is not valid for quantized model, where each layer's k_cache has difference qparam. (Same for q_cache) I will keep this structure since it is for cpu, not npu.
- `remove_unused_input` may have side-effect if ExportedProgram uses the unused placeholder. (What side-effect? No one knows, but it is not required pass. I just want to make the graph tidy. I will not push `remove_unused_input` pass.
- `op_circle_attention` seems onert-specific. It will be put under `serialize/operators/onert` as `op_attention`.
- It is okay to put `record_input.py` since it does not break something before using this.
- New test case which uses `record_input.py` will use function unittest, instead of module unittest (which requires `get_example_inputs`).

tico/serialize/operators/onert/op_attention.py
```diff
@@ -0,0 +1,179 @@
...
+    attention_mask,
+    past_key,
+    past_value,
+    layer_idx,
+    cache_position,
+):
+    return hidden_states
+
+
+@torch.library.register_fake("circle::attention.llama")
```
glistening, line 73: Without this,

```
torch._dynamo.exc.Unsupported: Operator does not support running with fake tensors

  Developer debug context: unsupported operator: circle.attention.llama
```

glistening:
- Added decode.py for the whole model for decoding.
- I removed `attention_mask` from fused op_attention because:
  - I am thinking of running decode phase on cpu, not npu. Therefore I assume there is no such thing of pad token either left or right of actual tokens.

glistening:
#400

PR #221, opened 2025-07-18
Do not decompose aten.upsample_nearest2d

mhs4670go:
This commit adds an aten.upsample_nearest2d to preserve op list.

We don't have to check the version with this patch.


jinevening:
Ah, there is a conflict.

seockho-kim (approved):
LGTM

Seunghui98 (approved):
LGTM 👍

PR #229, opened 2025-07-22
Handle dynamic shape after serialize

dayo09:
Let's fix it to re-interpret the dynamic shape after serialization.

---

### WHY?

`add_tensor_from_node` is called when `tensors are exported to node before serialization`.
Then, each `operator serialization` is performed. The input of the operator could be circle-interpreted symbolic shape, which has lost its dynamic shape information. It will mislead **all the add_tensor_from_scratch operators used in operator visitor**.

For example,
```py
lass BasePowVisitor(NodeVisitor):
    def __init__(self, op_codes: Dict[OpCode, int], graph: CircleSubgraph):
        super().__init__(op_codes, graph)

    def cast_to_float(self, node: torch.fx.Node) -> circle.Tensor.TensorT:
        assert isinstance(node, torch.fx.Node), type(node)
        node_tensor: circle.Tensor.TensorT = self.graph.get_tensor(node)
        node_shape: List[int] = node_tensor.shape # <------- Already interpreted as circle shape. [s0, 3, 3, 4] -> [1, 3, 3, 4]
        op_index = get_op_index(
            circle.BuiltinOperator.BuiltinOperator.CAST, self._op_codes
        )
        cast_name = f"{node.name}_cast"
        cast_dtype = circle.TensorType.TensorType.FLOAT32
        cast_tensor = self.graph.add_tensor_from_scratch(
            prefix=cast_name,
            dtype=cast_dtype,
            shape=node_shape, # <----- [1, 3, 3, 4] is given. `cast_tensor` never know it is dynamic shape.
            source_node=node,
        )
```

Therefore, conversion of torch symbolic integer into circle dynamic shape schema must be performed at the end of `build_circle`.

### DIFF
--- a/tico/serialize/circle_graph.py
+++ b/tico/serialize/circle_graph.py
@@ -152,15 +152,6 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         assert node.meta.get("val") is not None
         tensor.type = extract_circle_dtype(node)
         tensor.shape = list(extract_shape(node))
-
-        # Handle dynamic shape
-        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
-            tensor.shapeSignature = tensor.shape.copy()
-            for idx, s in enumerate(tensor.shape):
-                if isinstance(s, torch.SymInt):
-                    tensor.shape[idx] = 1
-                    tensor.shapeSignature[idx] = -1
-
         if QPARAM_KEY in node.meta:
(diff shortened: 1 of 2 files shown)


jinevening:
> unexpected SymInt values can end up in tensor.shape, leading to complications due to mixing with Torch symbolic shapes

I agree.

In the example of `BasePowVisitor`, I think the visitor has to be updated to consider dynamic shapes.

PR #234, opened 2025-07-23
Pass shape signature to circle tensor creation interface

dayo09:
Let's enforce teh type contract by ensuring that only int values are stored in circle tensor.shape.

### DIFF
--- a/test/modules/op/depthwise_conv2d.py
+++ b/test/modules/op/depthwise_conv2d.py
@@ -32,6 +32,7 @@ class SimpleDepthwiseConv(torch.nn.Module):
         return (torch.randn(1, 8, 64, 64),)
 
 
+@tag.skip
 @tag.use_onert
 class SimpleDepthwiseConvDynamicShape(torch.nn.Module):
     def __init__(self):
--- a/tico/serialize/circle_graph.py
+++ b/tico/serialize/circle_graph.py
@@ -24,7 +24,7 @@ from torch._subclasses.fake_tensor import FakeTensor
 
 from tico.serialize.circle_mapping import (
     extract_circle_dtype,
-    extract_shape,
+    extract_circle_shape,
     str_to_circle_dtype,
     to_circle_dtype,
 )
@@ -151,15 +151,7 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         self.name_to_node[tensor.name] = node
         assert node.meta.get("val") is not None
         tensor.type = extract_circle_dtype(node)
-        tensor.shape = list(extract_shape(node))
-
-        # Handle dynamic shape
-        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
-            tensor.shapeSignature = tensor.shape.copy()
-            for idx, s in enumerate(tensor.shape):
-                if isinstance(s, torch.SymInt):
-                    tensor.shape[idx] = 1
-                    tensor.shapeSignature[idx] = -1
+        tensor.shape, tensor.shapeSignature = extract_circle_shape(node)
 
         if QPARAM_KEY in node.meta:
             tensor.quantization = to_circle_qparam(node.meta[QPARAM_KEY])
@@ -208,6 +200,7 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         self,
         prefix: str,
         shape: List[int],
+        shape_signature: Optional[List[int]],
         dtype: int,
         qparam: Optional[QuantParam] = None,
         source_node: Optional[torch.fx.Node] = None,
@@ -230,6 +223,8 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
             A name prefix used to generate a unique tensor name.
         shape : List[int]
             The shape of the tensor.
+        shape_signature : Optional[List[int]]
+            The shape signature of the tensor to express Dynamic Shape. Defaults to `None` for Static Shape.
         dtype : int
             The Circle-compatible dtype of the tensor. Use `to_circle_dtype()` to convert.
         qparam : Optional[QuantParam]
@@ -250,14 +245,8 @@ class CircleSubgraph(circle.SubGraph.SubGraphT):
         if source_node is not None:
             self.name_to_node[tensor.name] = source_node
         tensor.shape = shape
-
-        # Handle dynamic shape
-        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
-            tensor.shapeSignature = tensor.shape.copy()
-            for idx, s in enumerate(tensor.shape):
-                if isinstance(s, torch.SymInt):
-                    tensor.shape[idx] = 1
-                    tensor.shapeSignature[idx] = -1
+        if shape_signature:
+            tensor.shapeSignature = shape_signature
 
         if qparam is not None:
             tensor.quantization = to_circle_qparam(qparam)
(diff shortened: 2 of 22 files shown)


mhs4670go:
build failed.

tico/serialize/operators/op_copy.py
```diff
@@ -102,6 +110,12 @@ def define_node(
         # To connect 'dst' to Reshape node in the graph, 'dst' must be converted to Shape op.
         dst_tensor: circle.Tensor.TensorT = self.graph.get_tensor(dst)
         dst_shape: List[int] = dst_tensor.shape
+        dst_shape_signature: List[int] = dst_tensor.shapeSignature
+
+        if dst_shape_signature:
```
jinevening, line 115:
```suggestion
        if dst_shape_signature is not None:
```

tico/serialize/operators/op_copy.py
```diff
@@ -119,9 +134,16 @@ def define_node(
 
         src_tensor: circle.Tensor.TensorT = self.graph.get_tensor(src)
         src_shape: List[int] = src_tensor.shape
+        src_shape_signature: List[int] = src_tensor.shapeSignature
+
+        if src_shape_signature:
```
jinevening, line 139:
```suggestion
        if src_shape_signature is not None:
```

tico/serialize/operators/op_depthwise_conv2d.py
```diff
@@ -115,12 +115,17 @@ def define_node(self, node: torch.fx.Node) -> circle.Operator.OperatorT:
...
-        weight_shape = list(extract_shape(weight))  # 1HWO
+        input_shape, input_shape_signature = extract_circle_shape(input_)  # OHWI
+        output_shape, _ = extract_circle_shape(node)  # OHWI
+        weight_shape, _ = extract_circle_shape(weight)  # 1HWO
         assert len(input_shape) == 4, len(input_shape)
         assert len(output_shape) == 4, len(output_shape)
-        assert len(weight_shape) == 4
+        assert len(weight_shape) == 4, len(weight_shape)
+
+        if input_shape_signature:
```
jinevening, line 125:
```suggestion
        if input_shape_signature is not None:
```

tico/serialize/operators/op_max_pool2d_with_indices.py
```diff
@@ -88,7 +92,12 @@ def define_padding_node():
                 ],
                 dtype=torch.int32,
             )
-            input_shape = list(extract_shape(input))
+            input_shape, input_shape_signature = extract_circle_shape(input)
+
+            if input_shape_signature:
```
jinevening, line 97:
```suggestion
            if input_shape_signature is not None:
```

tico/serialize/operators/op_copy.py
```diff
@@ -52,7 +52,15 @@ class CopyVisitor(NodeVisitor):
...
-    def check_to_do_broadcast(self, dst: List[int], src: List[int]) -> bool:
+    def check_to_do_broadcast(
+        self,
+        dst: List[int],
+        dst_sig: Optional[List[int]],
+        src: List[int],
+        src_sig: Optional[List[int]],
+    ) -> bool:
+        assert dst_sig is None
+        assert src_sig is None
```
jinevening, line 63: Why shape signature are ignored?

jinevening (approved):
LGTM

tico/serialize/circle_graph.py
```diff
@@ -208,6 +200,7 @@ def add_tensor_from_scratch(
         self,
         prefix: str,
         shape: List[int],
+        shape_signature: Optional[List[int]],
         dtype: int,
```
mhs4670go, line 204:
```suggestion
        dtype: int,
        shape_signature: Optional[List[int]] = None,
```

Then, the change of `test/unit_test/serialize_test/test_circle_graph.py` wouldn't be necessary.

test/modules/op/depthwise_conv2d.py
```diff
@@ -32,6 +32,7 @@ def get_example_inputs(self):
         return (torch.randn(1, 8, 64, 64),)
 
 
+@tag.skip
```
mhs4670go, line 35: Why was this test marked as skipped?

tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -57,7 +57,12 @@ def has_padding(self, args: AvgPool2dArgs) -> bool:
             return True
 
     def has_same_padding(self, args: AvgPool2dArgs) -> bool:
-        input_shape = list(extract_shape(args.input))
+        input_shape, input_shape_signature = extract_circle_shape(args.input)
+
+        if input_shape_signature is not None:
+            # TODO: support dynamic shapes
+            raise NotImplementedError("Dynamic shape is not supported yet")
```
mhs4670go, line 64: Are these guards added because there's no test modules? Or, they really can't be run?

tico/serialize/circle_mapping.py
```diff
@@ -157,7 +175,8 @@ def check_if_i32_range(axis: Union[list, int]):
     return all(INT32_MIN <= val <= INT32_MAX for val in values)
 
 
-def circle_legalize_dtype_to(values, *, dtype: torch.dtype):
+# TODO: Revisit this dtype legalization function as it breaks SRP
```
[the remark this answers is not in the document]
dayo09: It does (1) cast dtype with range check (2) mapping to tensor
And it's used even only (1) is required.

tico/serialize/operators/op_avg_pool2d.py
dayo09: Or if it's the functionality of this, then I think it should be renamed.

tico/serialize/operators/op_any.py
```diff
@@ -99,17 +100,10 @@ def define_node(
...
-        dim_i32 = None
         if dim is None:
-            dims = tuple(i for i in range(0, len(input_shape)))
-            dim_i32 = tuple(
-                circle_legalize_dtype_to(dim, dtype=torch.int32) for dim in dims
-            )
-        if isinstance(dim, int):
-            dim_i32 = circle_legalize_dtype_to(dim, dtype=torch.int32)
-        if isinstance(dim, tuple):
-            dim_i32 = tuple(circle_legalize_dtype_to(d, dtype=torch.int32) for d in dim)
```
dayo09, line 111: @mhs4670go Look at this usage of circle_legalize_dtype_to. I guess there's some bad usage of this too.

tico/serialize/circle_graph.py
dayo09: Changing some unit tests wasn't big. 
I think it's better to manually passing None to shape_signature.

tico/serialize/operators/op_avg_pool2d.py
mhs4670go: Just for note, blocking dynamic dimensions only when the batch dimension is not dynamic could be a good idea as well, since in most cases it's the batch that's dynamic. However, blocking all dynamic dimensions entirely also makes sense, as it clearly defines the scope of coverage.

tico/serialize/circle_graph.py
```diff
@@ -250,14 +245,8 @@ def add_tensor_from_scratch(
...
         tensor.shape = shape
-
-        # Handle dynamic shape
-        if any(isinstance(s, torch.SymInt) for s in tensor.shape):
-            tensor.shapeSignature = tensor.shape.copy()
-            for idx, s in enumerate(tensor.shape):
-                if isinstance(s, torch.SymInt):
-                    tensor.shape[idx] = 1
-                    tensor.shapeSignature[idx] = -1
+        if shape_signature:
```
mhs4670go, line 248: Similar to other suggestions.
```suggestion
        if shape_signature is not None:
```

mhs4670go (approved):
LGTM

Issue #231, opened 2025-07-22
[CI] Introducing Ruff to Our Codebase


dayo09:
> [!TIP]
> Run `ruff check --fix` or `ruff check --unsafe-fixes` in your project to format quickly.

**Background**  
We currently use `pylint`, `mypy`, and `black` to manage linting, type checking, and formatting in our Python project. I would like to propose introducing [`ruff`](https://docs.astral.sh/ruff/) as an additional code quality tool. **The main motivation of this is because ruff makes auto fix (remove unused import/variable, etc).**

**Compatibility**  
- **Ruff and Black:** Ruff is designed to be compatible with Black, provided that the line-length settings are synchronized between the two tools. Ruff defers to Black for formatting, and any overlap can be resolved through configuration.
- **Ruff and Pylint:** Ruff and Pylint can run side-by-side. Ruff significantly overlaps with Pylint’s basic checks and is much faster, but the two tools can be complementary during staged adoption and in CI.
- **Ruff and Mypy:** Ruff focuses on style and linting, while Mypy remains the standard for type checking. They do not overlap in functionality and can be used together without issues.

**Migration Plan**
1. **Manual Run:**  
   - Run Ruff locally to apply and review all suggested code changes and auto-fixes.
   - Evaluate the changes, address any issues/conflicts, and ensure that existing Pylint, Black, and Mypy workflows are not adversely affected.
2. **Integrate into CI:**  
   - Once the codebase is compliant and stable with Ruff, add Ruff checks to the CI pipeline alongside the existing linters and type checkers.

Let me know if there are any objections or additional requirements before starting the process.

PR #239, opened 2025-07-23
Define __all__

dayo09:
Let's expose our APIs with __all__.
It has several advantages:
1. It clearly limits with symbols are publicly expoed.
2. When using `from module import *`, only the names listed in __all__ are imported.
3. It minimizes the impact of internal changes on external users.

---

From #231


seockho-kim (approved):
LGTM

mhs4670go (approved):
LGTM

PR #243, opened 2025-07-25
Support i8o16 relu in InsertQuantizeOpPass

jinevening:
This supports i8o16 relu in InsertQuantizeOpPass.


glistening (approved):
LGTM. By default, `8` seems to mean `u8`.

Issue #244, opened 2025-07-25
Circle Dynamic Shape Specification


dayo09:
Since there is no formal documentation for the circle dynamic shape specification, confusion and errors may occur during development and operation. Below is a summary of the spec, including potential error cases and important notes.

### Dynamic Shape in Torch Exported Program

- **torch tensor.shape:** `[s, 3, 3, 4]`  
  (where `s` represents a dynamic dimension. type: torch.SymInt)

### Dynamic Shape Definition in Circle

| Field                          | Example                               |
|--------------------------------|-------------------------------------|
| `circle::tensor.shape`          | `[1,3,3,4]`                      |
| `circle::tensor.shape_signature`| `[-1, 3, 3, 4]` (if dynamic shape)  |
| `circle::tensor.shape_signature`| `None` (if no dynamic shape)         |

- `tensor.shape_signature = None`  
  → **Valid (OK)**
- `tensor.shape_signature = []`  
  → **Invalid, causes errors in the ONE runtime!**

> **Note:** For static shapes, `shape_signature` must be set to `None`.  
> Using an empty list (`[]`) is not allowed because ONE runtime may mistakenly treat it as a shape, resulting in errors.

#### Related: ONE Runtime Behavior

- If the `shape_signature` field exists in the circle,  
  → The runtime **uses `shape_signature`**.
- If `shape_signature` does not exist,  
  → The runtime **falls back to using `shape`**.

### Caution When Generating Circle Files (e.g., using TICO)

- Assigning an empty list (`[]`) to `shape_signature` causes the field to be created in the circle file.
- This leads to malfunctions in the ONE runtime.
- Therefore, for static shapes, the `shape_signature` field **should not be created at all**.

#### Invalid Typing in Circle Schema

- The `circle_schema` python package we are using has an significant typing error.
  - To set `tensor.shapeSignature = None`, we should force mypy assert ignore manually. Related Issue: https://github.sec.samsung.net/AIP/python-tools/issues/887

PR #245, opened 2025-07-25
[serialize] Fix type annotation

dayo09:
Let's fix type annotation of Tensor.shapeSignature. NOTE: Tensor.shapeSignature has type of Optional[List[int]]. Flatbuffers fails in type annotation, generating it as List[int].

---

Related to https://github.com/Samsung/TICO/issues/244

> Never use [] for shape_signature. Always use None for static shapes.
> If you get typing errors from this, just add a # type: ignore comment to suppress mypy warnings.


jinevening (approved):
LGTM

Issue #233, opened 2025-07-23
Minimal Support for Dynamic Shapes with Padding Consideration


dayo09:
## Background
Currently, dynamic shapes are partially supported, but the handling of padding (e.g., in batch size or other dimensions) has not been properly considered. This limitation affects the accurate processing of operators where dynamic shapes overlap with padding dimensions.

## Problem Statement
There is a need for **minimal support** to handle cases where dynamic shapes exist alongside padding dimensions. Specifically:

- Even if a shape includes dynamic dimensions, when dynamic shapes appear in dimensions unrelated to padding shape calculations (such as batch size), 
- The other dimensions should still support dynamic shapes accordingly,
- Allowing for more flexible dynamic shape handling without fully restricting to fixed shapes due to padding.

## Objective
- Implement the minimal necessary support to handle dynamic shapes in this manner.
- Ensure that dimensions unrelated to padding computations correctly support dynamic shapes.
- Add comprehensive tests to verify the new support.

## Related Operators
- Average Pooling (avg pool)
- Convolution (conv)
- Depthwise Convolution (depthwise conv)
- Transpose Convolution (transpose conv)
- Instance Normalization (inst norm)
- Others as relevant

This issue aims to enhance the dynamic shape handling robustness across operators commonly used in neural network operations, particularly when padding dimensions complicate shape inference.

dayo09:
### TConv: Backend API differs (luci-interpreter VS onert)

- luci-interpreter expects TransposeConv to have 4 inputs: output_shape, filter, input, bias
- onert expects TransposeConv to have 3 inputs: output_shape, filter, input (no bias)
  - when bias is given, onert fails in `prepare`
  - when bias is not given (==proper input number), onert fails because it doesn't have TConv backend kernel.

<details>

#### luci-interpreter
```cpp
# compiler/luci-interpreter/src/kernels/TransposeConv.h

class TransposeConv : public KernelWithParams<TransposeConvParams>
{
public:
  TransposeConv(const Tensor *output_shape, const Tensor *filter, const Tensor *input,
                const Tensor *bias, Tensor *output, Tensor *scratch_tensor,
                const TransposeConvParams &params);

  ~TransposeConv();

  const Tensor *output_shape() const { return _inputs[0]; }
  const Tensor *filter() const { return _inputs[1]; }
  const Tensor *input() const { return _inputs[2]; }
  const Tensor *bias() const { return _inputs[3]; }
```

#### onert

```cpp
[...]

PR #247, opened 2025-07-25
Enable Dynamic Shapes with Padding Consideration

dayo09:
Let's utilize torch.SymInt for shape check.

Operation Coverage:
- Average Pooling (avg pool)
- Convolution (conv)
- Depthwise Convolution (depthwise conv)
- Transpose Convolution (transpose conv)
- Max Pool (max pool)
- Repeat

---

For #233

### DIFF
--- a/test/modules/op/avg_pool2d.py
+++ b/test/modules/op/avg_pool2d.py
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 import torch
+from torch.export import Dim
 
-from test.utils.tag import skip
+from test.utils.tag import skip, use_onert
 
 
 class SimpleAvgPool(torch.nn.Module):
@@ -28,6 +29,27 @@ class SimpleAvgPool(torch.nn.Module):
 
     def get_example_inputs(self):
         return (torch.randn(2, 4, 8, 16),)
+
+
+@use_onert
+class SimpleAvgPoolDynamicShape(torch.nn.Module):
+    def __init__(self):
+        super().__init__()
+        self.avgpool = torch.nn.AvgPool2d(kernel_size=3, stride=2)
+
+    def forward(self, tensor):
+        result = self.avgpool(tensor)
+        return result
+
+    def get_example_inputs(self):
+        return (torch.randn(2, 4, 8, 16),)
+
+    def get_dynamic_shapes(self):
+        batch = Dim("batch", min=1, max=128)
+        dynamic_shapes = {
+            "tensor": {0: batch},
+        }
+        return dynamic_shapes
 
 
 class AdaptiveAvgPool(torch.nn.Module):
--- a/test/modules/op/conv_transpose2d.py
+++ b/test/modules/op/conv_transpose2d.py
@@ -18,7 +18,7 @@ from torch.export import Dim
 from test.utils.tag import skip, use_onert
 
 
-class SimpleConvTranspose1(torch.nn.Module):
+class SimpleConvTranspose(torch.nn.Module):
     def __init__(self):
         super().__init__()
         self.tconv2d = torch.nn.ConvTranspose2d(16, 33, 3, stride=2)
--- a/test/modules/op/conv_transpose2d.py
+++ b/test/modules/op/conv_transpose2d.py
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 import torch
+from torch.export import Dim
+
+from test.utils.tag import skip, use_onert
 
 
-class SimpleConvTranspose(torch.nn.Module):
+class SimpleConvTranspose1(torch.nn.Module):
     def __init__(self):
         super().__init__()
         self.tconv2d = torch.nn.ConvTranspose2d(16, 33, 3, stride=2)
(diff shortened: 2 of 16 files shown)


tico/utils/serialize.py
```diff
@@ -47,4 +47,9 @@ def validate_tensor_shapes(
     Let's validate all tensors' shapes against their shape signatures.
     """
     for tensor in graph.tensors:
-        validate_circle_shape(tensor.shape, tensor.shapeSignature)
+        try:
```
dayo09, line 50: For better debug

tico/serialize/operators/op_avg_pool2d.py
```diff
@@ -142,11 +141,7 @@ def define_padding_node():
                 ],
                 dtype=torch.int32,
             )
-            input_shape, input_shape_signature = extract_circle_shape(input)
-
-            if input_shape_signature is not None:
-                raise RuntimeError("Dynamic shape is not supported yet.")
-
+            input_shape = extract_shape(input)
```
mhs4670go, line 144: Just curiosity, H or W can be a dynamic dimension, too?

test/modules/op/conv_transpose2d.py
```diff
@@ -24,7 +27,30 @@ def forward(self, input):
         return self.tconv2d(input)
 
     def get_example_inputs(self):
-        return (torch.randn(1, 16, 50, 100),)
+        return (torch.randn(4, 16, 50, 100),)
```
mhs4670go, line 30: This change is just for alignment with `SimpleConvTransposeDynamicShape`?

test/modules/op/depthwise_conv2d.py
```diff
@@ -55,6 +54,29 @@ def get_dynamic_shapes(self):
         return dynamic_shapes
 
 
+@tag.skip
```
mhs4670go, line 57: It would be good to have a reason for skip.

tico/serialize/operators/op_avg_pool2d.py
mhs4670go: Ah, I asked this because I wondered if arithmetic operation can be done with symbolic values like below. Most of cases will be a batch dimension though.

```python
padded_input_shape[1] += padding[0] * 2
```

dayo09: @mhs4670go Yes, it is possible. `torch.SymInt support all dimensions to be symbolic integer in this code.` -> Symbolic Integers are able to compute arithmetic operations with integers.

test/modules/op/conv_transpose2d.py
dayo09: Ah... it's a meaningless change. let me revert this.

test/modules/op/avg_pool2d.py
```diff
@@ -30,6 +31,27 @@ def get_example_inputs(self):
...
+        result = self.avgpool(tensor)
+        return result
+
+    def get_example_inputs(self):
+        return (torch.randn(2, 4, 8, 16),)
+
+    def get_dynamic_shapes(self):
+        batch = Dim("batch", min=1, max=128)
+        dynamic_shapes = {
+            "tensor": {0: batch},
```
Seunghui98, line 50: May I ask what `0` means?? 😃

test/modules/op/conv_transpose2d.py
```diff
@@ -27,6 +30,29 @@ def get_example_inputs(self):
         return (torch.randn(1, 16, 50, 100),)
 
 
+@skip(
+    reason="luci-interpreter does not support dynamic shape yet && onert does not support TransposeConv yet"
```
Seunghui98, line 34:
```suggestion
    reason="luci-interpreter does not support dynamic shape yet && onert does not support ConvTranspose yet"
```
(optional) maybe..?

test/modules/op/repeat.py
```diff
@@ -30,6 +31,27 @@ def get_example_inputs(self):
...
+        return z
+
+    def get_example_inputs(self):
+        torch.manual_seed(1)
+        return (torch.randn(1, 4),)
+
+    def get_dynamic_shapes(self):
+        dim = Dim("dim", min=1, max=128)
+        dynamic_shapes = {
+            "x": {1: dim},
```
Seunghui98, line 50: it's just out of curiousity.
Why do you set key to `1`..?

tico/serialize/circle_mapping.py
```diff
@@ -132,18 +132,36 @@ def extract_circle_shape(node: torch.fx.Node) -> Tuple[List[int], Optional[List[
...
+def to_circle_shape(
+    torch_shape: Union[
+        torch.Size, Sequence[int | torch.SymInt]
+    ],  # Sequence[int | torch.SymInt] is added for type covariance
+) -> Tuple[List[int], Optional[List[int]]]:
 
-    if any(isinstance(s, torch.SymInt) for s in shape):
-        shape_signature = shape.copy()
-        for idx, s in enumerate(shape):
+    if any(isinstance(s, torch.SymInt) for s in torch_shape):
```
llFreetimell, line 141: Since I am not good at torch... if `torch_shape is typeof(torch.Size)`, what happens here?

test/modules/op/repeat.py
llFreetimell: Because `dim0` is 128, which is not dynamic dimension..?

test/modules/op/conv_transpose2d.py
dayo09: 'TransposeConv' is a Circle Operator name :-D

tico/serialize/circle_mapping.py
dayo09: I don't fully understand the code but I think that torch.Size is a tuple of torch-defined integer type, including `int` and `torch.SymInt`.

-----

**DETAILS**

torch.Size is defined as a tuple of `_int`(pytorch private-defined integer type)

```py
# Defined in torch/csrc/Size.cpp
class Size(Tuple[_int, ...]):
    # TODO: __reduce__
...
```

See that in has this definition. I guessed that it includes torch.SymInt.
```py
class int:
    @overload
    def __new__(cls, x: ConvertibleToInt = ..., /) -> Self: ...
    @overload
    def __new__(cls, x: str | bytes | bytearray, /, base: SupportsIndex) -> Self: ...
    def as_integer_ratio(self) -> tuple[int, Literal[1]]: ...
    @property
    def real(self) -> int: ...
    @property
    def imag(self) -> Literal[0]: ...
    @property
    def numerator(self) -> int: ...
    @property
    def denominator(self) -> Literal[1]: ...
    def conjugate(self) -> int: ...
    def bit_length(self) -> int: ...
    if sys.version_info >= (3, 10):
```

Seunghui98 (approved):
LGTM!!

Issue #207, opened 2025-07-10
Auto example input generation


glistening:
It would be nice if we don't need to write code for example inputs.

For example, we need something like:

https://github.com/Samsung/TICO/blob/a3ed23ec86bd2024bee9a4a095882e05d4f1f523/test/modules/model/LlamaWithKVCache/model.py#L36-L70

To correctly working example input, we need to understand much about the target model.

I tried to capture the input from user-level inputs.

I succeed to write a working version, tested with `Maykeye/TinyLLama-v0`.

It confirmed it works for transformers  version below:

- 4.49.0 :x: DynamicCache is not pytree-flattenable
- 4.51.3 :o:
- 4.52.4 :o:

A good news is that PyTorch supports transformers DynamicCahce as pytree-flatenable since 4.50.0. (though it seems some bug on MacOS, all versions after 4.50.1 would work.

I hope and guess it will work for other models by modifying model name and user inputs.

glistening:
@mhs4670go Wow, I didn't know the issue. Let me close this issue, and move my comment there.

glistening:
I added a context manager for recording forward's inputs.

Just run generate() with recording enabled:

```py
with torch.no_grad(), RecordingInput(model, input_to_remove=input_to_remove) as rec:
    outputs = model.generate(
        **inputs,
        max_new_tokens=32,
        do_sample=False,
        pad_token_id=tokenizer.eos_token_id,
    )
    captured_input = rec.captured_input
```

- No change in model.generate(...)
- Just run generate(...) in `RecordingInput` context

It will capture model.forward(...)'s inputs to captured_input.

Then, just pass this captured input as input for `tico.convert`.


```py
circle_model = tico.convert(model, captured_input)
```

glistening:
@dayo09 I am using torch 2.7.1 and I have no problem. Could you please explain in detail?

PR #249, opened 2025-07-28
[utils] Add forward's input recorder

glistening:
It adds forward's input recorder.

Related: #207 
Draft: #217

### DIFF
--- a/tico/utils/record_input.py
+++ b/tico/utils/record_input.py
@@ -1,0 +1,89 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+
+import inspect
+from contextlib import contextmanager
+from typing import Callable, List, Optional
+
+import torch.nn as nn
+
+
+class RecordingInput:
+    r"""Context-manager that records the input values of model::forward()
+
+    Recording input is useful for preparing example input for torch.export
+
+    Args:
+        condition: lambda to provide the condition whether to record or not
+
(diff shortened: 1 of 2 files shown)


tico/utils/record_input.py
```diff
@@ -0,0 +1,89 @@
...
+    ):
+        self.module = module
+        self.forward_org = module.forward
+        self.condition = condition
+        self.input_to_remove = input_to_remove
+        sig = inspect.signature(self.forward_org)
+        self.args_names = [
+            name for name in sig.parameters.keys() if name not in ("self", "kwargs")
+        ]
+        self.captured_input = ()
```
dayo09, line 65:
```suggestion
        self.captured_input = None
```

How about initializing this with None? 
Because, in some rare case, captured input could be void...

tico/utils/record_input.py
```diff
@@ -0,0 +1,89 @@
...
+
+    def __enter__(self):
+        def capture_and_forward(*args, **kwargs):
+            args_dict = dict(zip(self.args_names, args))
+            args_dict.update(kwargs)
+
+            def populate_args(args_dict, filter):
```
dayo09, line 72:
```suggestion
            def populate_args(args_dict, input_to_remove):
```
`filter` sounds like a function here, but it's a list of argument name.

glistening: I've never thought filter is related to a function. It is just something to filter. However, if you feels like that, I will change.

tico/utils/record_input.py
```diff
@@ -0,0 +1,89 @@
...
+        sig = inspect.signature(self.forward_org)
+        self.args_names = [
+            name for name in sig.parameters.keys() if name not in ("self", "kwargs")
+        ]
+        self.captured_input = ()
+
+    def __enter__(self):
+        def capture_and_forward(*args, **kwargs):
+            args_dict = dict(zip(self.args_names, args))
+            args_dict.update(kwargs)
```
dayo09, line 70: FYI, this works!
```
args_dict = dict(sig.bind(*args, **kwargs).arguments)
```

glistening: Hmm, I am not sure it is perfectly equivalent to my original code. I have to refer to `sig.bind` and `arguments` and so on. Even it is equivalent, I found no benefit of using this API. Is there any advantage to replace 2 lines of straightforward code to your suggestion?

tico/utils/record_input.py
dayo09: @glistening Could you check this..?

dayo09 (approved):
LGTM Thanks!

PR #252, opened 2025-07-30
[utils] Use signature param.kind to filter out var_positional

glistening:
It uses signature's param.kind to filter out **kwargs since VAR_KEYWORD argument's name may be anything else like the below.

```py
def forward (...
  **flash_attn_kwargs: Unpack[FlashAttentionKwargs],
)
```


seockho-kim (dismissed):
LGTM

tico/utils/record_input.py
```diff
@@ -61,8 +61,14 @@ def __init__(
...
         self.args_names = [
             name
-            for name in self.sig.parameters.keys()
-            if name not in ("self", "kwargs")
+            for name, param in self.sig.parameters.items()
+            if param.kind
+            in (
+                inspect.Parameter.POSITIONAL_ONLY,
+                inspect.Parameter.POSITIONAL_OR_KEYWORD,
+                inspect.Parameter.KEYWORD_ONLY,
```
mhs4670go, line 69: Maybe except KEYWORD_ONLY?

glistening:
@mhs4670go Thank you for review.

My code gets complicated since `get_example_inputs()` supports **tuple (args) only**.
Main job of `RecordingInput` is reconstructing only tuple from args and kwargs.

Is it okay not to support `get_example_inputs()`?

It means that we can keep both `args` and `kwargs` as you suggested.
Then, it may be simplified. I will check your code.

PR #255, opened 2025-07-30
[test] Revise get_example_inputs

dayo09:
Let's revise get_example_inputs to return tuple of args, kwargs.


mhs4670go (approved):
LGTM

PR #256, opened 2025-07-31
[test] Run more calib for GPTQ

mhs4670go:
This commit runs more calib for GPTQ.

Before this PR, sometimes the test fails because the number of calibration is a bit small.


dayo09 (approved):
LGTM

PR #258, opened 2025-07-31
[test] Revise get_calibration_data

dayo09:
Let them return generator to save from further possible memory problem.


mhs4670go (approved):
LGTM

Issue #248, opened 2025-07-28
Lesson Learned: Torch Export with Dynamic Shape


dayo09:
## What?

Let's share some tips or experiences related to `torch.export.export` with `dynamic_shapes`.

dayo09:
> ### Constraints violated error
> 
> ```py
> Constraints violated (batch)!
> Not all values of batch = L['x'].size()[0] in the specified range batch <= 128 are valid because batch was inferred to be a constant (3).
> ```
> 
> While adding some dynamic shape unit test cases, I confront this error. 
> Sometimes, it can be resolved by putting non-1 value to example input dimension.
> If above trick doesn't work, it's possibily because the dimension is fixed while tracing - which means, that cannot be `dynamic` shape.
> 
> 
> #### Maybe..?
> 
> They hints me to remove direct referencing to shape dimension.
> ```py
> def forward(self, x):
>     bsz = x.shape[0]  #<---Nope!
>     y = x.view(bsz, -1)
>     ...
> 
> def forward(self, x):
>     y = x.view(-1, x.shape[-1]) #<---- Fine :-D
>     ...
> 
> ```
> 
> I didn't find an exact match to above case, so it's not tested. I just note it down to keep in memory. 

 _Originally posted by @dayo09 in [#238](https://github.com/Samsung/TICO/issues/238#issuecomment-3126555029)_

mhs4670go:
I think adding a full export script would be better. It's hard to know where the error happened.

dayo09:
@mhs4670go Okay! But, actually, it happens a lot...

jinevening:
Where does this `forward` come from?

```
def forward(self, x):
    bsz = x.shape[0]  #<---Nope!
    y = x.view(bsz, -1)
    ...
```

It's not in the example codes.

dayo09:
@jinevening 

> It's not in the example codes.

Yes, I just noted down the code fragment to refer later.. 😅

PR #263, opened 2025-08-04
DRAFT: Parse args properly

dayo09:


test/modules/op/avg_pool2d.py
```diff
@@ -34,22 +34,28 @@ def get_example_inputs(self):
 
 
 @use_onert
+@skip_if(
+    Version(torch.__version__) >= Version("2.9.0.dev"),
+    reason="From torch 2.9.0.dev, dynamic shape requires more torch ir conversions related to symbolic runtime assertions",
```
dayo09, line 39: For details, see https://github.com/Samsung/TICO/issues/248#issuecomment-3149772872

tico/utils/signature.py
```diff
@@ -0,0 +1,227 @@
...
+            result[k] = v
+
+    # 3. Convert to tensors
+    for k, v in result.items():
+        result[k] = v if isinstance(v, torch.Tensor) else torch.tensor(v)
+
+    return result
+
+
+def flatten_and_convert_args(args: Sequence) -> tuple:
```
dayo09, line 60: This flattening is required because `circle model` doesn't take tuples or list as argument. They must be flattened.

test/utils/infer.py
```diff
@@ -42,7 +42,9 @@ def infer_with_circle_interpreter(
         The output produced by the 'circle-interpreter'
     """
     circle_model = tico.utils.model.CircleModel.load(circle_path)
-    circle_result = circle_model(*forward_args, **forward_kwargs)
+    ispec = ModelInputSpec.load(circle_path)
+    inputs = ispec.bind(forward_args, forward_kwargs, check=True)
+    circle_result = circle_model(*inputs)
```
mhs4670go, line 47: Introducing feature looks very good. But I guess it would be more helpful if input bindings are automatically done inside the CircleModel like below. 

```python
circle_model = tico.utils.model.CircleModel.load(circle_path)
circle_result = circle_model(*forward_args, **forward_kwargs) # input bindings are done inside the `__call__` method.
```

Anyway, my suggestion doesn't have to be applied in this PR. Let's consider this later.

test/modules/op/avg_pool2d.py
dayo09: @mhs4670go I think there's both up and down sides. `Version(torch.__version__) >= Version("2.9.0.dev"),` let us know that `_assert_scalar` is produced from torch 2.9.0.dev, so we can target the operation's support with torch upgrade.

mhs4670go: Okay. Let's adapt as we go. Then, `export_produces_assert_scalar` is just implemented as a fallback?

dayo09: I forgot to remove it 😓 I updated the PR!

mhs4670go (approved):
LGTM

test/modules/model/LlamaAttentionWithKVCache/model.py
```diff
@@ -57,5 +57,4 @@ def get_example_inputs(self):
             hidden_states,
             position_embeddings,
             attention_mask,
-            past_key_values,
-        ), {}
+        ), {"past_key_values": past_key_values}
```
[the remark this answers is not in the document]
jinevening: Ah, this is `get_example_inputs`.. Sorry for confusion.

tico/utils/signature.py
```diff
@@ -0,0 +1,241 @@
...
+                else:
+                    # base type (scalar etc.) directly stored
+                    store[name] = value
+
+                return store
+
+            unpack_recursive(k, v, result)
+        elif is_dynamic_cache_instance(v):
+            # 2. handle DynamicCache
+            for idx, cache_val in enumerate(v.key_cache):
```
seockho-kim, line 59: it's minor, how about `val` instead of `cache_val`?

seockho-kim (approved):
LGTM, with minor comment.

tico/utils/signature.py
dayo09: 'v' was used above so I added it to distinguish them.. :-D

tico/utils/signature.py
glistening: Could you please show an example of flattening? Your comment is saying circle modle doesn't take tuples or list as argument. But the result of flatten is tuple. I don't understand. Maybe...

args = (Tensor1, Tensor2, scalar); is okay.
args = (Tensor1, tuple(Tensor21, Tensor22), scalar); is not okay.
So you want to make (Tensor1, newTensor2, scalar) by flattening. Is it right?

I look up the flattening example in test_signature.py. But I can't find.

tico/utils/dtype.py
```diff
@@ -15,6 +17,29 @@
     np.dtype("bool"): torch.bool,
 }
 
+CIRCLE_TO_TORCH_DTYPE_DICT = {
+    circle.TensorType.TensorType.FLOAT32: torch.float32,
+    circle.TensorType.TensorType.FLOAT32: torch.float,
```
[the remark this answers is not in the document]
dayo09: Thanks for the mention, I flipped the torch->circle dmap so it was mistaken..

tico/utils/signature.py
dayo09: @glistening I added the test test_bind_tuple, PTAL!

seockho-kim (dismissed):

mhs4670go (dismissed):
LGTM

glistening (approved):
LGTM

mhs4670go (approved):
LGTM

PR #271, opened 2025-08-08
[ptq] Introduce Observer class

mhs4670go:
This commit introduces Observer class. It needs for calibration and computing qparms.


dayo09 (approved):
LGTM

tico/experimental/quantization/ptq/utils/reduce_utils.py
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+
+def reduce_except(x: torch.Tensor, dim_to_keep: int):
```
jinevening, line 18:
```suggestion
def reduce_minmax_except(x: torch.Tensor, dim_to_keep: int):
```

tico/experimental/quantization/ptq/observers/base.py
```diff
@@ -0,0 +1,160 @@
...
+        self.qscheme = qscheme
+        self.channel_axis = channel_axis if qscheme.is_per_channel() else None
+        self.enabled = True
+        self.reset()
+
+    def reset(self) -> None:
+        self.min_val: torch.Tensor = torch.tensor(math.inf)
+        self.max_val: torch.Tensor = torch.tensor(-math.inf)
+
+    def load_qparams(self, scale: torch.Tensor, zp: torch.Tensor, *, lock: bool = True):
```
seockho-kim, line 49: how about `set_qparams` or `inject_qparams` instead of `load_qparams`?

tico/experimental/quantization/ptq/observers/base.py
```diff
@@ -0,0 +1,160 @@
...
+
+from tico.experimental.quantization.ptq.dtypes import DType, UINT8
+from tico.experimental.quantization.ptq.qscheme import QScheme
+from tico.experimental.quantization.ptq.utils import reduce_except
+
+
+class ObserverBase:
+    """
+    Tracks activation/weight ranges and converts them into (scale, zero-point).
+    """
```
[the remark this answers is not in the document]
mhs4670go: Other observers have been implemented in the draft. Anyway, I also think it's a good idea to make this class a real base class.

tico/experimental/quantization/ptq/observers/base.py
mhs4670go: Thanks for the suggestion. But, here's what I thought.

- `set_*` reads like an internal setter for implementation details.
- `inject_*` feels like a testing/force-override hook.

I can also say `load_*` mirrors common PyTorch "load_*" semantics. (e.g. load_state_dict(), load()): the values come from an external source rather than being computed here.

tico/experimental/quantization/ptq/utils/reduce_utils.py
mhs4670go: Nice catch. I'll rename it to `channelwise_minmax` for more readability.

tico/experimental/quantization/ptq/observers/base.py
```diff
@@ -0,0 +1,165 @@
...
+from abc import ABC, abstractmethod
+from typing import Optional
+
+import torch
+
+from tico.experimental.quantization.ptq.dtypes import DType, UINT8
+from tico.experimental.quantization.ptq.qscheme import QScheme
+
+
+class ObserverBase(ABC):
```
jinevening, line 25: I think algorithm-specific implementations are still included in this class. For example,

channel_axis,
min_val, max_val
compute_qparams,
fake_quant

The above fields and methods are for min/max-based affine quantization. Could you pull them down to MinMaxObserver? ObserverBase would be like

ObserverBase
-> `__init__()`
-> `reset` (abstract method -> Or we can make this as a final method and define another abstract method for algorithm-specific implementation. (see strategy pattern https://refactoring.guru/design-patterns/strategy) )
-> `load_qparams`
-> `has_qparams`
-> `collect` (abstrat method)
-> `compute_qparams` (abstrat method)
-> `fake_quant` (abstrat method)
-> `__repr__` (abstrat method)

mhs4670go: @jinevening Well, I think most practical observers like MinMax, Percentile, KL/Histogram, etc reduce to producing [min_val, max_val] and compute quantization parameter accordingly. Other deviations can be overridden. 

Are you considering any specific observer that can't match with this base class?

mhs4670go: Got it. This is what I'm gonna do.

1. Make `ObserverBase` slim.
2. Introduce `AffineObserverBase` that inherits `ObserverBase` for current observers in the draft having common logic. Current observer now inherits this class.
3. Introduce `MXObserver` that doesn't track min/max or produce affine qparams as an example of a non-affine observer.

test/quantization/ptq/utils/test_reduce_utils.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+from tico.experimental.quantization.ptq.utils.reduce_utils import channelwise_minmax
+
+
+class TestReduceExcept(unittest.TestCase):
```
jinevening, line 21: Nit

```suggestion
class TestChannelWiseMinMax(unittest.TestCase):
```

mhs4670go: I'll apply this in next PR.

seockho-kim (approved):
LGTM

tico/experimental/quantization/ptq/observers/base.py
```diff
@@ -0,0 +1,87 @@
...
+        raise NotImplementedError
+
+    @abstractmethod
+    def collect(self, x: torch.Tensor) -> None:
+        """
+        Update running statistics with a new batch of data.
+
+        Must be implemented by subclasses (e.g., MinMax, EMA, Histogram, MSE).
+        """
+        raise NotImplementedError
```
dayo09, line 63: When enabled == False, `collect` supposed to be void function in this design, right? How about putting that implementation here in base?

dayo09: It's optional. AFAIK, Abstract class can still have some implemented functions. 
```py
def collect(...
  if not enabled:
    return
  update_statistics()

@abstractmethod
def update_statistics(...


```

Like above.

This is optional, though

seockho-kim (approved):

PR #278, opened 2025-08-13
[ptq] Add more observers

mhs4670go:
This commit adds more observers that are commonly used.

- Percentile
- EMA
- Identity

### DIFF
--- a/test/quantization/ptq/observers/test_ema.py
+++ b/test/quantization/ptq/observers/test_ema.py
@@ -1,0 +1,86 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import unittest
+
+import torch
+from tico.experimental.quantization.ptq.dtypes import DType
+from tico.experimental.quantization.ptq.observers.ema import EMAObserver
+from tico.experimental.quantization.ptq.qscheme import QScheme
+
+
+class TestEMAObserver(unittest.TestCase):
+    def test_ema_updates(self):
+        torch.manual_seed(0)
+        obs = EMAObserver(
+            name="dummy",
+            dtype=DType.uint(8),
+            qscheme=QScheme.PER_TENSOR_ASYMM,
+            momentum=0.5,
+        )
+
+        # First batch initialises stats hard
+        obs.collect(torch.tensor([-1.0, 2.0, 3.0]))
+        self.assertAlmostEqual(obs.min_val.item(), -1.0)
+        self.assertAlmostEqual(obs.max_val.item(), 3.0)
+
+        # Second batch updates via EMA
+        obs.collect(torch.tensor([-9.0, 9.0]))
+        # new_min = 0.5 * (-1) + 0.5 * (-9) = -5
+        # new_max = 0.5 * 3   + 0.5 * 9     =  6
+        self.assertAlmostEqual(obs.min_val.item(), -5.0)
+        self.assertAlmostEqual(obs.max_val.item(), 6.0)
+
+    def test_per_channel_ema(self):
+        x1 = torch.tensor([[1.0, -2.0], [3.0, -4.0]])  # shape (C=2, N=2)
+        x2 = torch.tensor([[10.0, -10.0], [5.0, -5.0]])
+        obs = EMAObserver(
+            name="dummy",
+            dtype=DType.int(4),
+            qscheme=QScheme.PER_CHANNEL_ASYMM,
+            channel_axis=0,
+            momentum=0.5,
+        )
+
+        obs.collect(x1)
+        self.assertTrue(
+            torch.equal(obs.min_val, torch.tensor([-2.0, -4.0])),
+            f"obs.min_val: {obs.min_val}",
+        )
+        self.assertTrue(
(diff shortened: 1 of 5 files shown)


tico/experimental/quantization/ptq/observers/percentile.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def __init__(self, *, percentile: float = 99.9, **kwargs):
+        super().__init__(**kwargs)
+        assert (
+            0 <= percentile <= 100
+        ), f"percentile must be between 0 and 100, got {percentile}"
+        self.percentile = percentile
+
+    @torch.no_grad()
+    def _update_stats(self, x):
```
jinevening, line 34: This behavior is different from circle-quantizer's percentile implementation.

Do you have any reference on this implementation?

dayo09 (dismissed):
LGTM

seockho-kim (dismissed):
LGTM

tico/experimental/quantization/ptq/observers/percentile.py
jinevening: Your observer finds percentile whenever each datapoint produces activation distribution.

Here is how record-minmax works.
```
For data in dataset:
  X = layer(data)
  min_array.append(min(X))
  max_array.append(max(X))
min = percentile(min_array, low)
max = percentile(max_array, high)
compute_qparam(min, max)
```

And, I don't understand why percentile_low is divided by 200.

```
        q_low = (100.0 - self.percentile) / 200.0  # two-sided
```

jinevening: If you intend to sync implementation with circle-quantizer, you'd be better update the algorithm and add reference (https://github.com/Samsung/ONE/blob/release/1.30.0/compiler/record-minmax/src/MinMaxComputer.cpp).

mhs4670go: Current codes computes two-sided percentiles on the samples and clips to that central coverage. That is why it is divided by 200. 

If we want to keep _p_%, the total we discard is (100 - p)%. Splitting that equally between the two tails (lower and upper) gives (100 - p)/2 % per tail. To convert a percentage to a probability in [0, 1], divide by 100. 

So, q_low = ((100-p)/2)/100 = (100-p)/200.

And seems to me that circle-quantizer's approach builds arrays of batch min/max and then takes their percentiles, which is not the same as the data's percnetiles. It's the distribution of extremes. 

FYI, I've found some references from other frameworks.

- [Quark](https://github.com/amd/Quark/blob/1241e270912113218b85189debbbf096db03f394/quark/torch/quantization/observer/observer.py#L1113)
- [ONNXRuntime](https://github.com/microsoft/onnxruntime/blob/main/onnxruntime/python/tools/quantization/calibrate.py#L940)
- [neural-compressor](https://github.com/intel/neural-compressor/blob/1ab2011c2f8149b63ce270f34bd3983c81161958/neural_compressor/adaptor/ox_utils/calibrator.py#L113)

Those implementations use the original, data-based percentile clipping (via histogram/CDF), not min/max-array percentile variant.

mhs4670go: Hmmm.. IIUC, I don't think this PR's impl is different from histogram one that much. 

I've visualized the difference like below. Please correct me if there's something wrong.

That being said, as you said, the circle quantizer has been used with lots of models. So, I'll create dedicated issue for this percentile implementation and check the effectiveness of each method.

Let's just merge `Identity` and `EMA` in this PR.

```python
import numpy as np
import matplotlib.pyplot as plt
import pandas as pd

# ---------- Config ----------
seed = 7
p = 99.9     # central coverage percentage to KEEP (middle p%)
m = 128      # batch size
B = 1500     # number of batches (total samples = m*B)

# Heavy-tailed mixture to make outliers visible
main_mu, main_sigma = 0.0, 1.0
tail_mu, tail_sigma = 0.0, 8.0
tail_prob = 0.015

# ---------- Generate data ----------
rng = np.random.default_rng(seed)
N = m * B
X = rng.normal(main_mu, main_sigma, N)
mask = rng.random(N) < tail_prob
X[mask] = rng.normal(tail_mu, tail_sigma, mask.sum())
X_batches = X.reshape(B, m)

# ---------- Targets (central coverage) ----------
q_low = (1.0 - p / 100.0) / 2.0
q_high = 1.0 - q_low

# A) Data-based percentiles (all samples)
lo_A, hi_A = np.quantile(X, [q_low, q_high])

# B) Min/Max arrays
mins = X_batches.min(axis=1)
maxs = X_batches.max(axis=1)

# B) circle-quantizer's
lo_B = np.quantile(mins, q_low)
hi_B = np.quantile(maxs, q_high)

# ---------- Summary table ----------
def width(lo, hi): return float(hi - lo)
summary = pd.DataFrame.from_records([
    {"Method": "A Data-based",            "Lower": lo_A,  "Upper": hi_A,  "Width": width(lo_A, hi_A)},
    {"Method": "B Min/Max batch",         "Lower": lo_B, "Upper": hi_B, "Width": width(lo_B, hi_B)},
])
base_w = summary.loc[0, "Width"]
summary["Width / A"] = summary["Width"] / base_w
print(summary.round(6))

# ---------- Plot ----------
plt.figure()
plt.hist(X, bins=200)
plt.axvline(lo_A, linestyle="--", label="A lower")
plt.axvline(hi_A, linestyle="--", label="A upper")
plt.axvline(lo_B, linestyle="-.", label="B lower")
plt.axvline(hi_B, linestyle="-.", label="B upper")
plt.title(f"Central coverage = {p}%  |  batch size m={m}, batches B={B}")
plt.xlabel("value"); plt.ylabel("count"); plt.legend(); plt.show()
```

jinevening: Your impl applies percentile first (clipping in local data), and then get minmax.

Histogram approach gathers distribution of all data points as a histogram, and then gets percentile (clipping in global data).

In normal dist, your impl will likely discard more data points in tails.

Anyway, I think it's good to postpone implementation of percentile observer.

jinevening (approved):
LGTM

seockho-kim (approved):

dayo09 (approved):
LGTM

PR #280, opened 2025-08-13
[passes] Add missing assert keyword

mhs4670go:
This commit adds missing assert keyword.


dayo09 (approved):
LGTM

PR #283, opened 2025-08-14
[utils] Use a generator

mhs4670go:
This commit uses a generator instead of list comprehension.

This resolves pylint error.


jinevening (approved):
LGTM

PR #284, opened 2025-08-18
[ptq] Introduce wrapper base class and qconfig class

mhs4670go:
This commit introduces a wrapper base class and related config class.

You can see the actual use case for these classes in the draft codes.


dayo09 (dismissed):
LGTM with minor comments

tico/experimental/quantization/ptq/quant_config.py
```diff
@@ -0,0 +1,111 @@
...
+        Two-level mapping of *scopes* → *observer-kwargs*.
+
+        • **Scope** can be either
+            - the *attribute name* of a child wrapper
+              (e.g. ``"gate_proj"`` or ``"up_proj"``), or
+            - an *observer logical name* inside *this* wrapper
+              (e.g. ``"mul"``, ``"act_in"``).
+
+        • **Observer-kwargs** is forwarded verbatim to the observer constructor
+          (``dtype``, ``qscheme``, ``channel_axis``, ``observer``, …).
```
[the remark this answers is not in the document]
mhs4670go: Let's apply this change at once in another PR.

tico/experimental/quantization/ptq/wrappers/quant_module_base.py
```diff
@@ -0,0 +1,153 @@
...
+        name: str,
+        **default_kwargs,
+    ) -> ObserverBase:
+        """
+        Instantiate an observer named *name*.
+
+        Precedence (3-tier) for keys:
+           • observer:  user > wrapper-default > QuantConfig.default_observer
+           • dtype:     user > wrapper-default > QuantConfig.default_dtype
+           • qscheme  : user  > wrapper-default > QuantConfig.default_qscheme
```
miusic, line 109:
```suggestion
           • qscheme  : user > wrapper-default > QuantConfig.default_qscheme
```
(nit)

test/quantization/ptq/wrappers/test_quant_module_base.py
```diff
@@ -0,0 +1,209 @@
...
+        assert isinstance(self.qm.obs, AffineObserverBase)
+        lo = self.qm.obs.min_val.item()
+        hi = self.qm.obs.max_val.item()
+        self.assertLess(lo, hi)  # stats updated
+
+        # QUANT pass – output must differ from FP32
+        self.qm.freeze_qparams()
+        q_out = self.qm(self.x)
+        fp_out = self.x * 2.0
+        self.assertFalse(torch.allclose(q_out, fp_out))
```
miusic, line 95:
```suggestion
        self.assertFalse(torch.equal(q_out, fp_out))
```
(optional)
How about to use `torch.equal` here?
Is there no possibility for the case when the results are in the tolerance range, since quantization is done **well**?

tico/experimental/quantization/ptq/quant_config.py
```diff
@@ -0,0 +1,111 @@
...
+
+    cfg = QuantConfig(
+        default_dtype   = DType.uint(8),
+        default_qscheme  = QScheme.PER_TENSOR_SYMM,        # <- global scheme
+        default_observer = PercentileObserver,             # <- global algorithm
+        overrides={
+            # local override: input observer now MinMax & 4-bit, per-channel affine
+            "act_in": {"observer": MinMaxObserver,
+                       "dtype":    DType.uint(4),
+                       "qscheme":  QScheme.PER_CHANNEL_ASYMM},
```
miusic, line 67:
```suggestion
            # local override: input observer now MinMax & 4-bit, per-channel affine <-- question for this 1
            "act_in": {"observer": MinMaxObserver,
                       "dtype":    DType.uint(4),
                       "qscheme":  QScheme.PER_CHANNEL_ASYMM}, <-- question for this 2
```
(Just out of curiosity)
Does the `affine` in the comment mean `ASYMM` in the `QScheme.PER_CHANNEL_ASYMM`..? 🤔

test/quantization/ptq/wrappers/test_quant_module_base.py
mhs4670go: > How about to use torch.equal here?

`torch.equal` always results in `False` on float values. So, `allclose` is appropriate in this case.

> Is there no possibility for the case when the results are in the tolerance range, since quantization is done well?

I think it won't be. Let's fix it when it is the case.

tico/experimental/quantization/ptq/wrappers/quant_module_base.py
```diff
@@ -0,0 +1,153 @@
...
+        name: str,
+        **default_kwargs,
+    ) -> ObserverBase:
+        """
+        Instantiate an observer named *name*.
+
+        Precedence (3-tier) for keys:
+           • observer:  user > wrapper-default > QuantConfig.default_observer
+           • dtype:     user > wrapper-default > QuantConfig.default_dtype
+           • qscheme:   user  > wrapper-default > QuantConfig.default_qscheme
```
miusic, line 109:
```suggestion
           • qscheme:   user > wrapper-default > QuantConfig.default_qscheme
```
Just FYI, there is a blank still remaining
I'm not requesting you to change it, but just in case you forgot it... 😓

miusic (approved):
LGTM :)

PR #287, opened 2025-08-19
[ptq] Introduce a Linear wrapper class

mhs4670go:
This commit introduces a Linear wrapper and related registry mechanism.


tico/experimental/quantization/ptq/wrappers/registry.py
```diff
@@ -0,0 +1,112 @@
...
+      (e.g. "ptq.wrappers.linear_quant").  Importing the module runs all
+      its `@register(nn.Layer)` decorators, populating `_WRAPPERS`.
+    * After the first call the function becomes a cheap constant-time no-op.
+    """
+    global _IMPORT_ONCE
+    if _IMPORT_ONCE:
+        return
+    for mod in _CORE_MODULES:
+        __import__(mod)  # triggers decorators
+    _IMPORT_ONCE = True
```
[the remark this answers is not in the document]
mhs4670go: Thanks for the letting me know. Actually, `_lazy_init()` is practically thread-safe because `__import__` are serialized by Python's import lock, and `sys.modules` ensures each module is initialized once.

Maybe worst case, multiple threads may run the import loop redundantly as you said, but registration is idempotent (same keys overwritten), so it's harmless. Let's fix it if it turns out it really matters.

test/quantization/ptq/wrappers/nn/test_quant_linear.py
```diff
@@ -0,0 +1,84 @@
...
+        self.x = torch.randn(128, 4)
+
+        self.q_lin = QuantLinear(self.fp32)  # default uint8
+
+    def test_mode_transitions(self):
+        self.assertIs(self.q_lin._mode, Mode.NO_QUANT)
+
+        # Calibration (re-collect static weight range right here)
+        self.q_lin.enable_calibration()
+        self.q_lin.weight_obs.collect(self.fp32.weight)  # restore after reset
```
seockho-kim, line 39: can you explain `restore after reset`? This is included in enable_calibration(), but you might ensure this again.

mhs4670go: Seems redundant.

dayo09 (approved):
LGTM

PR #292, opened 2025-08-25
[ptq] Fix double iteration of observer

mhs4670go:
This commit fixes double interation issue of PTQ wrapper.

Before this PR, there's a double interation bug when a parent wrapper class iterates subsequent observers. Because the submodule includes not only a quant wrapper module but also its `PTQWrapper` that wraps the module.


dayo09 (approved):
LGTM

seockho-kim (approved):
LGTM

PR #305, opened 2025-09-01
[ptq] Support PPL computation

mhs4670go:
This commit supports PPL computation.

```bash
python tico/experimental/quantization/ptq/examples/compare_ppl.py
Calibrating with 50,000 tokens.

Creating UINT-8 clone …
`rope_scaling`'s original_max_position_embeddings field must be less than max_position_embeddings, got 2048 and max_position_embeddings=2048
Using the latest cached version of the dataset since wikitext couldn't be found on the Hugging Face Hub
Found the latest cached dataset configuration 'wikitext-2-raw-v1' at /home/sw4670.chae/.cache/huggingface/datasets/wikitext/wikitext-2-raw-v1/0.0.0/b08601e04326c79dfdd32d625aee71d232d685c3 (last modified on Fri May  3 21:06:41 2024).
Calibration: 100%|█████████████████████████████████████████████████████████████████████████████████████████████| 22/22 [00:03<00:00,  6.20it/s]

Calculating perplexities …
Using the latest cached version of the dataset since wikitext couldn't be found on the Hugging Face Hub
Found the latest cached dataset configuration 'wikitext-2-raw-v1' at /home/sw4670.chae/.cache/huggingface/datasets/wikitext/wikitext-2-raw-v1/0.0.0/b08601e04326c79dfdd32d625aee71d232d685c3 (last modified on Fri May  3 21:06:41 2024).
Token indices sequence length is longer than the specified maximum sequence length for this model (289077 > 131072). Running this sequence through the model will result in indexing errors
PPL:  99%|██████████████████████████████████████████████████████████████████████████████████████████████████▎| 561/565 [07:52<00:03,  1.19it/s]

│ UINT-8 :  9969.43
```

### DIFF
--- a/tico/experimental/quantization/ptq/utils/metrics.py
+++ b/tico/experimental/quantization/ptq/utils/metrics.py
@@ -25,6 +25,7 @@ def perplexity(
     *,
     max_length: Optional[int] = None,
     stride: int = 512,
+    ignore_index: int | None = -100,
     show_progress: bool = True,
 ) -> float:
     """
@@ -53,6 +54,11 @@ def perplexity(
     stride : int, default = 512
         Step size by which the sliding window advances.  Must satisfy
         `1 ≤ stride ≤ max_length`.
+    ignore_index : int, default = -100
+        Label value to ignore in loss computation. This should match
+        the `ignore_index` used by the model's internal
+        `CrossEntropyLoss`. For Hugging Face causal LMs, the
+        convention is `-100`.
     show_progress : bool, default = True
         If True, displays a tqdm progess bar while evaluating.
 
@@ -92,7 +98,7 @@ def perplexity(
 
         input_ids = input_ids_full[:, begin:end]
         target_ids = input_ids.clone()
-        target_ids[:, :-trg_len] = -100  # mask previously-scored tokens
+        target_ids[:, :-trg_len] = ignore_index  # mask previously-scored tokens
 
         with torch.no_grad():
             outputs = model(input_ids, labels=target_ids)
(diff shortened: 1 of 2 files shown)


tico/experimental/quantization/ptq/utils/metrics.py
```diff
@@ -0,0 +1,117 @@
...
+        target_ids = input_ids.clone()
+        target_ids[:, :-trg_len] = -100  # mask previously-scored tokens
+
+        with torch.no_grad():
+            outputs = model(input_ids, labels=target_ids)
+            # loss is already averaged over non-masked labels
+            neg_log_likelihood = outputs.loss
+
+        # exact number of labels that contributed to loss
+        loss_tokens = (target_ids[:, 1:] != -100).sum().item()
```
dayo09, line 103: How about exposing an api to set this `ignore_index` from outer side, and name it as `ignore_index`?
I see that your `DummyLM` carefully set the value as `ignore_index` for CrossEntropyLoss function. However some model can have different `ignore_index` value.

In the same context, meta llama may use another ignore_index, so it's safe to get one from tokenizer's `pad_token_id`.
So I suggest to

1. expose ignore_index for perplexity function.
2. set pad_token_id as ignore_index in example meta's case

mhs4670go: Good catch! 

I think exposing `ignore_index` at the API level makes sense. Different models (or loss setups) may not always use -100, so allowing it to be passed in makes the function more general and future-proof.

One thing to be careful about: using `pad_token_id` as `ignore_index` is not generally correct for Hugging Face causal LMs (including Meta LLaMA). By convention, they always use -100 as the masked-label value for `CrossEntropyLoss`, while `pad_token_id` is still a valid token ID and should not be ignored. So this is what I'll do.

1. Expose `ignore_index` as a parameter (default -100).
2. Only use `pad_token_id` as `ignore_index` if the specific model is designed that way. For HF causal LMs, keep -100.
3. If `ignore_index=None`, auto-detect from `model.config.ignore_index` when available, otherwise fall back to -100.

This way the API is flexible for non-standard models, but stays consistent with the standard HF behavior. Thanks for the review.

dayo09 (approved):
LGTM Thanks!

seockho-kim (approved):
LGTM

PR #309, opened 2025-09-03
[ptq] Do not use __all__ in __init__.py

mhs4670go:
This commit removes __all__ from __init__.py.

Using `__all__` in `__init__.py` to import all modules has clear pros and cons. It simplifies the public API where users can access everything directly from the package root. Nevertheless, it increases the risk of circular imports and loads unnecessary dependencies, slowing startup and increasing memory use. Moreover, it also pollutes the namespace, raising chances of name collisions.

Therefore, I've removed them and I'll expose only the intended public API, not every internal module later. This keeps the package clean, avoids circular import issues, and improves maintainability.


seockho-kim (approved):
LGTM

PR #311, opened 2025-09-03
[ptq] Add an example that does PTQ with GPTQ

mhs4670go:
This commit adds an example that does PTQ with GPTQ.

```bash
python tico/experimental/quantization/ptq/examples/quantize_with_gptq.py 
Loading FP model …
Applying GPTQ …
GPTQ calibration: 100%|███████████████████████████████████████████████████████████████████████████████████| 4358/4358 [00:04<00:00, 902.18it/s]
Quantizing layers: 100%|██████████████████████████████████████████████████████████████████████████████████████| 8/8 [01:29<00:00, 11.23s/layer]
Token indices sequence length is longer than the specified maximum sequence length for this model (12034 > 2048). Running this sequence through the model will result in indexing errors
Act-calibration: 100%|█████████████████████████████████████████████████████████████████████████████████████████| 24/24 [00:01<00:00, 23.80it/s]

Calculating perplexities …
PPL:  99%|██████████████████████████████████████████████████████████████████████████████████████████████████▍| 630/634 [01:15<00:00,  8.39it/s]

│ UINT-8 :  9149.92
```

### DIFF
--- a/tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
+++ b/tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
@@ -1,0 +1,165 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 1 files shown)


seockho-kim (approved):
LGTM

dayo09 (approved):
LGTM

PR #312, opened 2025-09-03
[ptq] Support multiple registration

mhs4670go:
This commit supports multiple registration.

This fixes an error that happened when old `transformers` use `LlamaSdpaAttention` instead of `LlamaAttention` which is used latest `transformers`.


dayo09 (approved):
LGTM

PR #315, opened 2025-09-05
[quantization] Support output hook

mhs4670go:
This commit supports output hook option and add fairseq logic as an example.

### DIFF
--- a/test/quantization/algorithm/test_smooth_quant.py
+++ b/test/quantization/algorithm/test_smooth_quant.py
@@ -13,10 +13,14 @@
 # limitations under the License.
 
 import os
+import sys
+import types
 import unittest
 
 import numpy as np
 import torch
+import torch.nn as nn
+import torch.nn.functional as F
 
 from tico.experimental.quantization import convert, prepare
 from tico.experimental.quantization.config import SmoothQuantConfig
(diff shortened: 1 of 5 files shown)


tico/experimental/quantization/algorithm/smoothquant/quantizer.py
```diff
@@ -30,13 +30,33 @@
...
+    Q) Why allow choosing between input and output activations?
+
+    SmoothQuant relies on channel-wise activation statistics to balance
+    weights and activations. In practice, there are two natural sources:
+
+    - "input": captures the tensor right before a Linear layer
+      (forward-pre-hook). This matches the original SmoothQuant paper
+      and focuses on scaling the raw hidden state.
+
+    - "output": captures the tensor right after a Linear layer
```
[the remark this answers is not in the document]
mhs4670go: It's more like smoothing before activation functions (fc1 output). Below `_apply_if_fairseq_relu_bridge` function is one of the example (fc1-relu-fc2). By setting "output", we can collect the outputs of fc1, instead of the inputs of the fc2.

SmoothQuant’s scale s is defined from the per-channel activation range of the input to the target `Linear`. When getting the amax on the `ReLU` output (post-activation), negatives were zeroed, systematically underestimating amax. This drives s << 1 and thus 1/s >> 1, so the fusion of the scale effectively amplified fc1 outputs, causing huge diffs.

Therefore, measuring amax on the pre-ReLU tensor (fc1 output) reflects the true input distribution to fc2, yields well-behaved s (often ≥ 1 for outlier channels).

test/quantization/algorithm/test_smooth_quant.py
```diff
@@ -75,3 +79,178 @@ def test_value(self):
             atol=1e-5,
             err_msg=f"Value mismatches.\nbefore result: {base_output}\nafter result: {target_output}",
         )
+
+
+# ────────────────────────────────────────────────────────────
+# Faux fairseq injection (so the handler path works w/o fairseq)
```
jinevening, line 85: Just curiosity. Is the term "faux" is used somewhere? I've seen "mock" or "fake" more frequently.

test/quantization/algorithm/test_smooth_quant.py
```diff
@@ -75,3 +79,178 @@ def test_value(self):
...
+        # Baseline forward & weight snapshots
+        base_out = model(x)
+        assert hasattr(model.encoder.layer, "fc1")
+        assert hasattr(model.encoder.layer, "fc2")
+        assert isinstance(model.encoder.layer.fc1, nn.Linear)
+        assert isinstance(model.encoder.layer.fc2, nn.Linear)
+        base_fc1_w = model.encoder.layer.fc1.weight.detach().clone()
+        base_fc2_w = model.encoder.layer.fc2.weight.detach().clone()
+        base_fc1_b = model.encoder.layer.fc1.bias.detach().clone()
+        base_fc2_b = model.encoder.layer.fc2.bias.detach().clone()
```
miusic, line 201:
```suggestion
```
(optional)
It seems not to be used
How about to remove the unused variable?

test/quantization/algorithm/test_smooth_quant.py
```diff
@@ -75,3 +79,178 @@ def test_value(self):
...
+
+        # 2) Weights must have changed (row/col scaling fused)
+        assert hasattr(model_sq.encoder.layer, "fc1")
+        assert hasattr(model_sq.encoder.layer, "fc2")
+        assert isinstance(model_sq.encoder.layer.fc1, nn.Linear)
+        assert isinstance(model_sq.encoder.layer.fc2, nn.Linear)
+        new_fc1_w = model_sq.encoder.layer.fc1.weight.detach()
+        new_fc2_w = model_sq.encoder.layer.fc2.weight.detach()
+        new_fc1_b = model_sq.encoder.layer.fc1.bias.detach()
+        new_fc2_b = model_sq.encoder.layer.fc2.bias.detach()
```
miusic, line 236:
```suggestion
```
ditto

miusic (approved):
LGTM with minor comments 😄

dayo09 (approved):
LGTM

test/quantization/algorithm/test_smooth_quant.py
mhs4670go: A term is suggested by a generative model :smile:.

test/quantization/algorithm/test_smooth_quant.py
mhs4670go: I'll apply this in the next PR! Thanks.

PR #317, opened 2025-09-08
[ptq] Introduce fairseq encoder layer wrapper

mhs4670go:
This commit introduces a wrapper for fairseq encoder layer.

### DIFF
--- a/test/quantization/ptq/wrappers/fairseq/test_quant_encoder_layer.py
+++ b/test/quantization/ptq/wrappers/fairseq/test_quant_encoder_layer.py
@@ -1,0 +1,257 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


test/quantization/ptq/wrappers/fairseq/test_quant_encoder_layer.py
```diff
@@ -0,0 +1,257 @@
...
+
+    # 2) Mask handling: boolean causal attn_mask and boolean key_padding_mask
+    def test_mask_handling(self):
+        layer, _ = self._build_layer(True, False, QuantConfig())
+        x = self._make_inputs()
+        # Upper-triangular causal mask [T,S] (True = masked)
+        attn_mask_bool = torch.triu(
+            torch.ones(self.T, self.T, dtype=torch.bool), diagonal=1
+        )
+        # Key padding mask [B,S] (True = pad)
```
dayo09, line 141: Q. About abbreviation `T` and `S`, it seems both refers to sequence length but what's difference?

jinevening (approved):
LGTM

BTW, have you tested whether the exported circle is compiled by npu compiler?

test/quantization/ptq/wrappers/fairseq/test_quant_encoder_layer.py
mhs4670go: `T` refers to target sequence length and `S` refers to source sequence length. 

When it comes to the comments in the codes, `attn_mask` has shape (T, S), where T = query length and S = key length. And, `key_padding_mask` is (B, S), since it marks padding positions in the key sequence for each batch.

mhs4670go:
> BTW, have you tested whether the exported circle is compiled by npu compiler?

All I pushed have tested with npu compiler. You can refer to my personal repo's issue number 18.

dayo09 (approved):
LGTM, thanks for the explanation!

PR #321, opened 2025-09-09
[ptq] Introduce fairseq encoder wrapper

mhs4670go:
This commit introduces a wrapper for fairseq encoder.

### DIFF
--- a/tico/experimental/quantization/ptq/wrappers/fairseq/quant_encoder.py
+++ b/tico/experimental/quantization/ptq/wrappers/fairseq/quant_encoder.py
@@ -1,0 +1,333 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


tico/experimental/quantization/ptq/wrappers/fairseq/quant_encoder.py
```diff
@@ -0,0 +1,333 @@
...
+        x = self.embed_scale * token_embedding
+        if self.embed_positions is not None:
+            x = x + self.embed_positions(src_tokens)
+        if self.layernorm_embedding is not None:
+            x = self.layernorm_embedding(x)
+        # No dropout, no quant_noise here (inference-only)
+        return x, embed
+
+    # ----------------------------------------------------------------------
+    def forward(
```
mhs4670go, line 140: The original implementation is in [here](https://github.com/facebookresearch/fairseq/blob/d13e14a800bb588e5a77fb4e551f554ff9b24a72/fairseq/models/transformer/transformer_encoder.py#L173).

Issue #323, opened 2025-09-10
Use `torch.testing.assert_close` to show mismatched index


periannath:
# What

Use `torch.testing.assert_close` to report the index of mismatched elements.

# Why

When a test fails, the current output prints long messages with entire tensor contents. What we actually need is the **index of the mismatch**, not the full tensors.

`torch.testing.assert_close` reports the tensor indices with the largest differences, making it much easier to locate and debug the specific elements that diverge.

## To-be

```bash

TEST FAILED at 'validate_result ...'
 Value mismatches.

Tensor-likes are not close!

Mismatched elements: 1 / 453120 (0.0%)
Greatest absolute difference: 0.0028784796595573425 at index (0, 8, 100, 10) (up to 0.002 allowed)
Greatest relative difference: 0.07503077387809753 at index (0, 8, 100, 10) (up to 0.002 allowed)
Traceback (most recent call last):
  File "/home/jhyeo/workspace/TICO/./scripts/value_check_test.py", line 133, in value_test
    is_valid(torch_result, circle_result, **tolerance)
  File "/home/jhyeo/workspace/TICO/scripts/ci_utils.py", line 73, in is_valid
    validate_result(torch_result, circle_result, rtol, atol)
  File "/home/jhyeo/workspace/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 57, in wrapper
    raise e
  File "/home/jhyeo/workspace/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 54, in wrapper
    return function(*args, **kwargs)
  File "/home/jhyeo/workspace/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 231, in validate_result
    torch.testing.assert_close(
  File "/home/jhyeo/miniforge3/envs/tico-dev/lib/python3.10/site-packages/torch/testing/_comparison.py", line 1530, in assert_close
    raise error_metas[0].to_error(msg)
AssertionError: Value mismatches.

Tensor-likes are not close!

Mismatched elements: 1 / 453120 (0.0%)
Greatest absolute difference: 0.0028784796595573425 at index (0, 8, 100, 10) (up to 0.002 allowed)
Greatest relative difference: 0.07503077387809753 at index (0, 8, 100, 10) (up to 0.002 allowed)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/jhyeo/workspace/TICO/./scripts/value_check_test.py", line 151, in <module>
[...]

PR #324, opened 2025-09-10
[test] Use torch.testing.assert_close for value check

periannath:
This commit uses torch.testing.asser_close for value check

---

For issue #323

### DIFF
--- a/test/pt2_to_circle_test/test_pt2_to_circle.py
+++ b/test/pt2_to_circle_test/test_pt2_to_circle.py
@@ -227,14 +227,22 @@ def validate_result(
                 desired=circle_res.shape,
                 err_msg=f"Shape mismatches.\nexpected result: {expected_res.shape}\ncircle result: {circle_res.shape}",
             )
-        np.testing.assert_allclose(
-            actual=circle_res,
-            desired=expected_res,
-            rtol=rtol,
-            atol=atol,
-            err_msg=f"Value mismatches.\nexpected result: {expected_res}\ncircle result: {circle_res}",
-        )
-
+            circle_tensor = torch.from_numpy(circle_res)
+            torch.testing.assert_close(
+                actual=expected_res,
+                expected=circle_tensor,
+                rtol=rtol,
+                atol=atol,
+                msg=lambda msg: f"Value mismatches.\n\n{msg}"
+            )
+        else:
+            np.testing.assert_allclose(
+                actual=circle_res,
+                desired=expected_res,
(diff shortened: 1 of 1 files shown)


test/pt2_to_circle_test/test_pt2_to_circle.py
```diff
@@ -227,14 +227,22 @@ def validate_result(
...
-        np.testing.assert_allclose(
-            actual=circle_res,
-            desired=expected_res,
-            rtol=rtol,
-            atol=atol,
-            err_msg=f"Value mismatches.\nexpected result: {expected_res}\ncircle result: {circle_res}",
-        )
-
+            circle_tensor = torch.from_numpy(circle_res)
+            torch.testing.assert_close(
```
periannath, line 231: Use `torch.testing.assert_close` when `expected_res` is instance of `torch.Tensor`.

test/pt2_to_circle_test/test_pt2_to_circle.py
```diff
@@ -227,14 +227,23 @@ def validate_result(
...
+                actual=circle_tensor,
+                expected=expected_res,
+                equal_nan=True,
```
periannath, line 234: > NaN’s are only considered equal to each other if equal_nan is True.
From https://docs.pytorch.org/docs/stable/testing.html

I found that some unittests return `NaN` value.

jinevening (approved):
LGTM

test/pt2_to_circle_test/test_pt2_to_circle.py
periannath: I've updated a little bit to use `torch.testing.assert_close` in all cases.

One side-effect is that error message was changed, as `assert_close` also supports `dtype` check.

## Value mismatch log
```bash
TEST FAILED at 'validate_result ...'
 Tensor-likes are not close!

Mismatched elements: 1 / 453120 (0.0%)
Greatest absolute difference: 0.0028784796595573425 at index (0, 8, 100, 10) (up to 0.002 allowed)
Greatest relative difference: 0.08111704885959625 at index (0, 8, 100, 10) (up to 0.002 allowed)
```

## dtype mismatch log
```bash
TEST FAILED at 'validate_result ...'
 The values for attribute 'dtype' do not match: torch.uint32 != torch.int64.
```

miusic (approved):
LGTM

mhs4670go (approved):
LGTM

jinevening (approved):

PR #327, opened 2025-09-10
[ptq] Make compare_ppl script configurable

mhs4670go:
This commit introduces an argument parser to compare_ppl script.

This is necessary for integrating example scripts into the CI.

```bash
python tico/experimental/quantization/ptq/examples/compare_ppl.py --model Maykeye/TinyLLama-v0
=== Config ===
Mode             : fp
Model            : Maykeye/TinyLLama-v0
Device           : cuda
DType            : float32
Stride           : 512
Use HF cache?    : False
Calib preset     : debug

`torch_dtype` is deprecated! Use `dtype` instead!

Calculating perplexities …
Using the latest cached version of the dataset since wikitext couldn't be found on the Hugging Face Hub
Found the latest cached dataset configuration 'wikitext-2-raw-v1' at /home/sw4670.chae/.cache/huggingface/datasets/wikitext/wikitext-2-raw-v1/0.0.0/b08601e04326c79dfdd32d625aee71d232d685c3 (last modified on Fri May  3 21:06:41 2024).
Token indices sequence length is longer than the specified maximum sequence length for this model (324381 > 2048). Running this sequence through the model will result in indexing errors
PPL:  99%|██████████████████████████████████████████████████████████████████████████████████████████████████▍| 630/634 [00:16<00:00, 38.89it/s]

│ FP     :  7609.58
```

### DIFF
--- a/tico/experimental/quantization/ptq/examples/compare_ppl.py
+++ b/tico/experimental/quantization/ptq/examples/compare_ppl.py
@@ -20,6 +20,10 @@
 #   • Full post-training UINT-8 flow (wrap → calibrate → eval).
 # =============================================================================
 
+import argparse
+import sys
+from typing import Optional
+
 import torch
 import tqdm
 from datasets import load_dataset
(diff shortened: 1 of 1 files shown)


seockho-kim (approved):
LGTM

Issue #322, opened 2025-09-09
Increase test coverage with Cline


jinevening:
## What

Let's increase test coverage.

## Why

The current test coverage is 84%. Let's increase the test coverage. Goal is to exceed 90%.

```
./ccex coverage -f txt
```

```
Name                                                                                                             Stmts   Miss  Cover   Missing
----------------------------------------------------------------------------------------------------------------------------------------------
test/__init__.py                                                                                                     0      0   100%
test/modules/__init__.py                                                                                             0      0   100%
test/modules/base.py                                                                                                10      1    90%   8
test/modules/model/BitNet_b1_58/__init__.py                                                                          0      0   100%
...
TOTAL                                                                                                            20619   3333    84%
```

jinevening:
Here is the cline's report about the test coverage.

## Prompt

```
@/test/reports/cov/coverage.txt 를 보고 coverage 가 낮은 파일들의 목록을 만들고 coverage 를 높일 수 있는 unit test 를 작성하는 계획을 세워줘
```

## Report

The user has provided a coverage report file (`test/reports/cov/coverage.txt`) and wants me to:

1. Create a list of files with low coverage
2. Create a plan to write unit tests to increase coverage for those files

I need to analyze the coverage report to identify files with low coverage percentages and then create a plan for writing unit tests.

Let me first analyze the coverage report to identify the files with the lowest coverage. I'll look for files with coverage below a certain threshold (e.g., below 80% or 70%).

From the coverage report, I can see several files with very low coverage:

1. `tico/experimental/quantization/algorithm/gptq/gptq.py` - 12% coverage (107 stmts, 94 miss)
2. `tico/experimental/quantization/algorithm/gptq/quant.py` - 10% coverage (103 stmts, 93 miss)
3. `tico/experimental/quantization/algorithm/gptq/quantizer.py` - 15% coverage (110 stmts, 94 miss)
4. `tico/experimental/quantization/algorithm/gptq/utils.py` - 22% coverage (18 stmts, 14 miss)
5. `tico/experimental/quantization/evaluation/evaluate.py` - 24% coverage (70 stmts, 53 miss)
6. `tico/experimental/quantization/evaluation/utils.py` - 20% coverage (64 stmts, 51 miss)
7. `tico/experimental/quantization/ptq/utils/introspection.py` - 15% coverage (59 stmts, 50 miss)
[...]

PR #331, opened 2025-09-11
[test] Add tests for quantization/evaluation

jinevening:
This adds tests for quantization/evaluation.py

---
Related to: #322


test/quantization/evaluation/test_evaluation.py
```diff
@@ -0,0 +1,593 @@
...
+    compare_layer_outputs,
+    save_fp_outputs,
+)
+from tico.experimental.quantization.ptq.wrappers.quant_module_base import (
+    QuantModuleBase,
+)
+from tico.utils.model import CircleModel
+
+
+class TestEvaluateFunctions(unittest.TestCase):
```
seockho-kim, line 48: How about `TestEvaluate` instead of `TestEvaluateFunctions`?

glistening (approved):
LGTM. I am okay for the both names.

test/quantization/evaluation/test_evaluation.py
jinevening: The class name was selected for consistency with other tests. Let me update the name when necessary. Thanks for the review!

Issue #339, opened 2025-09-12
batchmatmul with lhs constant to fullyconnected


glistening:
It is preferred at least for onert to use fullyconnected if lhs is const.

onert (and similarly tflite) does not allow constant for batchmatmul lhs.

See https://github.com/Samsung/ONE/issues/16064.

In this case, fullyconnected provides more efficient kernels.

I tried to emit fulyconnected instead of batchmatmul.

Perhaps, `op_bmm.py`

```py
def define_node(
        self,
        node: torch.fx.Node,
    ) -> circle.Operator.OperatorT:
        args = BmmArgs(*node.args, **node.kwargs)  # type: ignore[arg-type]
        input = args.input
        mat2 = args.mat2

        // TODO: check input is constant
        // Then create FULLY_CONNECTED instead of BATCH_MATMUL

        op_index = get_op_index(
            circle.BuiltinOperator.BuiltinOperator.BATCH_MATMUL, self._op_codes
        )

        inputs = [input, mat2]
        outputs = [node]

        operator = create_builtin_operator(self.graph, op_index, inputs, outputs)
```

However, I am not familiar with TICO yet.

Could anyone help me?

dayo09:
> pass config to serializer

Do you still need it? For what purpose..?

glistening:
> Do you still need it? For what purpose..?

Perhaps I will use your PR #341 as soon as it is finished.
However, as of this writing, your PR does not produce what I want yet.


`BEFORE`

<img width=65% src="https://github.com/user-attachments/assets/a276121a-144c-4ce0-aaaa-e9aea42b5675">

`#344` 

(same)

It does not solve https://github.com/Samsung/ONE/issues/16064 yet.


##


In addition, there are other reasons to write my own patch in my way:

1) To get the needed circle in advance before you complete the PR
2) To know the AI assistant works well.
  - At a first glance, it is better than I expected. It seems to work.
  - After 1 hour, it is useless, it does not succeed to finish the work.
  - Maybe I need to try better AI assistant.
3) To understand more about TICO.
  - For example I've found TICO already generate two (or more) circle nodes. It is the opposite of my understand yesterday. https://github.com/Samsung/TICO/issues/339#issuecomment-3290318821
  - In `op_mm.py`, 1 aten op is serialized into 2 or more circle operators.



## 

With https://github.com/Samsung/TICO/commit/41dd277e7ff8a0440e173b567c8cfba4e5c98e2a

[...]

PR #341, opened 2025-09-12
[passes] Add ConvertMatmulToLinear pass

dayo09:
Let's add convert matmul to linear pass.
This commit...
 refactors mm serialization logic and make convert_matmul_to_linear pass
 introduces new CompileConfig attribute convert_lhs/rhs_const_mm_to_fc.

---

For #339


tico/passes/convert_matmul_to_linear.py
```diff
@@ -0,0 +1,212 @@
...
+class ConvertMatmulToLinear(PassBase):
+    """
+    This pass converts matmul to linear selectively
+
+    How to select between `matmul` and `linear`?
+
+    * Linear has better quantization accuracy (NPU backend)
+        Due to ONE compiler's quantization policy;
+        FullyConnected(=Linear) uses per-channel quantization for weight and per-tensor for input.
+        BatchMatmul(=matmul) uses per-tensor quantization for both rhs and lhs.
```
jinevening, line 143: FYI, a new generation of NPU would support cwq for matmul.

tico/serialize/operators/op_mm.py
```diff
@@ -57,122 +57,16 @@ def set_bmm_option(operator):
...
-        TODO set prior_latency outside
-        """
         args = MatmulArgs(*node.args, **node.kwargs)  # type: ignore[arg-type]
         input = args.input
         other = args.other
 
         inputs = [input, other]
         outputs = [node]
 
-        if not is_const(other) and prior_latency:
```
jinevening, line 173: prior_latency is not used anymore?

test/modules/op/mm.py
```diff
@@ -27,3 +30,64 @@ def forward(self, lhs, rhs):
...
+
+    def forward(self, rhs):
+        out = torch.mm(self.weight, rhs)
+        return out
+
+    def get_example_inputs(self):
+        return (torch.randn(4, 5),), {}
+
+    def get_compile_config(self):
+        return CompileConfigV1(convert_lhs_const_mm_to_fc=True)
```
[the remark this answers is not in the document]
seockho-kim: @dayo09 Could you improve it to handle bmm, too?

``` py
@tag.use_onert
class BmmTest(TestModuleBase):
    def __init__(self):
        super().__init__()
        self.weight = torch.randn(2, 3, 4)

    def forward(self, rhs):
        out = self.weight @ rhs
        return out

    def get_example_inputs(self):
        return (torch.randn(2, 4, 5),), {}

    def get_compile_config(self):
        return CompileConfigV1(convert_lhs_const_mm_to_fc=True)
```

tico/passes/convert_matmul_to_linear.py
```diff
@@ -0,0 +1,212 @@
...
+            if is_lifted_tensor_constant(exported_program, lhs):
+                return True
+            elif is_param(exported_program, lhs):
+                return True
+            elif is_buffer(exported_program, lhs):
+                return True
+            else:
+                return False
+
+    def convert(self, exported_program, node) -> torch.fx.Node:
```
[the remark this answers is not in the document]
dayo09: Yes, the codes are duplicate 🤔 Let me extract it and make two functions share it.

tico/serialize/operators/op_mm.py
dayo09: @jinevening Yes, the old feature is basically part of the new one.

```py
# BEFORE prior_latency==False (default)
# AFTER (default)
If rhs is const: conversion ON
else: conversion OFF

# BEFORE prior_latency==True 
# AFTER convert_rhs_const_mm_to_fc ==False 
always: conversion OFF
```

tico/passes/convert_matmul_to_linear.py
dayo09: @jinevening Do you mean 3rd generation?

test/modules/op/mm.py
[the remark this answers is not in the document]
dayo09: @seockho-kim Above case is not supported because matmul to fc conversion can be done only if weight is 2dim. Circle FullyConnected operation assumes its weight to be in rank 2.

tico/passes/convert_matmul_to_linear.py
```diff
@@ -0,0 +1,212 @@
...
+                graph,
+                torch.ops.aten.linear.default,
+                args=(lhs, transpose_node),
+            )
+            node.replace_all_uses_with(fc_node, propagate_meta=True)
+
+        return fc_node
+
+
+class ConvertLhsConstMatmulToLinear(Converter):
```
mhs4670go, line 88: Why should it consider lhs and rhs? The left and right matter when converting?

dayo09: @mhs4670go onert doesn't run lhs const matmul. So this PR converts matmul to fullyconnected for onert, conditionally with config

mhs4670go:
@glistening I saw [this comment](https://github.com/Samsung/ONE/issues/16064#issuecomment-3294679564). Is this pass still needed?

glistening:
> > I saw [Samsung/ONE#16064 (comment)](https://github.com/Samsung/ONE/issues/16064#issuecomment-3294679564). Is this pass still needed?

@dayo09 Sure. I need this PR. The maintainer of `ONERT` will keep the constraints (not allowing lhs const). My PR and commenting out is a workaround to solve the next steps in onert.

jinevening (approved):
LGTM

test/modules/op/mm.py
```diff
@@ -27,3 +30,64 @@ def forward(self, lhs, rhs):
...
+    def get_example_inputs(self):
+        return (torch.randn(3, 4),), {}
+
+
+@use_onert
+@test_negative(expected_err="NNFW_STATUS_ERROR")
+class SimpleMatmulConstLhsOnert(TestModuleBase):
+    """ """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
mhs4670go: I think it would be good to describe what error is expected. `NNFW_STATUS_ERROR` is a bit ambiguous.

tico/config/v1.py
```diff
@@ -20,6 +20,8 @@
 @dataclass
 class CompileConfigV1(CompileConfigBase):
     legalize_causal_mask_value: bool = False
+    convert_lhs_const_mm_to_fc: bool = False
+    convert_rhs_const_mm_to_fc: bool = True
```
mhs4670go, line 24: On second thought, just `convert_const_mm_to_fc` could be simpler choice. Do you have any reasons that chose this design?

glistening:
If I understand correctly, TICO will generate `lhs const matmul` to `linear + transpose1 + transpose2`. I don't fully understand why it happens. Maybe for keeping `TICO` simple. However, it requires yet-another tool like `circle2circle`. If so, it would better to add a pass in `circle2circle` to fuse `batchmatmul` + `transpose` to 1 fullyconnected.

My question: Who and why `batchmatmul` + `transpose` is generated for matmul? Does the initial aten graph has `batchmatmul` + `transpose`? Or TICO does?

test/modules/op/mm.py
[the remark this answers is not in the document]
dayo09: That is how onert throws. It should match.

mhs4670go (dismissed):
LGTM

glistening:
Hmm. I succeed to `gh pr checkout 341`.
I converted my model, but there is no change.
It still emit `batchmatmul`.

dayo09:
@@jinevening @seockho-kim @mhs4670go It's rebased. PTAL  again 😅

seockho-kim:
> Hmm. I succeed to `gh pr checkout 341`. I converted my model, but there is no change. It still emit `batchmatmul`.
> 
> @seockho-kim Does this PR solve the same issue in gemma3?

This pr does not solve issue in gemma3, because bmm is not changed.
I'm waiting for another PR.

seockho-kim (approved):
LGTM

dayo09:
On a second thought, just adding `1-batch bmm to mm` may include redundant reshape, so it's not a general pass. 

@jinevening @mhs4670go Do you think conversion of `1-batch bmm to mm` is allowed or may I introduce the pass optionally for `matmul to linear` case?

mhs4670go:
> Do you think conversion of 1-batch bmm to mm is allowed or may I introduce the pass optionally for matmul to linear case?

I think it should be optional because it's not kind of circle legalization.

glistening:
If `TICO` by its own cannot convert my case at once, it would be good to write a pass using circle level optimizer. ( I don't think `circle2circle` has such pass, and I don't like circle2circle — c++ solution. ) Maybe it would be better to introduce my circle level graph optimizer written in python.

PR #342, opened 2025-09-15
Support ccex format -d

jinevening:
This adds aliasing -d for --diff-only.


miusic (approved):
LGTM

Issue #343, opened 2025-09-15
Remove constant input during Circle export


periannath:
# What

Let's remove constant input during Circle exprot.

# Why

`torch.export` embeds constant arguments within the graph.[^1]

However, `circle_serializer` saves constant argument as input of graph, thus makes dangling input.

```python
    def get_example_inputs(self):
        hidden_states = torch.Tensor(np.random.rand(5 * 2048)).reshape(
            1, 5, 2048)
        position_ids = torch.arange(5).unsqueeze(0).repeat(1, 1).long()
        return (), {
            "hidden_states": hidden_states,
            "position_ids": position_ids,
            "use_cache": True,
        }
```

Above script makes `use_cache` as dangling input.

[^1]: https://docs.pytorch.org/docs/stable/export.ir_spec.html#call-function

periannath:
Added as option in `CompileConfigV1`. Please set `remove_constant_input` to `True` to remove constant inputs.

PR #347, opened 2025-09-16
[serialize] Remove constant input during circle export

periannath:
This commit removes constant input during circle export

---

For issue #343

### DIFF
--- a/test/modules/model/SmolVLM_text_model/model.py
+++ b/test/modules/model/SmolVLM_text_model/model.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import torch
-from tico.config.v1 import CompileConfigV1
 from transformers import AutoModelForImageTextToText
 
 from test.modules.base import TestModuleBase
@@ -42,6 +41,3 @@ class SmolVLM_text_model(TestModuleBase):
             "return_dict": True,
         }
         return (), kwargs
-
-    def get_compile_config(self):
-        return CompileConfigV1(remove_constant_input=True)


periannath:
## SmolVLM_text_model

### Original

Image

- 6 inputs
  - 4 of them are constant and dangling
- 1 output

### This PR

Image

- 2 inputs
  - Constant inputs are removed
- 1 output

tico/serialize/circle_serializer.py
```diff
@@ -68,8 +68,8 @@ def build_circle(ep: ExportedProgram) -> bytes:
     for in_spec in ep.graph_signature.input_specs:
         if in_spec.kind != InputKind.USER_INPUT:
             continue
-        # NoneType ConstantArgument is ignored.
-        if isinstance(in_spec.arg, ConstantArgument) and in_spec.arg.value == None:
+        # ConstantArgument is ignored.
+        if isinstance(in_spec.arg, ConstantArgument):
```
periannath, line 72: Main change of this PR.

mhs4670go:
First of all, the reason why the unittest failure has been resolved after moving the constants to kwargs was because `ModelInputSpec` didn't parse keyword arguments properly. It should be fixed. So, we should choose another approach.

When scalars like bool or int are passed into `torch.export.export`, they appear as `ConstantArguments` in the input signature, even if the export has already specialized away the branch (e.g., only the True path remains).

The key reason is that these constants act as guards and those are two main benefits.

1. Correctness & reproducibility

The exported graph **assumes the example values**. If a different value is passed later, the original function may have taken a different branch that no longer exists in the graph. Keeping it in the signature ensures the runtime can detect mismatches instead of silently executing an invalid program.

2. Backend & cache stability

Backends like Inductor/XLA/TVM use the full input signature for cache keys. Including these constants ensures compiled results are reused only for the same values, and a re-export happens if they change.

So even if the constant isn’t used in the graph body anymore, it stays in the signature to protect against invalid executions and to stabilize backend caching.

Then, what should we do with this? AFAIK, there are two practical tips. If you really want to remove such constants from the input signature (and bake them directly into the graph), you can:

- Store them as module attributes (e.g., self.use_cache = True), so they become internal constants instead of inputs.

```python
class Model(torch.nn.Module):
    def forward(self, x, use_cache: bool):
        if use_cache:
            ...
        else:
            ...
        return out

m = Model().eval()

class Wrapped(torch.nn.Module):
    def __init__(self, m, use_cache=True):
        super().__init__()
        self.m = m
        self.use_cache = use_cache  # NOTE HERE

    def forward(self, x):
        return self.m(x, self.use_cache)

wrapped = Wrapped(m, use_cache=True)
ep = torch.export.export(wrapped, (example_x,))
```

- Use `functools.partial` to pre-bind the argument before export.

```python
import functools
f = functools.partial(m.forward, use_cache=True)
ep = torch.export.export(f, (example_x,))
```

This way the constant is no longer part of the input signature—but note that if you ever want to change it, you’ll need to re-export.

mhs4670go:
First of all, the reason why the unittest failure has been resolved after moving the constants to kwargs was because `ModelInputSpec` didn't parse keyword arguments properly. It should be fixed. So, we should choose another approach.

When scalars like bool or int are passed into `torch.export.export`, they appear as `ConstantArguments` in the input signature, even if the export has already specialized away the branch (e.g., only the True path remains).

The key reason is that these constants act as guards and those are two main benefits.

1. Correctness & reproducibility

The exported graph **assumes the example values**. If a different value is passed later, the original function may have taken a different branch that no longer exists in the graph. Keeping it in the signature ensures the runtime can detect mismatches instead of silently executing an invalid program.

2. Backend & cache stability

Backends like Inductor/XLA/TVM use the full input signature for cache keys. Including these constants ensures compiled results are reused only for the same values, and a re-export happens if they change.

So even if the constant isn’t used in the graph body anymore, it stays in the signature to protect against invalid executions and to stabilize backend caching.

Then, what should we do with this? AFAIK, there are two practical tips. If you really want to remove such constants from the input signature (and bake them directly into the graph), you can:

- Store them as module attributes (e.g., self.use_cache = True), so they become internal constants instead of inputs.

```python
class Model(torch.nn.Module):
    def forward(self, x, use_cache: bool):
        if use_cache:
            ...
        else:
            ...
        return out

m = Model().eval()

class Wrapped(torch.nn.Module):
    def __init__(self, m, use_cache=True):
        super().__init__()
        self.m = m
        self.use_cache = use_cache  # NOTE HERE

    def forward(self, x):
        return self.m(x, self.use_cache)

wrapped = Wrapped(m, use_cache=True)
ep = torch.export.export(wrapped, (example_x,))
```

- Use `functools.partial` to pre-bind the argument before export.

```python
import functools
f = functools.partial(m.forward, use_cache=True)
ep = torch.export.export(f, (example_x,))
```

This way the constant is no longer part of the input signature—but note that if you ever want to change it, you’ll need to re-export.

mhs4670go:
That being said, removing constant inputs would be convenient. Therefore, I think we can introduce this feature as optional with `CompileConfig` feature.

periannath:
Thanks! I'll update this PR to add and use `remove_constant_args` flag.

test/modules/model/SmolVLM_text_model/model.py
```diff
@@ -41,3 +42,6 @@ def get_example_inputs(self):
             "return_dict": True,
         }
         return (), kwargs
+
+    def get_compile_config(self):
+        return CompileConfigV1(remove_constant_input=True)
```
periannath, line 47: Enabled this option only for `SmolVLM_text_model`.

tico/config/v1.py
```diff
@@ -20,6 +20,7 @@
 @dataclass
 class CompileConfigV1(CompileConfigBase):
```
periannath, line 21: Should I create `CompileConfigV2` instead of updating this?

mhs4670go: It's okay to update this class.

tico/serialize/circle_serializer.py
```diff
@@ -47,7 +48,7 @@ def _initialize_model() -> tuple[CircleModel, CircleSubgraph]:
     return model, graph
 
 
-def build_circle(ep: ExportedProgram) -> bytes:
+def build_circle(ep: ExportedProgram, config: CompileConfigBase) -> bytes:
```
mhs4670go, line 51: Setting default value keeps current behavior of `build_circle`. I think `test/unit_test/quantization_test/test_circle_executor.py` was modified because of this.

```suggestion
def build_circle(ep: ExportedProgram, config: CompileConfigBase = get_default_config()) -> bytes:
```

mhs4670go, line 51: Setting default value keeps current behavior of `build_circle`. I think `test/unit_test/quantization_test/test_circle_executor.py` was modified because of this.

```suggestion
def build_circle(ep: ExportedProgram, config: CompileConfigBase = get_default_config()) -> bytes:
```

test/modules/model/SmolVLM_text_model/model.py
```diff
@@ -41,3 +42,6 @@ def get_example_inputs(self):
             "return_dict": True,
         }
         return (), kwargs
+
```
mhs4670go, line 45: Hmm.. Could you exclude the change of this file? Current test infra can't test this feature because the example inputs and expected circle inputs are different.

periannath: Okay I'll remove those changes.

mhs4670go (approved):
LGTM

PR #350, opened 2025-09-17
[utils] Check the number of passed args

mhs4670go:
This commit checks the number of passed args when binding.

Related: https://github.com/Samsung/TICO/pull/347#issuecomment-3300913765


test/unit_test/utils_test/test_signature.py
mhs4670go, line 1: I also use `assertRaiseRegex` instead for checking specific messages as well because checking `ValueError` was not enough as there are many errors.

PR #352, opened 2025-09-17
[ptq] Remove redundant help messages

mhs4670go:
This commit removes redundant help messages.


tico/experimental/quantization/ptq/examples/compare_ppl.py
```diff
@@ -77,7 +77,7 @@ def main():
         "--dtype",
         choices=list(DTYPE_MAP.keys()),
         default="float32",
-        help="Model dtype for load (float32|bfloat16|float16).",
+        help=f"Model dtype for load.",
```
mhs4670go, line 80: The dtypes are showed with `choices` option.

```bash
  --dtype {float32,bfloat16,float16}
                        Model dtype for load.
```

PR #353, opened 2025-09-17
[ptq] Make quantize_with_gptq script configurable

mhs4670go:
This commit makes a script configurable.

This is necessary for integrating example scripts into the CI.

```
python tico/experimental/quantization/ptq/examples/quantize_with_gptq.py --model Maykeye/TinyLLama-v0
=== Config ===
Model            : Maykeye/TinyLLama-v0
Device           : cuda
DType            : float32
Stride           : 512
Calib preset     : debug (2,000 tokens)
Use HF cache?    : False

Loading FP model …
Applying GPTQ …
GPTQ calibration: 100%|███████████████████████████████████████████████████████████████████████████████████| 4358/4358 [00:04<00:00, 965.37it/s]
Quantizing layers: 100%|██████████████████████████████████████████████████████████████████████████████████████| 8/8 [01:43<00:00, 12.97s/layer]
Wrapping layers with PTQWrapper …                                                                                                              
Calibrating with 2,000 tokens.

Act-calibration: 100%|███████████████████████████████████████████████████████████████████████████████████████████| 1/1 [00:00<00:00, 11.71it/s]

Calculating perplexities …
Token indices sequence length is longer than the specified maximum sequence length for this model (324381 > 2048). Running this sequence through the model will result in indexing errors
PPL:   0%|                                                                                                             | 0/634 [00:00<?, ?it/s]
```

### DIFF
--- a/tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
+++ b/tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
@@ -24,6 +24,8 @@
 #   6. Freeze all Q-params and compute Wikitext-2 perplexity.
 # =============================================================================
 
+import argparse
+import sys
 from typing import Any
 
 import torch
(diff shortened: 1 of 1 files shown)


tico/experimental/quantization/ptq/examples/quantize_with_gptq.py
```diff
@@ -89,77 +96,192 @@ def inject_gptq_qparams(
...
+            token=args.hf_token,
+        )
+        .to(device)
+        .eval()
     )
-    new_layers.append(q_layer)
 
-q_m.model.layers = new_layers
+    # Disable KV cache to force full forward passes for introspection
+    model.config.use_cache = args.use_cache
```
seockho-kim, line 192: Why is this code disabling KV cache?

seockho-kim (approved):
LGTM

PR #355, opened 2025-09-17
[tests] Fix KVCache tests

periannath:
Shape of KVCache should be [batch, num_key_value_heads, seq_len, head_dim]


seockho-kim (approved):
LGTM

PR #356, opened 2025-09-18
[serialize] prefer fc over bmm on const lhs

glistening:
If lhs of matmul is const, emit fullyconnected instead of bmm.

NOTE: Please do not merge this PR. 
Hopefully, #341 will solve the problem better. (not only bmm to fullyconnected + transpose as my PR, but to 1 fullyconnected).


glistening:
I've done using [o2o](https://github.com/Samsung/ONE/pull/16233).

PR #360, opened 2025-09-19
[quantization] Split config classes

mhs4670go:
This commit splits config classes.

This change makes it easier to introduce new algorithms.


miusic (approved):
LGTM 👍

PR #363, opened 2025-09-22
[ptq] Introduce a wrapper for decoder export

mhs4670go:
This commit introduces a wrapper for decoder export.

This class is necessary for the decoder export that will be compiled with compilers of accelerators.


jinevening (approved):
LGTM

PR #371, opened 2025-09-24
[passes] Fix AtenOps view

periannath:
This commit fixes AtenOps view
- is_target_node assuems every nodes are instance of OpOverload
- torch.ops.aten.view is not a instance of OpOverload

### DIFF
--- a/tico/passes/ops.py
+++ b/tico/passes/ops.py
@@ -69,7 +69,6 @@ class AtenOps:
             torch.ops.aten.unsqueeze_copy.default,
         ]
         self.view = [
-            torch.ops.aten.view,
             torch.ops.aten.view.default,
             torch.ops.aten.view_copy.default,
         ]


tico/passes/ops.py
```diff
@@ -69,7 +69,6 @@ def __init__(self):
             torch.ops.aten.unsqueeze_copy.default,
         ]
         self.view = [
-            torch.ops.aten.view,
```
periannath, line 72: Without this PR, the following code fails with an AssertionError:

```python
if is_target_node(node, ops.aten.view):
```

Error log:
```bash
  File "/home/jhyeo/workspace/TICO/tico/passes/convert_expand_to_slice_cat.py", line 61, in call
    if not isinstance(node, torch.fx.Node) or not is_target_node(
  File "/home/jhyeo/workspace/TICO/tico/utils/utils.py", line 391, in is_target_node
    assert all(isinstance(t, torch._ops.OpOverload) for t in target_ops), target_ops
AssertionError: [<OpOverload(op='aten.reshape', overload='default')>, <OpOverloadPacket(op='aten.view')>, <OpOverload(op='aten.view', overload='default')>, <OpOverload(op='aten.view_copy', overload='defaul
t')>]
```

Issue #372, opened 2025-09-30
Prepare Project Documentation (Requirements, Design, Test Report)


mhs4670go:
## What

We need to prepare the following software development documents as part of the project process:

- Requirements Specification (SRS)

Define system scope, functional requirements, non-functional requirements, and constraints.

- Design Document

Describe the system architecture, module/component design, database schema, and API specifications.

- System Test Report

Summarize the test plan, environment, test case results, defect status, and overall evaluation.

### Notes

There are no strict format requirements, but each document should clearly convey its purpose. The documents will serve as a reference for development, QA, and stakeholder alignment.

mhs4670go:
I'll write down SRS and Design Documnet first.

jinevening:
STD has to be written based on features identified in the SRS. For instance, if SRS includes RF-1, then STD should have a test item for RF-1.

Could you write SRS first and share that? It would be possible to start writing STD after RFs are determined in SRS.

PR #373, opened 2025-09-30
[docs] Add SRS and SDD documents

mhs4670go:
This commit adds SRS and SDD documents.

Related: #372 

### DIFF
--- a/docs/design.md
+++ b/docs/design.md
@@ -1,0 +1,287 @@
+# System Design Document
+
+> Torch IR → Circle conversion & optimization library with pluggable passes and automated validation
+
+## Table of Contents
+
+- [0. Decision Log (Design Summary)](#0-decision-log-design-summary)
+- [1. Introduction](#1-introduction)
+  - [1.1 Purpose](#11-purpose)
+  - [1.2 Scope](#12-scope)
+  - [1.3 Non-Goals](#13-non-goals)
+- [2. Requirements Traceability](#2-requirements-traceability)
+- [3. Software Structure Design](#3-software-structure-design)
+  - [3.1 High-Level Architecture](#31-high-level-architecture)
+  - [3.2 Core Modules](#32-core-modules)
+  - [3.3 Configuration](#33-configuration)
+  - [3.4 Invariants](#34-invariants)
+- [4. Behavior Design](#4-behavior-design)
+  - [4.1 End-to-End Conversion Flow](#41-end-to-end-conversion-flow)
+  - [4.2 Pass Scheduling & Priority](#42-pass-scheduling--priority)
+  - [4.3 Dynamic Shapes Behavior](#43-dynamic-shapes-behavior)
+  - [4.4 Error Handling & Logging](#44-error-handling--logging)
+- [5. Data & Interface Design](#5-data--interface-design)
+  - [5.1 Public APIs](#51-public-apis)
+  - [5.2 Pass Interface](#52-pass-interface-conceptual)
+  - [5.3 Configuration Schema](#53-configuration-schema-excerpt)
+- [6. Performance Design](#6-performance-design)
+- [7. Reliability & Accuracy](#7-reliability--accuracy)
+- [8. Security & Compliance](#8-security--compliance)
+
+## 0. Decision Log (Design Summary)
+
+- **IR Strategy**: Use **PyTorch ExportedProgram (core ATen)** as the working IR. No separate custom mid-IR.
+- **Pass Architecture**: Separate **legalization** and **optimization**, but allow interleaving under a **Pass Manager** with scheduling & invariants.
+- **Quantization**: Post-Training Quantization (PTQ) is **detected and applied conditionally** (fold/propagate qparams, bias quantization, dtype mismatch fix).
+- **Validation**: Multi-stage validation (unsupported ops / training ops / invariant checks) + E2E inference parity check against PyTorch outputs.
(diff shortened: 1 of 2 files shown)


docs/requirements.md
```diff
@@ -0,0 +1,84 @@
...
+
+- **Developers**: Utilize the library to convert and optimize models for NPU execution.
+- **Researchers**: Apply advanced quantization and optimization techniques for experimental models.
+- **Product Teams**: Benefit from simplified workflows and faster time-to-market.
+
+### 1.4 Definitions and Acronyms
+
+- **PyTorch**: Open-source machine learning framework.
+- **NPU**: Neural Processing Unit.
+- **Circle**: Proprietary intermediate representation (IR) format compatible with NPU compilers.
```
jinevening, line 40:
```suggestion
- **Circle**: Open-source intermediate representation (IR) format compatible with NPU compilers.
```

docs/requirements.md
```diff
@@ -0,0 +1,84 @@
...
+| ID    | Requirement Name                 | Description |
+|-------|----------------------------------|-------------|
+| RF-1  | PyTorch Model Conversion         | Convert PyTorch models directly into Circle format, ensuring structural integrity and operator compliance. |
+| RF-2  | Optimization Algorithm Support   | Support advanced optimization techniques (e.g., kernel fusion, operator reordering), with options for user-defined configurations. |
+| RF-3  | Quantization Integration Support | Consume pre-quantized parameters, fold/remove redundant (de)quant ops, quantize missing biases where required, and insert bridging quantize ops on dtype mismatches to keep the graph legal for Circle/NPU. |
+| RF-4  | Verification & Debugging Tools   | Offer debugging mode, display detailed error messages, and validate correctness of outputs compared to original PyTorch models. |
+| RF-5  | NPU Compiler Compatibility       | Automatically test converted Circle files against NPU compilers and report compatibility results. |
+
+
+## 3. Non-Functional Requirements
```
jinevening, line 56: RNF-1 -> Which model?
RNF-2 -> What is the original PyTorch model size? How can it be measured?
RNF-3 -> What does fidelity mean? Maybe we should use PEIR and relax the number 99%.
RNF-4 -> "A wide range of" is ambiguous. It would be better to specify target models (ex: Llama, transformer, and possibly some CNNs?.
RNF-5 -> Better to change compatibility with update of torch version and circle schema. TICO converts torch model to circle. Strictly speaking, NPU compiler is not visible to TICO.

docs/design.md
```diff
@@ -0,0 +1,287 @@
...
+- Building an in-house mid-IR other than PyTorch `ExportedProgram` (EP).
+- Implementing a training-time quantization or compiler backend; TICO targets **export + legalization + validation**.
+
+
+## 2. Requirements Traceability
+
+| Design Area | Requirement IDs |
+|---|---|
+| Direct conversion to Circle | RF-1 |
+| Optimization pipeline | RF-2, RNF-1, RNF-2 |
```
seockho-kim, line 63: FR and NFR might be more accurate expressions.

mhs4670go: I just followed the internal template.

docs/requirements.md
mhs4670go: I've rewritten the contents the way it sounds more specific. PTAL:)

docs/requirements.md
```diff
@@ -0,0 +1,84 @@
...
+
+| ID     | Requirement Name               | Description |
+|--------|--------------------------------|-------------|
+| RNF-1  | Performance: Conversion Speed  | Conversion process should complete within approximately 10 seconds. |
```
jinevening, line 60:
```suggestion
| RNF-1  | Performance: Conversion Speed  | Conversion process should complete within approximately 10 seconds for Llama 3.2 1B. |
```

docs/requirements.md
```diff
@@ -0,0 +1,84 @@
...
+| RNF-2  | Performance: File Size         | Converted Circle files should be equal to or smaller than the serialized PyTorch model size saved via `torch.save(model.state_dict())`. |
+| RNF-3  | Reliability                    | Converted model outputs should remain close to the PyTorch reference outputs. The similarity is measured using **PEIR (Peak Error Interval Ratio)** defined as `max(\|a - b\|) / (max(a) - min(a))`. The PEIR must not exceed **3% (0.03)** for any output tensor. |
+| RNF-4  | Maintainability: Operator Support | The library should support operator coverage sufficient for modern Transformer and LLM architectures. |
+| RNF-5  | Maintainability: Schema & Framework Updates | Must remain compatible with updates in PyTorch (latest stable & LTS) and the Circle schema specification. |
```
jinevening, line 64:
```suggestion
| RNF-5  | Maintainability: Schema & Framework Updates | Must remain compatible with recent PyTorch versions (2.5 ~ latest stable) and the Circle schema specification. |
```

docs/requirements.md
mhs4670go: Isn't it too specific? The target models are described below RNF-4 and the conversion speed is applied to them.

seockho-kim: Conversion speed might be related with model size. 
How about just 1B instead of specific model?

jinevening: Llama model is just an example. As @seockho-kim suggested, 1B scale model (or 3B scale model) would be ok.

>  The target models are described below RNF-4

It would be better to write target models here too for self-containedness.

PR #387, opened 2025-10-24
[ptq] Add README.md

mhs4670go:
This commit adds README.md.


dayo09 (commented):
I wonder if this is the right place of this README. The same interface goes for other quantization algorithms in this project. Could you explain what the point of difference between `tico/experimental/quantization/algorithm` and this is? If interface are the same, how about extracting that part to the common directory root?

Issue #392, opened 2025-10-28
Vulnerability alerts from huggingface transformer


jinevening:
## What

Let's resolve vulnerability alerts from huggingface transformer.

<img width="655" height="367" alt="Image" src="https://github.com/user-attachments/assets/e02e2d0e-570a-493b-ba55-e39081e10133" />

<img width="1244" height="623" alt="Image" src="https://github.com/user-attachments/assets/f7ef061b-a3b6-4237-89a5-939d97631de6" />

jinevening:
There exist similar alerts from other models.

<img width="686" height="803" alt="Image" src="https://github.com/user-attachments/assets/de440775-ed39-478a-9a6a-42365f67616a" />

seockho-kim:
In case of other models like TinyLlama, it doesn't work with transformers=4.53.0

```
======================================================================
ERROR: test.modules.model.TinyLlama.model.TinyLlama (test.pt2_to_circle_test.test_model.test.modules.model.TinyLlama.model)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/seockho/repo/TICO/test/pt2_to_circle_test/builder.py", line 98, in wrapper
    self._run(
  File "/home/seockho/repo/TICO/test/pt2_to_circle_test/builder.py", line 162, in _run
    convert_nnmodule_to_pt2(
  File "/home/seockho/repo/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 57, in wrapper
    raise e
  File "/home/seockho/repo/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 54, in wrapper
    return function(*args, **kwargs)
  File "/home/seockho/repo/TICO/test/pt2_to_circle_test/test_pt2_to_circle.py", line 87, in convert_nnmodule_to_pt2
    torch.export.save(exported, pt2_model_path)
  File "/home/seockho/repo/TICO/.venv/lib/python3.10/site-packages/torch/export/__init__.py", line 439, in save
    artifact: SerializedArtifact = serialize(ep, opset_version)
  File "/home/seockho/repo/TICO/.venv/lib/python3.10/site-packages/torch/_export/serde/serialize.py", line 2517, in serialize
    serialized_program = ExportedProgramSerializer(opset_version).serialize(
  File "/home/seockho/repo/TICO/.venv/lib/python3.10/site-packages/torch/_export/serde/serialize.py", line 1502, in serialize
    serialize_torch_artifact(constants),
[...]

PR #393, opened 2025-10-29
[test] Fix vulnerability alerts from huggingface transformers

seockho-kim:
We've got vulnerability alerts from huggingface transformers, so we need to update transformers.(>=4.53.0)

for https://github.com/Samsung/TICO/issues/392


mhs4670go:
It would be good to show the log of each test results as CI doesn't do those tests.

mhs4670go (approved):
LGTM

PR #396, opened 2025-10-29
[test] Improve model test filtering with exact matching

seockho-kim:
Improve model test filtering with model matching using environment variable CCEX_TEST_MODEL.
It supports exact matching and wildcard pattern matching.

### DIFF
--- a/test/pt2_to_circle_test/test_model.py
+++ b/test/pt2_to_circle_test/test_model.py
@@ -37,23 +37,12 @@ def load_tests(loader, tests, pattern):
 
     test_model = os.environ.get("CCEX_TEST_MODEL")
 
-    if test_model:
-        matches = fnmatch.filter(
-            os.listdir(str(testdir) + "/modules/model"), test_model
-        )
-    elif hasattr(loader, "testNamePatterns") and loader.testNamePatterns:
-        # Fall back to pattern matching for backward compatibility
-        pattern_name = loader.testNamePatterns[0]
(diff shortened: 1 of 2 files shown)


jaemini-shin:
Is it possible to apply wildcard(`*`) pattern on test_model?

I also feel inconvenience about lack of exact model matching, but I think pattern based filtering is still needed. e.g.) checking for `Llama` families, ...
If wildcard pattern is applicable, we can easily select which filtering to use!

jaemini-shin (approved):
LGTM!

test/pt2_to_circle_test/test_model.py
```diff
@@ -31,15 +31,29 @@ def load_tests(loader, tests, pattern):
...
+        )
+    elif hasattr(loader, "testNamePatterns") and loader.testNamePatterns:
+        # Fall back to pattern matching for backward compatibility
+        pattern_name = loader.testNamePatterns[0]
+        matches = fnmatch.filter(
+            os.listdir(str(testdir) + "/modules/model"), pattern_name
+        )
+    else:
+        # No filtering specified - return empty suite
+        return suite
```
mhs4670go: Are these lines really run?
```suggestion
    assert test_model
    matches = fnmatch.filter(
        os.listdir(str(testdir) + "/modules/model"), test_model
    )
```

seockho-kim: Okay, I'll update it.

PR #398, opened 2025-10-30
[DRAFT] Introduce FPI_GPTQ

stamalakhov:
This PR introduces Fixed Point Iteration for GPTQ.

Using `test_fpi_quantization_on_models.py` we can get the following results:

|model| FPI_speedup (%)|time_FPI_GPTQ (s)|time_GPTQ (s)|acc1_original|acc5_original|acc1_FPI_GPTQ|acc5_FPI_GPTQ|acc1_GPTQ|acc5_GPTQ|
|-|-|-|-|-|-|-|-|-|-|
|mobilenet_v2| **58.85**|1.7400|4.2287|78.5643|93.4613|76.8246|92.9014|76.9646|92.8614|
|mobilenet_v3_small| **48.21**|1.6642|3.2133|75.0050|91.0218|68.5263|87.2426|68.5863 | 87.1426|
|mobilenet_v3_large| **61.87**| 1.9597|5.1391|79.7041|94.6411|76.1648|92.3615|76.0848|92.4415|
|squeezenet1_1| **63.12**| 0.8621|2.3374|67.4265| 85.9028|66.3467|85.0630|66.3267|85.1030|
|shufflenet_v2_x1_5| **51.69**| 1.7218|3.5639|80.3839| 94.5411|77.8844|93.5213|77.9444|93.5013|
|convnext_large| **79.80**| 8.2193|40.6873|88.2423 |  98.1004|88.1024 | 98.1604|88.1024 | 98.1804|
|resnext101_64x4d| **77.96**| 4.3780|19.8659|87.2825 |  97.8204|86.9226 | 97.7005|86.9626 | 97.6605|
|vit_l_32| **80.07**| 6.9739|34.9904|81.6037 |  95.1210|81.8036 | 95.4809|81.6437 | 95.3209|

_Speed-up is measured as `speed_up = 100 * (1.0 - time_of_FPI_GPTQ/time_oF_GPTQ)`. Quantized nodes are (torch.nn.Linear, torch.nn.Conv2d, torch.nn.Conv1d).
All nodes are quantized to uint4. Activations are left unchanged.
Accuracy is measured on 5000 images of `imagenet/val`. GPU - is `cuda`_

Issue: https://github.com/Samsung/TICO/issues/397

TICO-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #397, opened 2025-10-30
Fixed Point Iteration for GPTQ


stamalakhov:
## What

Let's accelerate GPTQ https://github.com/Samsung/TICO/blob/c3eeb74c3aa5138da54ec79c3a97e35fbca7070f/tico/quantization/algorithm/gptq/gptq.py#L110-L152 (GPTQ cycle) using Fixed Point Iteration.

## Why
It occures that we need fast implementation of GPTQ cycle for quantizing `depthwise` convolution, because of many GPTQ cycles, so it can be pretty slow. Or we may need `mse` that aims to quantize GPTQ improved matrix instead of original `W`, in this case we also need a way to provide fast implementation of GPTQ cycle.

## How

[FPI_GPTQ.pdf](https://github.com/user-attachments/files/23382437/FPI_GPTQ.pdf)

In short : 
GPTQ quantized matrix follow the following equation:
$\tilde{W} = W_{original} - (\tilde{W} - Q(\tilde{W})) D(H^{-1})\tilde{H}$,
where 
$\tilde{W}$  - weight matrix right before quantization
$H$  -Hessian
$Q$ - quantizer,
$D(H^{-1})$ - inverse diagonal of $H$ ($\frac{1}{diag(H^{-1})}$)
$\tilde{H}$ - torch.triu($H^{-1}$)

**so using Fixed Point Iteration we can solve for $\tilde{W}$ and then use $Q(\tilde{W})$ as a final answer**:
$\tilde{W}^0 = W_{original}$
$\tilde{W}^1 = W_{original} - (\tilde{W}^0 - Q(\tilde{W}^0))D(H^{-1})\tilde{H}$,
...
$\tilde{W}^{k+1} = W_{original} - (\tilde{W}^{k} - Q(\tilde{W}^{k}))D(H^{-1})\tilde{H}$

_return_ $Q(\tilde{W}^{K})$
It occures that this sequence is converging pretty fast and using cuda we can get significant speed-up.

stamalakhov:
Now main linear nodes (`nn.Linear`, `nn.Conv2D`, `nn.Conv1D`, `nn.ConvTranspose2D`) are supported in FPI_GPTQ and GPTQ. So we can close this one. Feel free to reopen this otherwise.

PR #408, opened 2025-11-07
[quantization] Enable Conv's quantization in FPIGPTQ/GPTQ

stamalakhov:
This PR enables quantization of Convolution nodes in FPIGPTQ/GPTQ with tests enabled.

Draft: #398 
Related: #397

<details>
<summary> ./ccex test --include-internal</summary>

image
image

</details>

_Note for reviewers: Please let me know if test Conv1d is also in need. Or i should split the PR in two (the first for GPTQ, the scond for FPI_GPTQ ).Or is should exclude Conv1d_

### DIFF
--- a/tico/quantization/algorithm/gptq/quantizer.py
+++ b/tico/quantization/algorithm/gptq/quantizer.py
@@ -193,7 +193,15 @@ class GPTQQuantizer(BaseQuantizer):
             )
         ):
             # 1) Identify quantizable submodules within the layer
-            full = find_layers(layer)
+            full = find_layers(
+                layer, layers=[torch.nn.Linear, torch.nn.Conv2d, torch.nn.Conv1d]
+            )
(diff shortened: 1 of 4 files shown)


tico/quantization/algorithm/gptq/quantizer.py
```diff
@@ -193,7 +193,15 @@ def convert(self, model):
             )
         ):
             # 1) Identify quantizable submodules within the layer
-            full = find_layers(layer)
+            full = find_layers(
+                layer, layers=[torch.nn.Linear, torch.nn.Conv2d, torch.nn.Conv1d]
```
dayo09: If it's not a big burden, it's usually recommended that you also include test for `torch.nn.Conv1d`, too.

stamalakhov: Sure. I'll post it in the next PR.

PR #418, opened 2025-11-22
[utils] Allow condition to be None in RecordingInput

glistening:
Set default value of condition to None instead of a lambda.
Use a default lambda (lambda args_dict: True) internally if condition is not provided.


dayo09 (approved):
LGTM

glistening:
@dayo09, @seockho-kim PTAL

dayo09 (approved):
LGTM

PR #426, opened 2025-12-01
[passes] ConvertSymSizeToCircleShape

dayo09:
Let's support 'sym_size.int' torch ir which represents dynamic shape.


mhs4670go (approved):
LGTM

Issue #433, opened 2026-01-13
Upgrade torch nightly version


dayo09:
### What?

Let's upgrade torch nightly version. (periodic task)

#### before
torch==2.10.0.dev251012+cpu  (no more distributed)

#### after
torch==2.11.0.dev260112+cpu

### TODO

@mhs4670go NOTE that some torch.ao.quantization features are moved to torch/ao. They extracted the project outside.
Take a look at https://github.com/pytorch/pytorch/commit/0e38e0291ccb84d0ff0314846014f65c7080e1de

PR #437, opened 2026-01-15
Bump up torch nightly version

mhs4670go:
This commit bumps up torch nightly version.

It includes some changes to resolve errors caused by bump-up.

- Remove PT2E Qunatization.
  - torch has introduced a new library "torchao" that includes pt2e related codes.
  - pt2e quantization was introduced to TICO for test purposes.
  - Which means it doesn't have users.
  - Keep tracking the new library is too burden to have pt2e features in TICO.
- Replace PT2E related codes with the codes using `wrapq`.
- Suppress future warnings caused by torch.
- minor changes that are needed according to the bump-ups.

Related: #433


dayo09 (approved):
LGTM

PR #454, opened 2026-02-02
[quantization] Use QuantRMSNorm

mhs4670go:
This commit uses QuantRMSNorm in the decoder layer.

### DIFF
--- a/tico/utils/register_custom_op.py
+++ b/tico/utils/register_custom_op.py
@@ -710,7 +710,7 @@ def CircleRMSNorm():
     def rms_norm(
         hidden_states: torch.Tensor,
         weight: torch.Tensor,
-        eps: float = 1e-05,
+        eps: float = 1e-06,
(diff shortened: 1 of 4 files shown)


tico/utils/register_custom_op.py
```diff
@@ -710,7 +710,7 @@ def CircleRMSNorm():
     def rms_norm(
         hidden_states: torch.Tensor,
         weight: torch.Tensor,
-        eps: float = 1e-05,
+        eps: float = 1e-06,
```
mhs4670go: This aligns with the one from `transformers`.

stamalakhov: @mhs4670go Sorry for my curiosity. But is it needed to declare epsilon as default parameter? In the draft its declaration as default leads to failure of export in case `eps` equals `default` value. (but it needs separate investigation, looks like a bug in pytorch, but i'm not sure)

PR #460, opened 2026-02-03
Unroll GQA and remove neg ops

mhs4670go:
This commit rewrites the wrapper for below.

- Remove 5-rank Broadcast ops by unrolling GQA
- Fuse neg ops into position embeddings.

The changes makes exported graph hardware-friendly.

### DIFF
--- a/tico/quantization/wrapq/wrappers/llama/quant_attn.py
+++ b/tico/quantization/wrapq/wrappers/llama/quant_attn.py
@@ -129,6 +129,8 @@ class QuantLlamaAttention(QuantModuleBase):
         self.obs_mask_add = mk("mask_add")
         self.obs_softmax = mk("softmax")
         self.obs_attn_out = mk("attn_out")
+        self.obs_attn_weights = mk("attn_weights")
+        self.obs_attn_out_h = mk("attn_out_h")
 
         # Static causal mask template
         assert hasattr(cfg, "max_position_embeddings")
@@ -141,7 +143,7 @@ class QuantLlamaAttention(QuantModuleBase):
         x1, x2 = torch.chunk(t, 2, dim=-1)
         x1 = self._fq(x1, o_x1)
         x2 = self._fq(x2, o_x2)
-        x2n = self._fq(-x2, o_neg)
+        x2n = x2
         return self._fq(torch.cat((x2n, x1), -1), o_cat)
 
     @staticmethod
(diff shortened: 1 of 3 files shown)


mhs4670go:
After [the patch](https://github.com/Samsung/TICO/issues/461#issuecomment-3839562854), the error in the main branch has been resolved. But, current branch still got an error.

```bash
circle2circle: /home/seongwoo/ONE/compiler/luci/lang/src/Nodes/CircleConst.cpp:47: typename loco::DataTypeImpl<DT>::Type& luci::CircleConst::at(uint32_t) [with loco::DataType DT = loco::DataType::FLOAT32; typename loco::DataTypeImpl<DT>::Type = float; uint32_t = unsigned int]: Assertion `dtype() == DT' failed.
```

The error happened in `ForwardTransposeOpPass` pass in `one-optimize`.

mhs4670go:
@stamalakhov Ah, seems that below is a problem.

> Turns out that ForwardTransposeOpPass doesn't support uint8 dtype properly.

Some passes hasn't considered quantized inputs. Maybe this is the case. I'll update the code soon.

tico/quantization/wrapq/wrappers/llama/quant_attn.py
```diff
@@ -239,15 +230,48 @@ def forward(
...
+
+            # out: (B, G, S, H)
+            out_i = self._fq(attn_i @ v_i, self.obs_attn_out)
+
+            attn_weights_parts.append(attn_i)
+            attn_out_parts.append(out_i)
+
+        # concat heads back: (B, n_h, S, K) / (B, n_h, S, H)
+        attn_weights = torch.cat(attn_weights_parts, dim=1)
+        attn_out_h = torch.cat(attn_out_parts, dim=1)
```
stamalakhov: @mhs4670go 
I believe `attn_weights` should be quantized also. Currently they are producing floats. Or is it intended?

mhs4670go:
I am observing a significant increase in PEIR after introducing the changes. Even though I ran it with float RMSNorm.

```bash
python tico/quantization/wrapq/examples/quantize_llama_decoder_layer.py
┌───────────── Quantization Error Summary ─────────────
│ Mean |diff|: 0.037503
│ PEIR       : 60.925078 %
└──────────────────────────────────────────────────────
    ┌────────────────────────────────────────────┐
20.6┤                                            │
    │                          •                 │
    │                                            │
15.7┤                                            │
    │                                            │
    │                                            │
    │                                            │
10.8┤                                            │
    │                             •              │
    │                                            │
 6.0┤                 ••                         │
    │                 •                          │
    │                •                           │
    │                                            │
 1.1┤              •                             │
    │            ••                              │
    │                                            │
-3.8┤                                            │
    │          •                                 │
    │    • • •                                   │
    │          •                                 │
-8.6┤                                            │
    └┬──────────┬──────────┬─────────┬──────────┬┘
   -8.6       -1.3        6.0      13.3      20.6 
```

tico/quantization/wrapq/wrappers/llama/quant_decoder_layer.py
```diff
@@ -147,6 +189,16 @@ def forward(
             L = hidden_states.size(1)
             attention_mask = self._slice_causal(L, hidden_states.device)
 
+        if position_embeddings is None:
```
stamalakhov: @mhs4670go I believe these ones should be used unconditionally, otherwise `LlamaModel` will send its own `position_embeddings` and they will be desynchronized with `_rot()`. IMHO

stamalakhov (approved):
LGTM! Thank you!

PR #463, opened 2026-02-03
[quantization] Increase the test threshold

mhs4670go:
This commit increases the test threshold according to the RMSNorm wrapper.

The quantization of RMSNorm causes some differences. `test_layerwise_diff` test failure will be resolved with this PR.

### DIFF
--- a/test/quantization/wrapq/utils/test_introspection.py
+++ b/test/quantization/wrapq/utils/test_introspection.py
@@ -179,7 +179,7 @@ class TestSmoothQuantPTQDiff(unittest.TestCase):
             self.assertLessEqual(
                 metric_dict["diff"],
                 3.0,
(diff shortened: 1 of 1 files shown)


test/quantization/wrapq/utils/test_introspection.py
```diff
@@ -178,7 +178,7 @@ def test_layerwise_diff(self):
         for name, metric_dict in stats.items():
             self.assertLessEqual(
                 metric_dict["diff"],
-                1.0,
+                3.0,
                 msg=f"{name}: diff={metric_dict['diff']:.3e} > 1.0",
```
shs-park: Shoud this value be changed also?

```suggestion
                msg=f"{name}: diff={metric_dict['diff']:.3e} > 3.0",
```

shs-park (commented):
I left a question, PTAL
=)

PR #464, opened 2026-02-04
[quantization] Introduce a script for vqa eval

mhs4670go:
This commit introduces a script for vqa eval.


tico/quantization/evaluation/script/mini_vqa_eval.py
```diff
@@ -0,0 +1,316 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
```
stamalakhov: nit? m.b. 2026?

PR #471, opened 2026-02-05
[quantization] Remove redundnat cache logic

mhs4670go:
This commit removes redundant cache logic.


stamalakhov (approved):
LGTM!

PR #472, opened 2026-02-05
[quantization] Fix decoder layer related test

mhs4670go:
This commit fixes tests where position embeddings logic has changed.


stamalakhov (approved):
LGTM! Thank you!

PR #474, opened 2026-02-06
[passes] Remove redundant SelectV2 from Conv3D Legalize

dayo09:
Let's remove redundancy in legalized Conv3D.
Here, SelectV2 node is utilized only when T-axis dilation goes off T_MAX limit. It creates zero paddings for that case. However, when dilation[0] (T-axis dilation) is 1, the case never happen.


seockho-kim (approved):
LGTM

PR #478, opened 2026-02-09
[quantization] Revisit examples directory

mhs4670go:
This commit revisits examples directory according to the target models.


seockho-kim (approved):
LGTM

PR #481, opened 2026-02-10
[quantization] Support max seq len option

mhs4670go:
This commit supports max-seq-len option.


glistening (approved):
LGTM

PR #490, opened 2026-02-13
[quantization] Unroll all heads

mhs4670go:
This draft unrolls all heads.


tico/quantization/wrapq/examples/llama/quantize_llama_decoder_layer.py
```diff
@@ -44,7 +44,7 @@
 MODEL_NAME = "Maykeye/TinyLLama-v0"
 MAX_S = 256
 
-model = AutoModelForCausalLM.from_pretrained(MODEL_NAME)
+model = AutoModelForCausalLM.from_pretrained(MODEL_NAME, dtype="float32")
```
mhs4670go: This change is necessary. Otherwise, the default dypte is bfloat, which is not supported by torch fake-quant api.

shs-park (approved):
LGTM!
=)

Issue #510, opened 2026-02-24
[test] Install transformers for wrapq feature coverage


mhs4670go:
## What

Our PTQ framework is designed to be independent of `transformers` and does not require it as a core dependency. However, some features are implemented to integrate with transformers when it is available (e.g., module wrappers, model export paths, or compatibility layers).

Currently, our test environment does not install `transformers`, which means these optional integration paths are not being exercised during CI. As a result, we risk regressions in functionality that depends on `transformers` without noticing.

## Proposed Solution

Install `transformers` as part of the test environment (CI) so that irelated tests can run.

dayo09:
**wrapq** technically have dependency to transformers version. For example, recently added qwen3-vl quantization feature assumes that transformers above that qwen3-vl introductory version of transformers. 
I believe it is time to specify that to prevent further possible difficulties in dependency control.

dayo09:
I suggest these

1. We should specify suggested transformers version for qwen3-vl/llama quantization
  - If it is too burdensome to specify transformers version range, we can just fix to the one recommended version
2. We should specify Ci-testing transformers version to README or extra-requires.

PR #512, opened 2026-02-24
[github] Install transformers for tests

mhs4670go:
This commit installs transformers for tests.

Related: #510

### DIFF
--- a/.github/workflows/check-pr.yaml
+++ b/.github/workflows/check-pr.yaml
@@ -126,6 +126,12 @@ jobs:
           TORCH_VERSION=${{ matrix.torch-version }}
           ./ccex configure test --torch_ver $TORCH_VERSION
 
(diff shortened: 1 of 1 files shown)


.github/workflows/check-pr.yaml
```diff
@@ -126,6 +126,13 @@ jobs:
           TORCH_VERSION=${{ matrix.torch-version }}
           ./ccex configure test --torch_ver $TORCH_VERSION
 
+      - name: "Install python dependencies for tests"
+        working-directory: "CODE"
+        run: |
+          pip install transformers==4.57.3
```
dayo09: Model test installs different versions of transformers. Won't it clash?

mhs4670go:
Let's re-post a PR after some discussion in #510.

PR #520, opened 2026-02-25
[test] Add and modify some validations to run_value_test

llFreetimell:
This commit adds and modifies following validations.
- Add shape checking
- Modify `torch.equal` to `torch.allclose`

TICO-DCO-1.0-Signed-off-by: Seok Namkoong <seok9311@naver.com>


dayo09 (approved):
LGTM

PR #523, opened 2026-02-26
[quantization] Load models with float32 dtype in examples

mhs4670go:
This commit loads models with float32 dtype in examples.


dayo09 (approved):
LGTM

PR #526, opened 2026-02-26
[quantization] Equalize calibration sets

stamalakhov:
This PR equalizes calibration sets for different `max_seq_len`.

Let's use #533 instead of this one.


stamalakhov:
Let's use https://github.com/Samsung/TICO/pull/533 instead of this one.

PR #528, opened 2026-02-27
[quantization] Implement llama wrappers for decoding

mhs4670go:
This commit implements llama wrappers for decoding.

### DIFF
--- a/tico/quantization/wrapq/wrappers/ptq_wrapper.py
+++ b/tico/quantization/wrapq/wrappers/ptq_wrapper.py
@@ -27,6 +27,25 @@ class PTQWrapper(QuantModuleBase):
 
     It is itself a QuantModuleBase so composite wrappers can treat
      it exactly like any other quant module.
+
+    Variant-Aware Wrapping
+    ----------------------
+    Wrapper selection depends on the execution "variant" specified in PTQConfig:
+
+        - "prefill": full-sequence execution
+        - "decode" : single-token decoding specialization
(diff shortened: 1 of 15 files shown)


tico/quantization/wrapq/wrappers/ptq_wrapper.py
```diff
@@ -37,7 +58,8 @@ def __init__(
         fp_name: Optional[str] = None,
     ):
         super().__init__(qcfg)
-        wrapped_cls = lookup(type(module))
+        variant = getattr(qcfg, "wrapper_variant", "prefill") if qcfg else "prefill"
```
dayo09:
```suggestion
        variant = getattr(qcfg, "wrapper_variant", "prefill") if qcfg else None
```

Quantized nn module may not be one of prefill/decode stage

mhs4670go: I desinged to use "prefill" as a default value. Do you think exact default value is needed?

```python
Resolution Order
    ----------------
    1. Exact variant match
    2. "prefill" fallback
    3. Any registered variant (last-resort compatibility)
```

dayo09: Using a technical term, it's a `leaky abstraction`. Let's choose to be 'opt-in' variance. Let modules to have 'None' or 'invariant' if they don't belong to either of 'prefill' or 'decode.

PR #541, opened 2026-03-09
[quantization] Add prefill suffix in examples

mhs4670go:
This commit adds _prefill suffix in exmaples.


dayo09 (approved):
LGTM

PR #544, opened 2026-03-10
[quantization] Clean up example script

mhs4670go:
This commit cleans up example scripts.


tico/quantization/wrapq/examples/llama/quantize_attn_decode.py
```diff
@@ -28,20 +28,21 @@
...
 
 MODEL_NAME = "Maykeye/TinyLLama-v0"
-MAX_S = 256
+MAX_SEQ = 256
 B = 1
 N_CALIB = 16  # number of random calibration batches
 DEVICE = "cpu"
 
 torch.set_grad_enabled(False)
+torch.manual_seed(123)
```
shs-park: Is this seed value intended?

mhs4670go: Yes. I just picked random value to get constant results.

PR #550, opened 2026-03-12
[quantization] Add missing observer calls

mhs4670go:
This commit applies missing observer calls.


stamalakhov (approved):
LGTM!

PR #552, opened 2026-03-16
[quantization] Remove redundant args

mhs4670go:
This commit removes redundant args when exporting.


shs-park (approved):
LGTM!
=)

PR #570, opened 2026-03-23
[quantization][draft] Prefill-decode logic

stamalakhov:
This PR implements prefill_decode logic for Llama compatible LLM's.

<details> <summary> --prefill_decode run on SmolLM </summary>

```
Namespace(model='HuggingFaceTB/SmolLM2-135M-Instruct', device='cuda', dtype='float32', seed=42, trust_remote_code=False, hf_token=None, no_tqdm=False, no_GPTQ=False, no_spinquant=True, no_PTQ=False, output_dir='.', save=['ptq_checkpoint'], cache_dir='/mnt/storage/transformers_cache', nsamples_for_qcalibration=128, linear_weight_bits=4, gptq_mse='mse', max_seq_len=2048, calibrate_seq_len=2048, embedding_weight_bits=8, lm_head_weight_bits=4, eval_tasks=None, sensitivity_path=None, prefill_decode=True)
=== Config ===
Model            : HuggingFaceTB/SmolLM2-135M-Instruct
Device           : cuda
DType            : float32

Loading FP model …
Warning: You are sending unauthenticated requests to the HF Hub. Please set a HF_TOKEN to enable higher rate limits and faster downloads.
Loading weights: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 272/272 [00:00<00:00, 1158.94it/s]
Skipping SpinQuant preprocessing …

Calculating perplexities …
Token indices sequence length is longer than the specified maximum sequence length for this model (304986 > 8192). Running this sequence through the model will result in indexing errors
PPL:  99%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▋ | 148/149 [00:35<00:00,  4.13it/s]

┌── Wikitext-2 test perplexity ─────────────
│ FP32 :    17.38
└───────────────────────────────────────────
100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 128/128 [00:28<00:00,  4.48it/s]

┌── Wikitext-2 prefill_prefill original test perplexity ─────────────
│ FP32 :    16.96
└───────────────────────────────────────────
Applying GPTQ …
Quantizing layers: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 30/30 [02:46<00:00,  5.55s/layer]
[...]

PR #606, opened 2026-04-03
[quantization] Remove `copy`

stamalakhov:
This commit removes redundant copy of key modules.
Copy was used for debugging, so we don't need it any more.

Draft: #570

### DIFF
--- a/tico/quantization/wrapq/wrappers/llama/quant_attn_decode.py
+++ b/tico/quantization/wrapq/wrappers/llama/quant_attn_decode.py
@@ -12,7 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import copy
 from typing import List, Optional, Tuple
 
 import torch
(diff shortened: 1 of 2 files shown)


tico/quantization/wrapq/wrappers/llama/quant_attn_decode.py
```diff
@@ -103,7 +102,7 @@ def __init__(
             fp_attn.q_proj, qcfg=q_cfg, fp_name=f"{fp_name}.q_proj"
         )
         self.k_proj = PTQWrapper(
-            copy.deepcopy(fp_attn.k_proj), qcfg=k_cfg, fp_name=f"{fp_name}.k_proj"
+            fp_attn.k_proj, qcfg=k_cfg, fp_name=f"{fp_name}.k_proj"
```
mhs4670go: Oh, I actually think that it's needed because this modifies the original model's weight.

mhs4670go: > i mean the model is loaded, then its weights are modified in GPTQ.

Right. I just think that wrappres just wrap the original nodes unlike GPTQ. Anyway, currnet PR seems okay for reducing overhead.

PR #607, opened 2026-04-03
[quantization] Support SpinQuant

mhs4670go:
This commit supports SpinQuant.

```bash
│ FP32 :    11.05
│ int16 :    11.92
```


tico/quantization/algorithm/spinquant/README.md
```diff
@@ -0,0 +1,163 @@
...
+- "hadamard": generate randomized Hadamard-structured orthogonal matrices
+- "external": use user-provided matrices
+
+- r1
+
+Global hidden-dimension rotation matrix. Required when init_method="external".
+
+- r2_map
+
+Optional mapping from module keys to per-layer head-dimension rotation matrices.
```
zetwhite: Image

Markdown looks a bit messy(?). `external` and `r1` should be in different level, but it looks not.

stamalakhov (dismissed):
LGTM! Thank you!

Torrero (approved):
LGTM

tico/quantization/algorithm/spinquant/rotation_utils.py
```diff
@@ -0,0 +1,506 @@
...
+def left_multiply_linear_weight_(module: nn.Linear, rotation_t: torch.Tensor) -> None:
+    """
+    Left-multiply a linear weight by a transpose rotation matrix.
+
+    This updates:
+        W <- R^T @ W
+
+    Parameters:
+        module: Target linear module with weight shape [out_features, in_features].
+        rotation_t: Transposed square rotation matrix of shape [out_features, out_features].
```
zetwhite: (Very Optional) How about adding these comment for future reader?

```suggestion
        rotation_t: Transposed square rotation matrix of shape [out_features, out_features].
        
    Notes:
	    The caller passes r1.T (i.e., rotation_t = R^T).
	    This stores W_new = R^T @ W, so:
	        y = x @ W_new.T = x @ W.T @ R
	    Equivalent to rotating the output by R after the original weight.
```

(Background) Same reason with https://github.com/Samsung/TICO/pull/607/changes#r3037530480.

tico/quantization/algorithm/spinquant/README.md
```diff
@@ -0,0 +1,160 @@
...
+- `model.embed_tokens`
+- `lm_head`
+
+Instead, it adds two dedicated runtime rotation layers to the custom `SpinLlama` model:
+
+- `model.model.rotate_embedding`
+- `model.rotate_lm_head`
+
+This preserves tied embedding behavior while still applying the same logical transforms
+during inference.
```
zetwhite: Just for curiosity, do you know how hadamard matrix is lowered in our internal NPU?

Because In paper, it said that hadamard matrix multiplication can be implemented with some effecient algorithm so it has marginal overhead. 

I was just wondering if we can implement it in a similar way.

(from the spin quant paper)
> Hadamard rotation can be computed with fast hadamard transform and introduce marginal overhead to the inference latency.

> This is because online Hadamard transforms can be efficiently implemented without significant overhead.

zetwhite (dismissed):
LGTM 👍 

I read the code except `hardamard_uitls.py` and all looks good 😄 
I added some comments to improve documents for future readers, but it totally optional!

tico/quantization/algorithm/spinquant/README.md
mhs4670go: Can't write down the details:) But, there's no difference between the hadamard matrix and other nn.Linear layers. `rotate_embedding` and `rotate_lm_head` are just linear layers.

FYI, the hadamard matrix computation (x @ H) has O(n^2) complexity. But, FHT (fast hadamard transform) has the same thing with only O(nlogn).

zetwhite (approved):

stamalakhov (approved):

Issue #612, opened 2026-04-06
Refactor PTQ Wrapper Design for Unified Calibration and Export Separation


mhs4670go:
## Background

Currently, we maintain separate quantization wrappers for **prefill** and **decode** paths (e.g., `QuantLlamaDecoderLayerPrefill`, etc.).

However, this introduces several issues:

- **Code duplication** across wrappers (observer registration, module wiring, quant logic)
- Risk of **inconsistent observer placement** between prefill and decode
- Increased **maintenance and testing overhead**
- Tight coupling between:
  - **calibration/evaluation logic (HF-compatible API)**
  - and **export requirements (accelerator-specific API)**

At the same time, export requires a **different input/output contract**, such as:
- Removing control-flow inputs like `use_cache`
- Fixing return types (e.g., always returning KV cache or not)
- Providing accelerator-friendly static inputs

## Problem

We are currently mixing two different concerns into the same abstraction:

1. **Runtime / Calibration Interface**
   - HF-compatible `forward`
   - Used for:
     - `prepare`
     - calibration
     - evaluation
     - regression checks

2. **Export Interface**
   - Accelerator-specific static graph
   - Requires:
     - fixed input/output signature
     - no control-flow flags (`use_cache`, `output_attentions`, etc.)

This leads to:
- Unnecessary complexity in `forward`
- Export graph pollution (unused inputs like `use_cache`)
- Difficulty maintaining consistency between prefill/decode variants

### 1. Unify Quantization Wrappers (Single Wrapper)

Use a **single quantization wrapper** for both prefill and decode:

- Example:
  - `QuantLlamaDecoderLayer`
  - `QuantLlamaAttention`

Key ideas:
- Keep **one unified `forward`**, similar to the original HF model
- Use:
  - `past_key_value is None` → prefill behavior
  - `past_key_value is not None` → decode behavior
- Maintain full HF compatibility for:
  - calibration
  - evaluation
  - integration

### 2. Separate Export via Thin Adapters

Introduce **export-specific adapter modules** instead of modifying the main wrapper.

Examples:

- `DecoderLayerPrefillExportAdapter`
- `DecoderLayerDecodeExportAdapter`

Responsibilities of adapters:
- Fix input/output contract
- Remove control-flow arguments:
  - `use_cache`
  - `output_attentions`
- Ensure deterministic return structure
- Provide accelerator-friendly signatures

Example:

```python
class DecoderLayerPrefillExportAdapter(nn.Module):
[...]

PR #618, opened 2026-04-10
[quantization] Remove variant features

mhs4670go:
This commit removes variant features.

Related: #612


tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
```diff
@@ -125,20 +125,12 @@ def save_layers_to(q_m, max_seq_len, save_layers_to_folder):
...
-        attention_mask = qlayer.wrapped._slice_causal(S, "cpu").squeeze(0)
-        dtype = example_hidden.dtype
-        pos_embeds = qlayer.wrapped._slice_rope(S, "cpu", dtype)
-
         print(f"Saving model layer_{i} to {save_path.resolve()}")
         with torch.no_grad():
             with SuppressWarning(UserWarning, ".*"):
                 cm = tico.convert(
-                    qlayer,
+                    qlayer.wrapped.as_export_module("prefill").eval(),
```
stamalakhov, line 140: Ahh. @mhs4670go 
But what about `attention_mask` and `pos_embeds`?

mhs4670go: They are internally generated. Isn't it okay not to give them?

stamalakhov: But they will be statically generated for every layer, and 2048 `seq_len` it will require much disk space. So i'm not sure. Externally set parameters were done on purpose to save disk space. But anyway it can be done later.

mhs4670go: Ah, I forgot that. Let's revert this change.

stamalakhov (approved):
LGTM! Thank you!

PR #638, opened 2026-04-16
[quantization] Fix fp_name mismatch with GPTQ

mhs4670go:
This commit fixes the mismatch and add verbose injection diagnostics.


stamalakhov (approved):
LGTM! Thank you!

PR #644, opened 2026-04-21
[passes] Introduce EliminateRankRoundTripRegion

mhs4670go:
### eliminate_rank_round_trip (experimental)

Eliminates reshape round-trip patterns (4D → 3D → 4D) by lifting the intermediate
3D computation region back to 4D.

This pass:
- Removes the source rank-reducing reshape
- Rewrites supported intermediate ops (e.g., elementwise, permute, transpose, split/chunk)
  to operate directly on the original 4D tensor
- Updates sink reshapes to consume the rewritten 4D values

Note:
This is a semantic region rewrite rather than a simple peephole optimization.
It is currently disabled by default until broader correctness coverage is verified.

### Before

image

### After

image

### DIFF
--- a/tico/passes/eliminate_rank_round_trip_region.py
+++ b/tico/passes/eliminate_rank_round_trip_region.py
@@ -47,21 +47,21 @@ class RankMapping:
     axis_3d_to_4d: dict[int, int]
     inserted_axis_4d: int
 
-    def remap_axis(self, axis: int, rank_3d: int = 3) -> int:
+    def remap_axis(self, axis: int) -> int:
         """
         Remap one 3D axis into the corresponding 4D axis.
         """
         if axis < 0:
-            axis += rank_3d
+            axis += 3
         if axis not in self.axis_3d_to_4d:
             raise ValueError(f"Cannot remap 3D axis {axis}")
         return self.axis_3d_to_4d[axis]
 
-    def remap_axes(self, axes: list[int], rank_3d: int = 3) -> list[int]:
+    def remap_axes(self, axes: list[int]) -> list[int]:
         """
         Remap a list of 3D axes into 4D axes.
         """
-        return [self.remap_axis(axis, rank_3d=rank_3d) for axis in axes]
+        return [self.remap_axis(axis) for axis in axes]
 
     def remap_permute(self, perm_3d: list[int]) -> list[int]:
         """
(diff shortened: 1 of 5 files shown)


tico/passes/eliminate_rank_round_trip_region.py
```diff
@@ -0,0 +1,855 @@
...
+
+    Then:
+        axis_3d_to_4d = {0: 0, 1: 2, 2: 3}
+        inserted_axis_4d = 1
+    """
+
+    axis_3d_to_4d: dict[int, int]
+    inserted_axis_4d: int
+
+    def remap_axis(self, axis: int, rank_3d: int = 3) -> int:
```
llFreetimell: Do we need `rank_3d` parameter?
According to comment, we are already assuming 3D tensor..!

zetwhite (dismissed):
LGTM

glistening (dismissed):
LGTM

tico/passes/eliminate_rank_round_trip_region.py
mhs4670go: You're right. It feels a bit over-engineered. Let's remove this in the next PR.

llFreetimell (dismissed):
LGTM

tico/passes/eliminate_rank_round_trip_region.py
```diff
@@ -0,0 +1,855 @@
...
+                    if user not in sink_reshapes:
+                        sink_reshapes.append(user)
+                    continue
+
+                if not self._is_supported_internal_node(user):
+                    continue
+
+                if user not in region_nodes:
+                    region_nodes.add(user)
+                    internal_nodes.append(user)
```
llFreetimell: Not critical, but just for removing redundancy :)
If we use following code, `region_nodes` is no more needed and can be removed from codes.

```suggestion
                if user not in internal_nodes:
                    internal_nodes.append(user)
```

tico/passes/eliminate_rank_round_trip_region.py
```diff
@@ -0,0 +1,855 @@
...
+        region_nodes: set[torch.fx.Node],
+        source_reshape: torch.fx.Node,
+    ) -> bool:
+        """
+        Check whether an external input is allowed for a region internal node.
+
+        Allowed external inputs:
+          - non-node constants
+          - the source reshape itself
+          - nodes already inside the region
```
llFreetimell: This function is for checking "external" input, but this comment is saying that we need to check "internal" nodes too... little bit confused :(

tico/passes/eliminate_rank_round_trip_region.py
mhs4670go: Very good suggestion. 

That being said, they have their own pruposes.

`internal_nodes` is a list, so membership checks are O(N), which can become inefficient as the region grows. They clearly separate concerns:
  - `regison_nodes` tracks whether a node has already been seen (O(1) lookup)
  - `internal_nodes` preserves insertion order for later processing

But, I think renaming `region_nodes` to `visited` would be more helpful to let readers know its purpose.

tico/passes/eliminate_rank_round_trip_region.py
mhs4670go: You're right. It's a bit confusing. I'll re-phrase the comments.

zetwhite (approved):

llFreetimell (approved):
LGTM, but just some comments for later :)

- Maybe in the future, renaming the pass name would be better since only 4D->3D->4D case is implemented
- Current tests are only validating the number of operators, but checking each shape would be better

PR #651, opened 2026-04-27
[quantization] Disable `use_cache` in GPTQ

stamalakhov:
This PR disables `use_cache` in GPTQ due to large vram overhead which is not needed.

Background: could not run quantization on A100 for Llama3B. :cry: 


mhs4670go (approved):
LGTM

PR #652, opened 2026-04-27
[quantization] Add a test for overrides propagation

mhs4670go:
This commit adds a test for checking if overrides propagation works well.


stamalakhov (approved):
LGTM!

PR #660, opened 2026-04-27
[quanization] Save prefill-decode layers

stamalakhov:
This PR exports llama decoder layers to circle format for prefill-decode pipeline.

<details> <summary> Sample output for `python tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py --model Maykeye/TinyLLama-v0 --output_dir "/home/stanislav" --max_seq_len 2048 --linear_weight_bits "4" --gptq_mse mse "--cache_dir" "/mnt/storage/transformers_cache" --decode_calibration_steps 1 --save circle_per_layer` </summary>

```
...
Saving decoder_layer_prefill_0 to /home/stanislav/decoder_layer_prefill_0.q.circle
Saving decoder_layer_decode_0 to /home/stanislav/decoder_layer_decode_0.q.circle
Saving decoder_layer_prefill_1 to /home/stanislav/decoder_layer_prefill_1.q.circle
Saving decoder_layer_decode_1 to /home/stanislav/decoder_layer_decode_1.q.circle
Saving decoder_layer_prefill_2 to /home/stanislav/decoder_layer_prefill_2.q.circle
Saving decoder_layer_decode_2 to /home/stanislav/decoder_layer_decode_2.q.circle
Saving decoder_layer_prefill_3 to /home/stanislav/decoder_layer_prefill_3.q.circle
Saving decoder_layer_decode_3 to /home/stanislav/decoder_layer_decode_3.q.circle
Saving decoder_layer_prefill_4 to /home/stanislav/decoder_layer_prefill_4.q.circle
Saving decoder_layer_decode_4 to /home/stanislav/decoder_layer_decode_4.q.circle
Saving decoder_layer_prefill_5 to /home/stanislav/decoder_layer_prefill_5.q.circle
Saving decoder_layer_decode_5 to /home/stanislav/decoder_layer_decode_5.q.circle
Saving decoder_layer_prefill_6 to /home/stanislav/decoder_layer_prefill_6.q.circle
Saving decoder_layer_decode_6 to /home/stanislav/decoder_layer_decode_6.q.circle
Saving decoder_layer_prefill_7 to /home/stanislav/decoder_layer_prefill_7.q.circle
Saving decoder_layer_decode_7 to /home/stanislav/decoder_layer_decode_7.q.circle
```

[log_myakeye_save_pref_decod.txt](https://github.com/user-attachments/files/27133109/log_myakeye_save_pref_decod.txt)

</details>

### DIFF
--- a/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
+++ b/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
@@ -195,8 +195,7 @@ def save_layers_to(
         suffix = "prefill_" if prefill_decode else ""
         layer_name = f"decoder_layer_{suffix}{i}"
         save_path = pathlib.Path(save_layers_to_folder, f"{layer_name}.q.circle")
-        B, D = 1, config.hidden_size
-        S = max_seq_len if not prefill_decode else max_seq_len - 1
+        B, S, D = 1, max_seq_len, config.hidden_size
         example_hidden = torch.randn(B, S, D)
 
         attention_mask = (
--- a/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
+++ b/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
@@ -144,7 +144,41 @@ def save_model_to(q_m, calib_inputs, save_circle_to_folder):
             cm.save(save_path)
 
 
-def save_layers_to(q_m, max_seq_len, save_layers_to_folder):
+# -----------------------------------------------------------------------------
+# copied from quantize_decoder_layer_decode.py
+# -----------------------------------------------------------------------------
+def make_random_decode_batch(model, B, DEVICE, MAX_SEQ):
+    # TODO reduce code duplication
+    D = model.config.hidden_size
+    head_dim = getattr(model.config, "head_dim", D // model.config.num_attention_heads)
+    n_kv = model.config.num_key_value_heads
+
+    # Single-token hidden state.
+    x = torch.randn(B, 1, D, device=DEVICE)
+
+    # RoPE tables for the *current token* only.
+    cos = torch.randn(B, 1, head_dim, device=DEVICE)
+    sin = torch.randn(B, 1, head_dim, device=DEVICE)
+    pos = (cos, sin)
+
+    # Additive mask of final static width: (B, 1, MAX_SEQ)
+    # Simulate that only the first L_eff positions are valid and the rest are padding.
+    L_eff = torch.randint(low=1, high=MAX_SEQ + 1, size=(1,)).item()
+    mask = torch.zeros(B, 1, MAX_SEQ, device=DEVICE, dtype=torch.float32)
+    if L_eff < MAX_SEQ:
+        mask[:, :, L_eff:] = float("-120")
+
+    # Static-sized past KV (already RoPE-applied for past tokens).
+    past_k = torch.randn(B, n_kv, MAX_SEQ - 1, head_dim, device=DEVICE)
+    past_v = torch.randn(B, n_kv, MAX_SEQ - 1, head_dim, device=DEVICE)
+    past = (past_k, past_v)
+
+    return x, pos, mask, past
+
+
+def save_layers_to(
+    q_m, max_seq_len, save_layers_to_folder, prefill_decode: bool = False
+):
     """
     Export and save quantized decoder layers one by one in circle format.
     """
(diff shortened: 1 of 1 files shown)


tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
```diff
@@ -287,7 +144,41 @@ def save_model_to(q_m, calib_inputs, save_circle_to_folder):
             cm.save(save_path)
 
 
-def save_layers_to(q_m, max_seq_len, save_layers_to_folder):
+# -----------------------------------------------------------------------------
+# copied from quantize_decoder_layer_decode.py
+# -----------------------------------------------------------------------------
+def make_random_decode_batch(model, B, DEVICE, MAX_SEQ):
```
stamalakhov, line 150: Or this one can be moved to some `utils.py`

tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
```diff
@@ -301,8 +192,11 @@ def save_layers_to(q_m, max_seq_len, save_layers_to_folder):
     layers = q_m.wrapped.model.wrapped.layers
     config = q_m.wrapped.config
     for i, qlayer in enumerate(layers):
-        save_path = pathlib.Path(save_layers_to_folder, f"decoder_layer_{i}.q.circle")
-        B, S, D = 1, max_seq_len, config.hidden_size
+        suffix = "prefill_" if prefill_decode else ""
+        layer_name = f"decoder_layer_{suffix}{i}"
+        save_path = pathlib.Path(save_layers_to_folder, f"{layer_name}.q.circle")
+        B, D = 1, config.hidden_size
+        S = max_seq_len if not prefill_decode else max_seq_len - 1
```
mhs4670go: The PR exports prefill with `S = max_seq_len - 1`, while the existing prefill export uses `MAX_SEQ` as the static sequence length.

stamalakhov: As far as i understand, in prefill-decode pipeline prefill produces kv-pairs and next token for `max_seq_len - 1`input sequence, this token will be input at `max_seq_len` at decode with prefill kv-pairs, which will produce the next (`max_seq_len`) token. Am i missing something? This approach was tested in https://github.com/Samsung/TICO/pull/570#issue-4120411406.

stamalakhov: I mean `prefill` layer produces next token for some `seq_len` then decode produces the next token at `seq_len + 1`, with `prefill` token already inserted into sequence. Please correct me if i'm wrong.

stamalakhov: Or prefill_layers should be saved at `max_seq_len`, so decode will be saved at `max_seq_len +1`?

mhs4670go: Prefill layer has fixed shape with `max_seq_len` because it's easier to accelerate on device. Decode layer receives kv cache whose sequence is (max_seq_len -1). 

> decode will be saved at max_seq_len +1?

The generated token just will be saved at current_token_len + 1. Which means decoding is only possible unitl max_seq_len.

stamalakhov: Ahh. So then kv-cache trimming must be managed by runtime, i suppose.

mhs4670go: Yes. The cache management will be done at runtime. I guess `tico/quantization/wrapq/examples/static_llama_layer_runtime.py` helps understaning the pipeline.

stamalakhov: > Yes. The cache management will be done at runtime. I guess `tico/quantization/wrapq/examples/static_llama_layer_runtime.py` helps understaning the pipeline.

But `static_llama_layer_runtime` does not use static shapes.

stamalakhov: @mhs4670go 
Fixed. Please take another look.

mhs4670go (approved):
LGTM

tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
mhs4670go: > But static_llama_layer_runtime does not use static shapes.

Ah, you're right. I'll make it more realistic.

PR #662, opened 2026-04-28
[quantization] Fix missing quant point

mhs4670go:
This commit fixes past kv not having qparam.


stamalakhov (approved):
LGTM!

PR #666, opened 2026-04-28
[quantization] Modularize main pipeline in llama example script

mhs4670go:
This commit modularizes main pipeline by extracting execution flow.


stamalakhov (approved):
LGTM!

PR #671, opened 2026-04-30
[quantization] Revise kv cache helper

mhs4670go:
This commit revises kv cache helper to remove duplicate codes.


stamalakhov (approved):
LGTM! Thank you!

PR #677, opened 2026-05-05
[quantization] Clear GPTQ quantizers

stamalakhov:
This PR clears GPTQ quantizers to save memory.

Related: #678

### DIFF
--- a/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
+++ b/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
@@ -320,6 +320,22 @@ def inject_gptq_qparams(
         _print_sample("unused GPTQ entries", unused)
 
 
+# -------------------------------------------------------------------------
+# Helper — clear gptq quantizers after injection
+# -------------------------------------------------------------------------
+def clear_gptq_quantizers(model: torch.nn.Module) -> None:
+    """Remove GPTQ quantizer attributes from the model to free memory.
+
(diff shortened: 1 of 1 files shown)


mhs4670go (dismissed):
LGTM. But, maybe instead of only compressing the internal `Quantizer` state, we could provide an explicit API/helper to clear GPTQ handoff artifacts after GPTQ → PTQ injection.

`model.quantizers` is only needed temporarily for injecting GPTQ qparams into PTQ observers. After `inject_gptq_qparams()` finishes, the final model no longer needs the GPTQ quantizer objects, including `scale`, `zero`, and `sensitivity`.

Something like this may make the lifecycle clearer:

```python
def clear_gptq_quantizers(model: torch.nn.Module) -> None:
    if hasattr(model, "quantizers"):
        delattr(model, "quantizers")
    if hasattr(model, "wrapped") and hasattr(model.wrapped, "quantizers"):
        delattr(model.wrapped, "quantizers")
```

Then the pipeline can call it immediately after qparam injection. This would reduce the final checkpoint size more directly than `compress_state()`, while keeping the shared convert() API unchanged.

Please feel free to merge this PR or apply above suggestion.

stamalakhov:
> Something like this may make the lifecycle clearer:
...
Then the pipeline can call it immediately after qparam injection. This would reduce the final checkpoint size more directly than compress_state(), while keeping the shared convert() API unchanged.

@mhs4670go 
Thank you for your review.
Ok. I'll change PR according to your suggestion.

mhs4670go (approved):
LGTM. Thank you!

PR #684, opened 2026-05-07
[quantization] SmoothQuant support for text components of Qwen3-vl

Torrero:
This PR supports SmoothQuant for text components of Qwen3-vl

Additionally `exclude_appliers` was introduced into `apply_smoothing` function for control to which part of model SmoothQuant will be apllied. 

SmoothQuant application for both parts (vision and text) of Qwen3-vl model leads to additional accuracy degradation. Tested with  `vqav2` dataset, 1000 samples.

|model|vqav2| accuracy degradation|
|--|--|--|
|original|0.895| - |
|quantized W4A16 + GPTQ + SMSE |0.695| 20%|
|quantized W4A16 + GPTQ + SMSE + smoothquant for vision part (alpha=0.8)|0.721| 17% |
|quantized W4A16 + GPTQ + SMSE + smoothquant for both parts text + vision (alpha=0.8)|0.405| 49% |
|quantized W4A16 + GPTQ + SMSE + smoothquant for text part (alpha=0.8)|0.695| 20%|

### DIFF
--- a/tico/quantization/algorithm/smoothquant/smooth_quant.py
+++ b/tico/quantization/algorithm/smoothquant/smooth_quant.py
@@ -72,9 +72,22 @@ def smooth_weights(
     elif isinstance(front_module, torch.nn.LayerNorm):
         front_numel = front_module.weight.numel()
     else:
-        raise NotImplementedError(
-            f"Unsupported module type: {type(front_module).__name__}"
-        )
+        # Try Qwen3VLTextRMSNorm
+        try:
+            from transformers.models.qwen3_vl.modeling_qwen3_vl import (
+                Qwen3VLTextRMSNorm,
+            )
+
+            if isinstance(front_module, Qwen3VLTextRMSNorm):
+                front_numel = front_module.weight.numel()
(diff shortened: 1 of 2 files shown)


mhs4670go:
Instead of separate negative flags such as `--smoothquant_exclude_vision` and `--smoothquant_exclude_text`, how about making the target components explicit?

```bash
--smoothquant_components vision|text|both
```

For example:

```python
parser.add_argument(
    "--smoothquant_components",
    choices=["vision", "text", "both"],
    default=None,
    help="Target components for SmoothQuant.",
)

# ..

if args.smoothquant:
    if args.smoothquant_components is None:
        raise ValueError(
            "--smoothquant_components must be specified when "
            "--smoothquant is enabled."
        )
```

---

Just curiosity, have you tried only appyling text part?

Torrero:
> Just curiosity, have you tried only applying text part?

yes, it doesn't produces improvement accuracy. I updated the table above.

mhs4670go (approved):
LGTM

PR #688, opened 2026-05-08
[quantization] PPL evaluation on the whole dataset

Torrero:
Let's conduct PPL evaluation on the whole dataset.

This PR introduces the `--ppl_split` parameters to split the dataset into samples and Eliminates the limit on the number of samples to evaluate the PPL by setting the entire dataset to be evaluated.


dvsav (approved):
LGTM

mhs4670go (approved):
LGTM

Issue #690, opened 2026-05-11
[quantization] Make LlamaAttention export configurable


mhs4670go:
## Motivation

The current `QuantLlamaAttention` implementation is heavily specialized for the NPU export graph. It applies several graph-level transformations directly in the wrapper, including:

- fusing the attention scale into `k_proj`
- assuming the RoPE `rotate_half` negation has been moved into the precomputed sine table
- unrolling attention heads to avoid transpose-heavy graph patterns

These choices are useful for NPU export, but they make the wrapper harder to use for other workflows. In particular:

- rollback or A/B testing requires code changes instead of configuration changes
- fast GPU evaluation is slower than necessary because the attention path is not close to the original Hugging Face implementation
- regression debugging is harder because the wrapper no longer has a reference-like execution path

Related: #551, https://github.com/Samsung/TICO/pull/687#discussion_r3207458583

## Proposal

Introduce a configurable execution profile for Llama PTQ wrappers.

The profile is stored in `PTQConfig.model_args` and can be set through `build_llm_ptq_config()`:

```python
qcfg = build_llm_ptq_config(
    model_type="llama",
    num_hidden_layers=len(model.model.layers),
    ...,
    profile="reference_eval",
)
```

Supported profiles:

```
reference_eval
  - scale_fusion = "none"
  - rope         = "hf"
  - layout       = "batched"

npu_export
  - scale_fusion = "k_proj"
  - rope         = "pre_negated_sin"
  - layout       = "unrolled"
```

The default profile remains `npu_export` to preserve existing behavior.

Advanced users can still override attention-specific behavior before `prepare()` by mutating `qcfg.model_args`:

```python
qcfg.model_args = {
    **dict(qcfg.model_args),
    "attention": {
        "layout": "unrolled",
        "scale_fusion": "none",
        "rope": "hf",
    },
}
```

### Design

Add a small typed resolver for Llama attention options:

```
tico/quantization/config/llama_attention.py
```

This resolver reads:

```python
qcfg.model_args["profile"]
qcfg.model_args["attention"]
```

and maps them into explicit attention options.

Update the Llama wrappers so they consume the resolved options:

- QuantLlamaAttention
  - optionally fuses attention scale into k_proj
  - supports HF-style RoPE and pre-negated-sin RoPE
  - supports both batched and head-unrolled attention paths
- QuantLlamaDecoderLayer
  - generate RoPE sine templates according to the selected profile
- QuantLlamaModel

PR #691, opened 2026-05-11
[quantization] Make LlamaAttention export configurable

mhs4670go:
Add configurable Llama attention execution profiles for wrapq.

The new profiles allow choosing between a reference-like GPU evaluation path and the existing NPU export-oriented path. The profile is stored in PTQConfig.model_args and can be set through build_llm_ptq_config().

Supported profiles:
- reference_eval: no scale fusion, HF RoPE convention, batched attention
- npu_export: k_proj scale fusion, pre-negated RoPE sine, head-unrolled attention

Also update Llama model and decoder-layer RoPE template generation to follow the selected profile.

Related: #690


tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
```diff
@@ -169,7 +183,7 @@ def parse_args():
     parser.add_argument(
         "--nsamples_for_qcalibration",
         type=int,
-        default="128",  # almost standard
+        default=128,  # almost standard
```
stamalakhov, line 186: 👍

stamalakhov (approved):
LGTM!

Issue #624, opened 2026-04-13
[quantization] Implement tie_word_embeddings


stamalakhov:
### What
Some models like `unsloth/Llama-3.2-3B-Instruct` use `tie_word_embeddings`, so `lm_head` weights are just a clone of `input_embedings`. But in export this flag is ignored (no known ways to share weights of `input_embeddings` and `lm_head` in export to `circle` (please correct me if i'm wrong)). So to fit to 2Gb of current size constraint of `circle` file we need to quantize `lm_head` to 4 bits (checked - `flatbuffers: cannot grow buffer beyond 2 gigabytes`), which decreases accuracy significantly. It would be nice just to share quantized `lm_head` weights with `input_embeddings` in `circle`.

stamalakhov:
#701 #720 fix the issue

PR #694, opened 2026-05-12
[quantization] Support tied embedding in Qwen3-VL GPTQ

mhs4670go:
This commit supports tied embedding in Qwen3-VL GPTQ pipeline.

Related: #624


mhs4670go:
As https://github.com/Samsung/TICO/pull/695#issuecomment-4428192330 said, let's skip GPTQ of lm head when the embeddings are tied.

PR #697, opened 2026-05-13
[quantization] Export embeddings

mhs4670go:
This commit exports embeddings.

This exports are needed when we export a model layer by layer.


stamalakhov (approved):
LGTM!

PR #670, opened 2026-04-30
[quantization][draft] [NO_MERGE] Fix instability

stamalakhov:
This PR tries to fix instability in llama quantization.

results of ppl evaluation for `LLama3.2-3B-Instruct` quantized using GPTQ only (smse) on 128 samples

torch and transformers packages were the same on all GPUs: 

```

torch 2.10.0
transformers 5.8.0

```

|environment| original| fix_stability==1  |fix_stability==2|fix_stability==3|
|--|-----------|------|--|-|
|GPU_0|12.023604393|11.953644753|12.002959251|12.002959251|
|GPU_1|11.913434029|12.059534073|11.875122070|11.875122070|
|GPU_2|12.054678917|12.145004272|11.989699364|11.989699364|

with `gptq_use_orig_model` enabled:

|environment| fix_stability==0| fix_stability==1  |fix_stability==2|
|--|-----------|------|--|
|GPU_0|11.892453194|11.894100189|11.898873329|
|GPU_1|11.902789116|11.898837090|11.893599510|
|GPU_2|11.905962944|11.892909050|11.891006470|

Relative error:
|parameters| fix_stability*==0 (%)| fix_stability==1  |fix_stability==2|
|--|-----------|------|--|
|NO_gptq_use_orig_model|1.27|1.73|1.16|
|gptq_use_orig_model|0.12|**0.05**|0.07|

Related: #656

TICO-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

Issue #656, opened 2026-04-27
[quantization] Instability on `llama` quantization


stamalakhov:
### What

There is some instability in llama-based models quantization. E.g. 
Running the same command 
```
python tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py --model unsloth/Llama-3.2-3B-Instruct --max_seq_len 2048 --linear_weight_bits 4 --gptq_mse smse --nsamples_for_qcalibration 128 --device cuda --lm_head_weight_bits 4 --save  "ptq_checkpoint"   --no_spinquant  --eval_tasks="mmlu,hellaswag,piqa,truthfulqa" --decode_calibration_steps 8 --sensitivity_path sensitivities_for_unsloth_Llama-3.2-3B-Instruct_wikitext_128_42.pt
```
on two different gpus produced two dirreferent ppl's (`11.86` vs `12.14`).

Running `mse`:
```
python tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py --model unsloth/Llama-3.2-3B-Instruct --max_seq_len 2048 --linear_weight_bits 4 --gptq_mse mse --nsamples_for_qcalibration 128 --device cuda --lm_head_weight_bits 4 --save  "ptq_checkpoint"   --no_spinquant  --eval_tasks="mmlu,hellaswag,piqa,truthfulqa" --decode_calibration_steps 8 
```
also produced two different ppl's (`12.56` vs `12.52`).

Let's make sure that: 
1. this is inevitable
2.  and/or reduce discrepancy of results.

mhs4670go:
I think you can narrow the points.

1. Check if there's the same problem with no_gptq and no_spinquant


2. Check if there's the same problem with 1B model that might use a single GPU. I'm not sure but two gpus could be the reason.

stamalakhov:
> I think you can narrow the points.
> 
>     1. Check if there's the same problem with no_gptq and no_spinquant
> 
>     2. Check if there's the same problem with 1B model that might use a single GPU. I'm not sure but two gpus could be the reason.

@mhs4670go 
Thank you for your reply.
1. ok, spinquant crashes with `out of memory` on llama (so it's just GPTQ/PTQ issue)
2. it's single GPU. I mean running the same script with the same parameters on different GPUs produce different results.

mhs4670go:
FYI, I got same PPL when I tried below options with a single GPU.

```bash
[...]

PR #703, opened 2026-05-13
[quantization] Tune `torch`

stamalakhov:
This PR tunes `torch` environment to improve reproducibility of results.

Draft: https://github.com/Samsung/TICO/pull/670
Related: https://github.com/Samsung/TICO/issues/656

### DIFF
--- a/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
+++ b/tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
@@ -36,9 +36,12 @@
 # =============================================================================
 
 import argparse
+import os
 import pathlib
 import random
 from typing import Any, Optional
+
+import numpy as np
 
 import torch
 import tqdm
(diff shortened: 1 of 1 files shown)


mhs4670go (dismissed):
LGTM

tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py
```diff
@@ -978,6 +981,14 @@ def setup_runtime(args) -> tuple[torch.device, torch.dtype]:
...
     """
     torch.manual_seed(args.seed)
+    np.random.seed(args.seed)
+    random.seed(args.seed)
+    torch.backends.cudnn.benchmark = False
+    torch.use_deterministic_algorithms(True)
+    torch.utils.deterministic.fill_uninitialized_memory = True
+    torch.backends.cuda.matmul.allow_tf32 = False
+    torch.backends.cudnn.allow_tf32 = False
+    os.environ["CUBLAS_WORKSPACE_CONFIG"] = ":4096:8"
```
mhs4670go: Could you move this environment variable setup before importing `torch`?

`CUBLAS_WORKSPACE_CONFIG` should be set before CUDA/cuBLAS is initialized. Since this script already imports `torch` before setting it in `setup_runtime()`, it is not fully clear whether the setting is always applied as intended.

I think it would be safer to move it near the top of the file, before `import torch`, and use `setdefault()` so that a user-provided value is not overwritten:

```python
import argparse
import os

os.environ.setdefault("CUBLAS_WORKSPACE_CONFIG", ":4096:8")

import pathlib
import random
```

Torrero (dismissed):
LGTM

mhs4670go (approved):

Torrero (approved):
LGTM, thank you

PR #702, opened 2026-05-13
[quantization] Add `gptq_use_orig_model_inference`

stamalakhov:
This PR adds `gptq_use_orig_model_inference` to stabilize GPTQ for deep models.

Compare GPTQ quantization stability with different `torch` versions for `HuggingFaceTB/SmolLM2-135M-Instruct`:

|torch| original_PPL|gptq_use_orig_model_inference_PPL|
|--|-|-|
|torch_2_6|22.76|22.95|
|torch_2_9|22.83|22.95|
|torch_2_10|22.81|22.95|

please see additional experiments in the draft #670

Draft: #670 
Related: #656

TICO-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #704, opened 2026-05-13
[quantization] Increase precision

stamalakhov:
This PR increases precision for GPTQ quantization:
1. replaces moving average in Hessian with just a sum
2. increases precision to double in Cholesky and its inversion to improve reproducibility.

Draft: https://github.com/Samsung/TICO/pull/670
Related: https://github.com/Samsung/TICO/issues/656

GPTQ quantization for `HuggingFaceTB/SmolLM2-135M-Instruct` provides the same 

```

│ int16 :    22.95
```
ppl as in https://github.com/Samsung/TICO/pull/702#issue-4437609355

### DIFF
--- a/tico/quantization/algorithm/gptq/gptq.py
+++ b/tico/quantization/algorithm/gptq/gptq.py
@@ -294,7 +294,8 @@ class GPTQ:
                 inp.shape[0] * inp.shape[1], inp.shape[2] * inp.shape[3]
             ).T  # inp.shape =(C_in * krn_size[0] * krn_size[1] * krn_size[2], N * num_patches)
 
-        self.H += inp.double().matmul(inp.double().t()).float().to(self.H.device)  # type: ignore[union-attr]
+        inp = inp.double()
+        self.H += inp.matmul(inp.t()).to(device=self.H.device, dtype=self.H.dtype)  # type: ignore[union-attr]
 
     def fasterquant(
         self,
--- a/tico/quantization/algorithm/gptq/gptq.py
+++ b/tico/quantization/algorithm/gptq/gptq.py
@@ -294,10 +294,7 @@ class GPTQ:
                 inp.shape[0] * inp.shape[1], inp.shape[2] * inp.shape[3]
             ).T  # inp.shape =(C_in * krn_size[0] * krn_size[1] * krn_size[2], N * num_patches)
 
-        self.H *= self.nsamples / (self.nsamples + tmp)
-        self.nsamples += tmp
-        inp = math.sqrt(2 / self.nsamples) * inp.float()
-        self.H += inp.matmul(inp.t()).to(self.H.device)
+        self.H += inp.double().matmul(inp.double().t()).float().to(self.H.device)  # type: ignore[union-attr]
 
     def fasterquant(
         self,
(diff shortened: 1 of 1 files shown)


tico/quantization/algorithm/gptq/gptq.py
```diff
@@ -356,10 +353,10 @@ def fasterquant(
         damp = percdamp * torch.mean(torch.diag(H))
         diag = torch.arange(self.columns, device=self.dev)
         H[diag, diag] += damp
-        H = torch.linalg.cholesky(H)
+        H = torch.linalg.cholesky(H.double()).float()
```
mhs4670go: IIUC, currently the code casts each intermediate result back to `float`.

Since the goal of this PR is to increase precision in Cholesky and its inversion, it would be better to keep H as fp64 across the full linear algebra path and cast back to fp32 only once before entering the quantization loop.

```python
H = H.double()
damp = percdamp * torch.mean(torch.diag(H))
diag = torch.arange(self.columns, device=self.dev)
H[diag, diag] += damp

H = torch.linalg.cholesky(H)
H = torch.cholesky_inverse(H)
Hinv = torch.linalg.cholesky(H, upper=True).float()
```

This preserves the intended double precision during Cholesky -> inverse -> Cholesky, while still keeping the existing fp32 quantization loop mostly unchanged.

tico/quantization/algorithm/gptq/gptq.py
```diff
@@ -294,10 +294,7 @@ def add_batch(self, inp, out):
                 inp.shape[0] * inp.shape[1], inp.shape[2] * inp.shape[3]
             ).T  # inp.shape =(C_in * krn_size[0] * krn_size[1] * krn_size[2], N * num_patches)
 
-        self.H *= self.nsamples / (self.nsamples + tmp)
-        self.nsamples += tmp
-        inp = math.sqrt(2 / self.nsamples) * inp.float()
-        self.H += inp.matmul(inp.t()).to(self.H.device)
+        self.H += inp.double().matmul(inp.double().t()).float().to(self.H.device)  # type: ignore[union-attr]
```
mhs4670go:
```suggestion
        inp = inp.double()
        self.H += inp.matmul(inp.t()).to(device=self.H.device, dtype=self.H.dtype)  # type: ignore[union-attr]
```

mhs4670go: Just for note, since this PR changes Hessian accumulation from a normalized moving average to a raw sum, the scale of `H` changes. The quantization result may still be similar, but the verbose `Losses` value depends on `Hinv` scale, so its absolute value is no longer directly comparable with the previous implementation.

For example, even if quantization quality is similar, this print may change significantly only because of the Hessian scale change:

```python
print("error", torch.sum(Losses).item())
```

For example, if the previous implementation printed:

```
error 1234
```

and this PR prints:

```
error 8.7e7
```

that does not necessarily mean the quantization quality became worse. It may simply come from the Hessian scale change.

mhs4670go: Additonally, looks like `tmp`, `self.nsamples`, and even `math` are now effectively dead code after switching Hessian accumulation to a raw sum.

Would it make sense to either:

1. keep `self.nsamples += tmp` for bookkeeping/debugging purposes, or
2. remove the unused variables/imports entirely?

Right now `tmp` is computed but never used anymore.

stamalakhov: Ok

tico/quantization/algorithm/gptq/gptq.py
stamalakhov: @mhs4670go 
Fixed.
Please take another look.

mhs4670go (approved):
LGTM

PR #700, opened 2026-05-13
[quantization] Make GPTQ lm_head quantization opt-in

mhs4670go:
This commit makes GPTQ lm_head quantization opt-in.

TICO-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #708, opened 2026-05-15
[quantizaiton] Decouple Qwen3-VL GPTQ and PTQ of lm_head

mhs4670go:
This commit decouples Qwen3-VL GPTQ and PTQ.

Related: https://github.com/Samsung/TICO/pull/700

### DIFF
--- a/tico/quantization/wrapq/examples/quantize_qwen3_vl_with_gptq.py
+++ b/tico/quantization/wrapq/examples/quantize_qwen3_vl_with_gptq.py
@@ -102,6 +102,15 @@ def parse_args():
         help="Skip GPTQ and keep the model in floating-point.",
     )
     parser.add_argument(
+        "--gptq_lm_head",
+        action="store_true",
+        default=False,
+        help=(
+            "Apply GPTQ to lm_head. Disabled by default because "
+            "lm_head.weight can be tied with the input embedding table."
+        ),
+    )
+    parser.add_argument(
         "--no_PTQ",
         action="store_true",
         default=False,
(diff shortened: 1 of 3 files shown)


tico/quantization/wrapq/examples/quantize_qwen3_vl_with_gptq.py
```diff
@@ -748,86 +793,141 @@ def get_num_deepstack_mergers(q_m) -> int:
...
+        print(f"Loading sensitivity tensors from {args.sensitivity_path}")
+        return torch.load(args.sensitivity_path, map_location="cpu")
+
+    print("Computing sensitivity tensors for GPTQ …")
+    calibrator = SensitivityCalibrator(
+        model,
+        calib_inputs,
+        show_progress=not args.hide_progress,
+    )
+    return calibrator.compute_sensitivity_info()
```
Torrero, line 893: Could you consider please the following changes for sensitivity computation: 

```
    import os
    ...
    ...
    if args.sensitivity_path is not None and os.path.exists(args.sensitivity_path):
        print(f"Loading sensitivity tensors from {args.sensitivity_path}")
        return torch.load(args.sensitivity_path, map_location="cpu")

    print("Computing sensitivity tensors for GPTQ …")
    calibrator = SensitivityCalibrator(
        model,
        calib_inputs,
        show_progress=not args.hide_progress,
    )

   # saving precalculated sensitivity
   if args.sensitivity_path is not None:
       torch.save(sens, args.sensitivity_path)
   return calibrator.compute_sensitivity_info()
```

mhs4670go: How about adding another option for saving sensitivies in another PR? Using single option for both of loading and saving seems complicated.

Torrero: > How about adding another option for saving sensitives in another PR? Using single option for both of loading and saving seems complicated.

Yes, it's good point, let's introduce it in another PR.

Torrero (approved):
LGTM, thank you!

PR #710, opened 2026-05-15
[quantization] Refactor Qwen3-VL GPTQ/PTQ options

mhs4670go:
This commit refactors Qwen3-VL GPTQ/PTQ stage control options.

We don't need to have fine-grained control of PTQ per modules. Let's make it GPTQ-only options.


Torrero (approved):
LGTM, thank you

Issue #645, opened 2026-04-22
[quantization] Calibration Dataset Selection & Sampling Strategy


mhs4670go:
## 1. Dataset Candidates

We limit the candidates to **practical, widely-used, and easy-to-integrate datasets**.

#### (A) Baseline
- WikiText-2
  - Purpose: baseline comparison
  - Characteristics: plain natural language

#### (B) Instruction / Chat
- Alpaca (Stanford Alpaca)
  - Instruction-following format
  - Single-turn prompt-response pairs

- ShareGPT (filtered subset)
  - Multi-turn conversations
  - Closer to real chat usage

#### (C) Structured Tasks
- FLAN-style mixtures (optional subset)
  - Includes QA, reasoning, summarization
  - Good for diverse activation patterns

---

#### (A) Text-only baseline
- Same as LLM (WikiText / Alpaca subset)

#### (B) Vision-Language
- COCO Captions
  - Image → caption mapping
  - Simple, stable baseline

- VQAv2
  - Image + question → answer
  - Introduces cross-modal attention

- Instruction-style VLM data (if available)
  - Image-grounded instructions
  - Closest to real usage

### 2.1 Number of Samples

Recommended:
- 512 ~ 2048 samples total

Guideline:
- Too small → unstable calibration
- Too large → diminishing returns

### 2.2 Sequence Length

We should match **deployment characteristics**:

| Scenario            | Strategy                          |
|--------------------|----------------------------------|
| Chat / QA          | 128 ~ 512 tokens                 |
| Long context       | include some 1K+ samples         |
| Mixed              | stratified by length             |

#### Option A (Simple Random)
- Uniform random sampling
- Fast and easy baseline

#### Option B (Recommended: Stratified)
Split by:
- Sequence length buckets
- Prompt types (instruction / plain / QA)

Example:
- 50% instruction/chat
- 30% general text
- 20% long-context samples

### 2.4 Prompt Formatting (Important)

For instruction models, use **actual inference format**:

Example:

```
### Instruction:
<instruction>

### Response:
<response>
```

or chat template:

```
<|system|>
...
<|user|>
...
<|assistant|>
...
```

Mismatch here can significantly affect activation distribution.

## 3. VLM Input Construction

For Qwen3-VL:

Each sample should include:
- Image (or dummy image if needed)
- Text prompt

Example:

```
User: What is happening in this image?
<image>
```

Important:
- Maintain **real inference preprocessing**
- Use actual tokenizer + image processor

[...]

PR #711, opened 2026-05-15
[quantization] Introduction of mixed calibration dataset

Torrero:
This PR introduces support for mixed calibration dataset that combines both image-text samples (e.g., VQAv2, COCO) and text-only samples (e.g., Wikitext2, Alpaca). This enables more comprehensive calibration for Qwen3-VL.

Text-only samples requires additional processing in `tico/quantization/algorithm/qwen3_vl_gptq/quantizer.py`

Additionally `--nsamples_for_qcalibration` argument was extended to comma-separated list. 
Formats supported:
- Just a number N -> uses default dataset "vqav2" with N samples
- "dataset:n_samples" -> uses default split
- "dataset:split:n_samples" -> uses specified split
- Multiple datasets: "vqav2:testdev:128,wikitext2:train:128"

Related issue:
- https://github.com/Samsung/TICO/issues/645

---------------------
#### 1. __quantizer.py__ (qwen3_vl_gptq) 

- Added __separate caches for vision batches__: `_vision_cache_args`, `_vision_cache_kwargs`, `_num_vision_batches`
- Vision stages (patch_embed, blocks, merger) now __only process batches with vision inputs__ via `vision_only=True` flag
- Text decoder layers process __all batches__ (both image-text and text-only)
- Prevents errors when text-only inputs lack image tokens
- Added warning when no vision inputs are found in calibration data

#### 2. __utils.py__ (qwen3_vl_gptq) - Cache Handling

- Enhanced `append_batch_to_cache()` to maintain __index alignment__ across batches with different kwargs
- New keys are backfilled with `None` for previous batches
- Missing keys in current batch get `None` placeholder
- Ensures consistent batch indexing across heterogeneous samples

#### 3. __utils.py__ (gptq) - Gradient Safety

- Added check to __skip modules without gradients__ during sensitivity calibration
- Prevents errors when vision modules don't participate in forward pass (text-only inputs)

#### 4. __vlm_eval_utils.py__ - Dataset Support

- Added __text-only datasets__: `alpaca` and adapters

- New `is_text_only` flag in dataset registry to distinguish dataset types

- New functions:

  - `build_text_only_inputs()` - builds processor inputs for text-only data
  - `_build_text_calib_inputs()` - samples random fixed-length sequences from text
  - `get_mixed_calib_inputs()` - combines multiple datasets into single calibration set

#### 5. __quantize_qwen3_vl_with_gptq.py__ 

- Changed `--nsamples_for_qcalibration` from `int` to `str` type

- Supports multiple formats:

  - Just a number: `128` → uses default `vqav2:testdev:128`
  - `dataset:n_samples` → uses default split
  - `dataset:split:n_samples` → uses specified split
  - Multiple: `vqav2:testdev:128,wikitext2:train:128,coco:32`

- Added `load_calib_inputs()` function to parse and load multiple datasets

### DIFF
--- a/tico/quantization/algorithm/qwen3_vl_gptq/utils.py
+++ b/tico/quantization/algorithm/qwen3_vl_gptq/utils.py
@@ -501,6 +501,15 @@ def split_model_inputs(
     )
 
 
+def _num_cached_batches(
+    cache_args: list[list[Any]],
+    cache_kwargs: dict[str, list[Any]],
+) -> int:
+    lengths = [len(v) for v in cache_args]
+    lengths.extend(len(v) for v in cache_kwargs.values())
+    return max(lengths, default=0)
+
+
 def append_batch_to_cache(
     cache_args: list[list[Any]],
     cache_kwargs: dict[str, list[Any]],
@@ -515,15 +524,26 @@ def append_batch_to_cache(
         cache_kwargs: Keyword cache storage.
         *args: Positional batch values.
         **kwargs: Keyword batch values.
+
+    Note:
+        The previous batch count is derived from the existing cache contents
+        and keys occurrences are aligned among cache_args and cache_kwargs.
+        This ensures index alignment across all batches.
     """
+    prev_batches = _num_cached_batches(cache_args, cache_kwargs)
+
     for idx, item in enumerate(args):
         if idx >= len(cache_args):
-            cache_args.append([])
+            cache_args.append([None] * prev_batches)
         cache_args[idx].append(detach_clone_tree(item))
+
+    for key in list(cache_kwargs.keys()):
+        if key not in kwargs:
+            cache_kwargs[key].append(None)
 
     for key, value in kwargs.items():
         if key not in cache_kwargs:
-            cache_kwargs[key] = []
+            cache_kwargs[key] = [None] * prev_batches
         cache_kwargs[key].append(detach_clone_tree(value))
 
 
--- a/tico/quantization/evaluation/vlm_eval_utils.py
+++ b/tico/quantization/evaluation/vlm_eval_utils.py
@@ -14,11 +14,11 @@
 
 import json
 import os
+import random
 import re
 import string
 import tempfile
 
-from collections.abc import Callable
 from typing import Any, Dict, Iterable, List, Optional, Tuple, TypedDict
 
 import torch
@@ -184,16 +184,61 @@ def get_item_coco(ex: dict[str, Any]) -> dict[str, Any]:
     }
 
 
+def get_item_wikitext2(ex: dict[str, Any]) -> dict[str, Any]:
+    """
+    Adapt a Wikitext2 sample to a common format for text-only calibration.
+
+    The returned schema is:
+
+    `{"text": text}`
+
+    Args:
+        ex: Raw dataset example.
+
+    Returns:
+        A normalized text item.
+    """
+    return {"text": ex.get("text", "")}
+
+
+def get_item_alpaca(ex: dict[str, Any]) -> dict[str, Any]:
+    """
+    Adapt an Alpaca-style sample to a common format for text-only calibration.
+
+    Alpaca format has "instruction", "input", and "output" fields.
+    The instruction and input are combined for the text.
+
+    The returned schema is:
+
+    `{"text": text}`
+
+    Args:
+        ex: Raw dataset example.
+
+    Returns:
+        A normalized text item.
+    """
+    instruction = ex.get("instruction", "")
+    input_text = ex.get("input", "")
+    if input_text:
+        text = f"{instruction}\n{input_text}"
+    else:
+        text = instruction
+    return {"text": text}
+
+
 DATASETS: dict[str, dict[str, Any]] = {
     "vqav2": {
         "default_split": "validation",
         "adapter": get_item_vqav2,
         "candidates": ["HuggingFaceM4/VQAv2", "vqav2", "lmms-lab/VQAv2"],
+        "is_text_only": False,
     },
     "textvqa": {
         "default_split": "validation",
         "adapter": get_item_textvqa,
         "candidates": ["textvqa", "HuggingFaceM4/TextVQA", "lmms-lab/textvqa"],
+        "is_text_only": False,
     },
     "coco": {
         "default_split": "val",
@@ -201,12 +246,20 @@ DATASETS: dict[str, dict[str, Any]] = {
         "candidates": [
             "lmms-lab/COCO-Caption2017",
         ],
+        "is_text_only": False,
     },
     "wikitext2": {
         "default_split": "test",
-        "adapter": None,  # Text-only dataset, no adapter needed
+        "adapter": get_item_wikitext2,
         "candidates": ["wikitext"],
         "config": "wikitext-2-raw-v1",
+        "is_text_only": True,
+    },
+    "alpaca": {
+        "default_split": "train",
+        "adapter": get_item_alpaca,
+        "candidates": ["tatsu-lab/alpaca"],
+        "is_text_only": True,
     },
 }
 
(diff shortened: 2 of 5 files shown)


tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+
+    Returns:
+        A list of processor output objects from all datasets.
+    """
+    calib_inputs = []
+
+    for dataset, config in dataset_config.items():
+        n_samples = config.get("n_samples", None)
+        split = config.get("split", None)
+        if n_samples <= 0:
```
dvsav: `n_samples = config.get("n_samples", None)` assumes `n_samples` may become `None`.
`if n_samples <= 0` will then raise a `TypeError`.
How about this:
```python
if n_samples is None or n_samples <= 0:
    continue
```

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+                seed=seed,
+            )
+            calib_inputs.extend(text_inputs)
+        else:
+            # For image-text datasets: take first n_samples directly
+            if adapter is None:
+                print(f"[warn] No adapter for dataset '{dataset}', skipping")
+                continue
+
+            count = 0
```
dvsav: Unused variable

Torrero:
Benchmarks evaluation for Qwen3-vl-4B with mixed calibrated datasets: 'vqav2:testdev:128,wikitext2:train:128' 

| Config ID| PPL | vqav2 | CIDEr | Bleu_4 |
|--|--|--|--|--|
| FP32                       | 10.54 | 0.895   | 0.361 | 0.025 |
| GPTQ_SMSE_w4A16_smoothQuant_vision | 13.09 | 0.7210  | 0.173 | 0.019 |
| GPTQ_MSE_w4A16_smoothQuant_vision | 12.21 | 0.6980  | 0.169 | 0.018 |

tico/quantization/algorithm/qwen3_vl_gptq/utils.py
```diff
@@ -515,15 +515,36 @@ def append_batch_to_cache(
         cache_kwargs: Keyword cache storage.
         *args: Positional batch values.
         **kwargs: Keyword batch values.
+
```
mhs4670go, line 527: I think this cache alignment logic still has a corner case for kwargs-only calibration inputs.

The problem is not that `move_batch_to_device()` returns a dict. That part is fine. The issue is that calibration calls the model as `q_m(**dev_inp)`, so `m_args` is empty and all inputs are stored through `cache_kwargs`. In that case, `cache_args` remains empty, and `expected_batch_idx` becomes `0` even if previous kwargs-only batches have already been cached.

This can break alignment when a text-only batch appears before a vision batch.

Example:

```python
cache_args = []
cache_kwargs = {}

# batch 0: text-only
append_batch_to_cache(
    cache_args,
    cache_kwargs,
    input_ids="text_ids",
    attention_mask="text_mask",
)

# batch 1: image-text
append_batch_to_cache(
    cache_args,
    cache_kwargs,
    input_ids="vlm_ids",
    attention_mask="vlm_mask",
    pixel_values="image_tensor",
    image_grid_thw="grid",
)
```

Expected cache:

```python
cache_kwargs["input_ids"]      == ["text_ids", "vlm_ids"]
cache_kwargs["attention_mask"] == ["text_mask", "vlm_mask"]
cache_kwargs["pixel_values"]   == [None, "image_tensor"]
cache_kwargs["image_grid_thw"] == [None, "grid"]
```

But with the current implementation, because `cache_args` is empty, a newly introduced key is initialized with:

```python
cache_kwargs[key] = [None] * expected_batch_idx  # expected_batch_idx == 0
```

So pixel_values may become:
```python
cache_kwargs["pixel_values"] == ["image_tensor"]
```

This means the logical batch indices are no longer aligned across kwargs. Later, when replaying batch 0, the code may combine text-only `input_ids` with the first vision `pixel_values`, or hit an index error when accessing a later batch.

This is especially relevant for mixed calibration because the user can pass datasets in either order, e.g.:

```
--nsamples_for_qcalibration wikitext2:train:128,vqav2:testdev:128
```

In this order, text-only samples are cached first and vision keys appear later.

A safer fix would be to derive the previous batch count from the existing cache contents, not only from cache_args:

```python
def _num_cached_batches(
    cache_args: list[list[Any]],
    cache_kwargs: dict[str, list[Any]],
) -> int:
    lengths = [len(v) for v in cache_args]
    lengths.extend(len(v) for v in cache_kwargs.values())
    return max(lengths, default=0)

def append_batch_to_cache(
    cache_args: list[list[Any]],
    cache_kwargs: dict[str, list[Any]],
    *args: Any,
    **kwargs: Any,
) -> None:
    prev_batches = _num_cached_batches(cache_args, cache_kwargs)

    for idx, item in enumerate(args):
        if idx >= len(cache_args):
            cache_args.append([None] * prev_batches)
        cache_args[idx].append(detach_clone_tree(item))

    for key in list(cache_kwargs.keys()):
        if key not in kwargs:
            cache_kwargs[key].append(None)

    for key, value in kwargs.items():
        if key not in cache_kwargs:
            cache_kwargs[key] = [None] * prev_batches
        cache_kwargs[key].append(detach_clone_tree(value))
```

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+
+        print(
+            f"[info] Loading calibration data from '{dataset}' (split={split}): {n_samples} samples"
+        )
+
+        if dataset not in DATASETS:
+            print(f"[warn] Unknown dataset '{dataset}', skipping")
+            continue
+
+        ds, adapter = get_dataset(dataset=dataset, n=n_samples, split=split)
```
mhs4670go: For text-only datasets, `n_samples` is used both as the number of source dataset rows to read and as the number of calibration windows to generate.

For example, `wikitext2:train:128` loads **only 128 source rows**, concatenates them, and then samples 128 windows from that concatenated text.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+                    processor=processor,
+                    image=item["image"],
+                    question=item["question"],
+                    return_tensors="pt",
+                    max_seq_len=max_seq_len,
+                )
+                calib_inputs.append(inputs)
+
+    print(f"[info] Total calibration samples: {len(calib_inputs)}")
+    return calib_inputs
```
mhs4670go, line 1103:
```suggestion
    if not calib_inputs:
        raise ValueError(
            "No calibration inputs were loaded. "
            "Please check --nsamples_for_qcalibration."
        )

    print(f"[info] Total calibration samples: {len(calib_inputs)}")
    return calib_inputs
```

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+        n_samples = config.get("n_samples", None)
+        split = config.get("split", None)
+        if n_samples <= 0:
+            continue
+
+        print(
+            f"[info] Loading calibration data from '{dataset}' (split={split}): {n_samples} samples"
+        )
+
+        if dataset not in DATASETS:
```
mhs4670go, line 1035: Seems that `{"dataset:split": n_samples}` key will be unknown. Is it intentional?

Parsing could be done like this instead.

```python
dataset_specs = [
    {"name": "wikitext2", "split": "train", "n_samples": 128},
    {"name": "wikitext2", "split": "test", "n_samples": 128},
]

```

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -856,3 +909,183 @@ def get_calib_inputs(
...
+            print(f"[warn] Unknown dataset '{dataset}', skipping")
+            continue
+
+        ds, adapter = get_dataset(dataset=dataset, n=n_samples, split=split)
+        is_text_only = DATASETS[dataset].get("is_text_only", False)
+
+        if is_text_only:
```
mhs4670go, line 1041: Text only inputs are just concatenated unlike vision inputs that are chagned with chat templates. I think we need to change text inputs, too. But, seems the code change would be big. Let's make it TODO.

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: Ahh, I've just forgot to change the comments, thank you. Actually input format is following: 
```
{"dataset": {"split":"split","n_samples": n_samples}}
```

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: fixed it

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: Added `TODO`

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: fixed it

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: Changed

tico/quantization/evaluation/vlm_eval_utils.py
Torrero: fixed it

tico/quantization/algorithm/qwen3_vl_gptq/utils.py
Torrero: Thank you, I fixed it

mhs4670go (approved):
LGTM

dvsav (approved):
LGTM

PR #724, opened 2026-05-26
[examples] Support weight tying and mixed calibration

mhs4670go:
This commit supports features that are introduced with recent commits.


Torrero (approved):
LGTM, thank you!

PR #729, opened 2026-05-27
[quantization] Fix bugs

mhs4670go:
This commit fixes some bugs in wrappers.

- Use quantized input in SiLU wrapper.
- Check right condition in Conv3D wrapper.


tico/quantization/wrapq/wrappers/nn/quant_silu.py
```diff
@@ -51,7 +51,7 @@ def forward(self, x: torch.Tensor):
         s = torch.sigmoid(x_q)
         s = self._fq(s, self.obs_sig)
 
-        y = x * s
+        y = x_q * s
```
stamalakhov, line 54: 👍

stamalakhov (approved):
LGTM!

PR #730, opened 2026-05-27
[quantization] Use right rotate_embedding

mhs4670go:
This commit uses wrapped rotate_embedding instead of fp module.


Torrero (approved):
LGTM

PR #731, opened 2026-05-27
[quantization] Preserve Qwen text attention bool masks

mhs4670go:
Build Qwen3-VL text attention masks through a shared helper instead of overwriting bool masks with a causal mask.

- Convert bool/int keep masks to additive masks
- Combine padding masks with causal masks
- Normalize supported mask shapes for logits broadcasting
- Use cache-aware causal mask rows when k_len includes past tokens


Torrero (approved):
LGTM

PR #732, opened 2026-05-27
[quantization] Fix padding logic in conv3d wrapper

mhs4670go:
This commit fixes padding logic in the conv3d wrapper.

Current logic doesn't consider asymmetric padding.


Torrero (approved):
LGTM, thank you!

Issue #494, opened 2026-02-13
[wrapq] Bug in _all_observers()


CMDR-DiTRay:
## What?
The `_all_observers()` method in Quantization Wrappers yields observers from child modules, but the child modules (PTQWrapper instances) don't actually own observers themselves.

When the _all_observers() method in the parent modules (like QuantQwen3VLTextMLP, QuantQwen3VLTextAttention, etc.) calls yield from m._all_observers() on a PTQWrapper instance, it gets an empty iterator because:

1. PTQWrapper._all_observers() returns an empty iterator ()
2. The actual observers are owned by the wrapped module inside the PTQWrapper

## Where?
This affetcs at least TextAttention (`tico/quantization/wrapq/wrappers/qwen_vl/quant_text_attn.py`) and Qwen3VLVisionMLP (`tico/quantization/wrapq/wrappers/qwen_vl/quant_vision_mlp.py`) Quantization Wrappers.

PR #733, opened 2026-05-27
[quantization] Refactor observer traversal to separate local and recursive loop

mhs4670go:
Update QuantModuleBase so _all_observers() is reserved for observers owned directly by each wrapper, while public named_observers()/get_observer() can recursively enumerate child quant modules with qualified names.

This avoids double-processing child observers during calibration/freeze while preserving recursive observer lookup for debugging and tests.

Related: https://github.com/Samsung/TICO/issues/494#issuecomment-3935517335


stamalakhov (approved):
LGTM!

PR #734, opened 2026-05-27
[quantization] Fix QuantGELU to preserve nn.GELU approximate mode

mhs4670go:
QuantGELU previously called _gelu(x) without forwarding the original nn.GELU.approximate setting, so nn.GELU(approximate="tanh") was executed as exact GELU. Store the original approximate mode and use it in forward. Add NO_QUANT parity coverage for tanh GELU.


stamalakhov (approved):
LGTM!

PR #743, opened 2026-05-28
[quantization] Support Qwen3VL inputs_embeds without attention mask

mhs4670go:
Use inputs_embeds as the shape/device source when Qwen3VL receives text-only inputs_embeds without input_ids or attention_mask.

Previously, the text-only RoPE fallback unconditionally referenced `input_ids` in the no-attention-mask path, causing inputs_embeds-only forward calls to fail with NoneType attribute errors.


Torrero (approved):
LGTM

PR #597, opened 2026-03-31
[quantization] [draft] Introduce smse_for_gptq/mse_for_gptq

stamalakhov:
This PR introduces smse_for_gptq to improve accuracy.

Please see results in https://github.com/Samsung/TICO/pull/597#issuecomment-4562432076.

TICO-DCO-1.0-Signed-off-by: s.malakhov <s.malakhov@partner.samsung.com>

PR #746, opened 2026-05-28
[quantization] Introduce GPTQ adjusted  mse

stamalakhov:
This PR introduces mse_for_gptq/smse_for_gptq to improve accuracy and ajusts codes accordingly.

Please see results in the draft.

Draft: #597 

### DIFF
--- a/tico/quantization/algorithm/gptq/gptq.py
+++ b/tico/quantization/algorithm/gptq/gptq.py
@@ -333,6 +333,11 @@ class GPTQ:
         H[dead, dead] = 1
         W[:, dead] = 0
 
+        if groupsize != -1 and self.quantizer.mse in {"mse_for_gptq", "smse_for_gptq"}:
+            raise ValueError(
+                "GPTQ-adjusted MSE currently does not support groupsize != -1"
+            )
+
         if static_groups:
             import copy
 
@@ -342,6 +347,8 @@ class GPTQ:
                 quantizer.find_params(W[:, i : (i + groupsize)], weight=True)
                 groups.append(quantizer)
 
+        perm = None
+        invperm = None
         if actorder:
             perm = torch.argsort(torch.diag(H), descending=True)
             W = W[:, perm]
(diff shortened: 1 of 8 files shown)


tico/quantization/algorithm/gptq/gptq.py
```diff
@@ -361,6 +361,8 @@ def fasterquant(
         H = torch.linalg.cholesky(H, upper=True).float()
         Hinv = H
 
+        self.quantizer.update(W, Hinv, perm)
```
mhs4670go, line 371: `perm` is defined only when the `actorder` is True. Therefore, this line may got error when the `actorder` is False.

```python
perm = None
invperm = None
if actorder:
    perm = torch.argsort(torch.diag(H), descending=True)
    W = W[:, perm]
    H = H[perm][:, perm]
    invperm = torch.argsort(perm)

...
self.quantizer.update(W, Hinv, perm)
```

tico/quantization/algorithm/gptq/gptq.py
mhs4670go, line 1: Seems that if `groupsize` is not -1, below logci is needed.

```python
if groupsize != -1 and self.quantizer.mse in {"mse_for_gptq", "smse_for_gptq"}:
    raise ValueError("GPTQ-adjusted MSE currently does not support groupsize != -1")

if static_groups:
    import copy

    groups = []
    for i in range(0, self.columns, groupsize):
        quantizer = copy.deepcopy(self.quantizer)
        quantizer.find_params(W[:, i : (i + groupsize)], weight=True)
        groups.append(quantizer)
```

tico/quantization/algorithm/gptq/gptq.py
stamalakhov: Ahh. 😢

tico/quantization/algorithm/gptq/gptq.py
stamalakhov: Yep .You're right. Sorry. I missed it. Although it's possible to generalize the approach, but let's do it when it will be needed.

stamalakhov: @mhs4670go 
Fixed. Please take another look.

mhs4670go (approved):
LGTM

PR #751, opened 2026-05-30
[quantization] Add stubs for decord and lmms_eval

mhs4670go:
This commit adds stubs for decord and lmms_eval.

TICO-DCO-1.0-Signed-off-by: seongwoo mhs4670go@naver.com

### DIFF
--- a/test/quantization/recipes/optional_dependency_stubs.py
+++ b/test/quantization/recipes/optional_dependency_stubs.py
@@ -32,6 +32,8 @@ def install_optional_dependency_stubs() -> None:
     """Install lightweight stubs for optional recipe dependencies."""
     _install_datasets_stub()
     _install_lm_eval_stub()
+    _install_decord_stub()
+    _install_lmms_eval_stub()
 
 
 def _has_attrs(module: types.ModuleType, names: tuple[str, ...]) -> bool:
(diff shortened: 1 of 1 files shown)


Torrero (approved):
LGTM, thank you

dvsav:
## QnA for Myself

1. **Who uses `test/quantization/recipes/optional_dependency_stubs.py` module?**
This module is used by unit tests only.

2. **Why do unit tests need `optional_dependency_stubs.py` module?**
Unit tests may indirectly depend on packages that are not installed in the test environment. `optional_dependency_stubs.py` module contains finction `install_optional_dependency_stubs()` called by the unit test modules that "installs" the required stub packages if the real packages are not available. Without this measure `import xxx` would fail with `ModuleNotFoundError` exception.

3. **What does it mean "to install a stub package"?**
That means:
    - Creating a new module instance for the parent package (`decord_module = types.ModuleType("decord")`) and registreing it in `sys.modules` ( `sys.modules["decord"] = decord_module`).
    - For each submodule of the parent package creating a new module instance (`bridge_module = types.ModuleType("decord.bridge")`), linking it as a parent's attribute (`setattr(decord_module, "bridge", bridge_module)`)
     and registering it in `sys.modules` (`sys.modules["decord.bridge"] = bridge_module`)
    - Filling-in the attributes of the package and its submodules (e.g. `setattr(decord_module, "VideoReader", VideoReader)`). This may require creating stub classes and functions that normally raise a `RuntimeError` when called to create a safety net for the case if one forgets to mock them in the unit test.

4. **Why can the unit tests function without the real packages available?**
Just importing a package doesn't mean one needs to use its full functionality during unit tests. On the other hand some functionality that is used but not available can be mocked/patched using `unittest.mock.patch`.

PR #266, opened 2025-08-06
Fuse LlamaRMSNorm class to Circle RMSNorm op

seockho-kim:
This shows how to fuse LlamaRMSNorm class to Circle RMSNorm operation.

TICO-DCO-1.0-Signed-off-by: Seockho Kim seockho.kim@samsung.com

Like https://github.com/Samsung/TICO/pull/217,
it does not check patterns, but replaces LlamaRMSNorm with custom op.

PR #754, opened 2026-06-02
[passes] Fuse RMSNorm to Circle RMSNorm op

dahlinPL:
This commit adds FuseRmsNorm() to the circle_legalize PassManager

### DIFF
--- a/tico/passes/fuse_rms_norm.py
+++ b/tico/passes/fuse_rms_norm.py
@@ -1,0 +1,100 @@
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    import torch.fx
+import torch
+from torch.export import ExportedProgram
+
+from tico.utils.graph import create_node
+from tico.utils.passes import PassBase, PassResult
+from tico.utils.trace_decorators import trace_graph_diff_on_pass
+from tico.utils.utils import is_target_node
+from tico.utils.validate_args_kwargs import RMSNormArgs
+
+
+@trace_graph_diff_on_pass
+class FuseRmsNorm(PassBase):
+    """
+    This pass fuses aten.rms_norm into circle_custom.rms_norm.
+
+    aten.rms_norm already includes the weight multiplication internally,
+    so its output is equivalent to circle_custom.rms_norm.
+
+    [before]
+
+            input (tensor, normalized_shape, weight, eps)
+                |
+           aten.rms_norm
+                |
+              output
+
+    [after]
+
+            input (tensor, weight, eps)
+                |
+        circle_custom.rms_norm
+                |
+              output
+
+    Note: normalized_shape is dropped because circle_custom.rms_norm
+    infers the normalization dimension from the weight shape.
+    """
+
+    def __init__(self):
+        super().__init__()
+
+    def call(self, exported_program: ExportedProgram) -> PassResult:
+        gm = exported_program.graph_module
+        graph: torch.fx.Graph = gm.graph
+        modified = False
+
+        for node in graph.nodes:
+            if not is_target_node(node, [torch.ops.aten.rms_norm.default]):
+                continue
+
+            args = RMSNormArgs(*node.args, **node.kwargs)
+            input_ = args.input
+            weight = args.weight
+            eps = args.eps
+
+            # weight is required for circle_custom.rms_norm
+            if weight is None:
+                continue
+
+            # Use default eps if not provided (PyTorch default is 1e-5)
+            if eps is None:
+                eps = 1e-5
+
+            with gm.graph.inserting_before(node):
+                rms_norm_node = create_node(
+                    graph,
+                    torch.ops.circle_custom.rms_norm.default,
+                    (input_, weight, eps),
+                    origin=node,
+                )
+
+            # Reset meta to allow propagate_meta=True in replace_all_uses_with
+            # (same pattern as DecomposeGroupNorm)
+            rms_norm_node.meta = {}
+            node.replace_all_uses_with(rms_norm_node, propagate_meta=True)
+            modified = True
+
+        gm.graph.eliminate_dead_code()
+        gm.graph.lint()
+        gm.recompile()
+
+        return PassResult(modified)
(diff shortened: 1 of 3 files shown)


dahlinPL:
sample code used for testing:

```python
import torch
import torch.nn as nn

import tico


class RMSNormModel(nn.Module):
    """Simple model with RMSNorm for LLaMA/Qwen-style normalization."""

    def __init__(self):
        super().__init__()
        self.rms_norm = nn.RMSNorm(normalized_shape=64, eps=1e-6)

    def forward(self, x):
        return self.rms_norm(x)


if __name__ == "__main__":
    torch_module = RMSNormModel()
    example_inputs = (torch.randn(1, 8, 64, dtype=torch.float32),)

    circle_model = tico.convert(torch_module.eval(), example_inputs)
    circle_model.save("rmsnorm.circle")

    print("Circle model saved to: rmsnorm.circle")
```

result from main branch with `tico==0.2.0.dev260601`:
image


result from this PR:
image

dahlinPL:
local run of test added:

```console
:~/git/fork/TICO$ ./ccex test -k FuseRmsNorm
RUN unit tests with -k FuseRmsNorm ...
2026-06-02 16:28:08.363913: I tensorflow/core/util/port.cc:153] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
2026-06-02 16:28:08.390370: I tensorflow/core/platform/cpu_feature_guard.cc:210] This TensorFlow binary is optimized to use available CPU instructions in performance-critical operations.
To enable the following instructions: AVX2 AVX_VNNI FMA, in other operations, rebuild TensorFlow with the appropriate compiler flags.
2026-06-02 16:28:09.272325: I tensorflow/core/util/port.cc:153] oneDNN custom operations are on. You may see slightly different numerical results due to floating-point round-off errors from different computation orders. To turn them off, set the environment variable `TF_ENABLE_ONEDNN_OPTS=0`.
test_pass (unit_test.passes.test_fuse_rms_norm.FuseRmsNormTest) ... ok

----------------------------------------------------------------------
Ran 1 test in 0.023s

OK
:~/git/fork/TICO$ 
```

mhs4670go:
Just curiosity, which model need this feature?

dahlinPL:
> Just curiosity, which model need this feature?

No specific model, I was just investigating why tests designed to use RMSNorm are failing and realized that only Llama customized pass exists (introduced in https://github.com/Samsung/TICO/pull/266) and "clean" RMSNorm is not fused into Circle as a single op. Therefore, model compilation is failing because of lack of support for rsqrt in the compiler.

tico/passes/fuse_rms_norm.py
mhs4670go, line 1: It looks more like a direct op serialization case than a graph rewrite pass. Could you implenet similar implementation in `tico/serialize/operators/`?

tico/passes/fuse_rms_norm.py
```diff
@@ -0,0 +1,100 @@
...
+    [after]
+
+            input (tensor, weight, eps)
+                |
+        circle_custom.rms_norm
+                |
+              output
+
+    Note: normalized_shape is dropped because circle_custom.rms_norm
+    infers the normalization dimension from the weight shape.
```
mhs4670go, line 54: CircleRMSNorm computes RMS over the last axis only. Therefore, following check shold be added.

```python
if len(args.normalized_shape) != 1:
    raise NotYetSupportedError(
        "Only 1-D normalized_shape RMSNorm is supported"
    )
```

tico/passes/fuse_rms_norm.py
```diff
@@ -0,0 +1,100 @@
...
+            args = RMSNormArgs(*node.args, **node.kwargs)
+            input_ = args.input
+            weight = args.weight
+            eps = args.eps
+
+            # weight is required for circle_custom.rms_norm
+            if weight is None:
+                continue
+
+            # Use default eps if not provided (PyTorch default is 1e-5)
```
mhs4670go, line 78: According to the [spec](https://docs.pytorch.org/docs/2.12/generated/torch.nn.modules.normalization.RMSNorm.html), eps is set to None.

mhs4670go: Let's just disallow this case.

```python
if eps is None:
    raise NotYetSupportedError("aten.rms_norm with eps=None is not supported yet")
```

tico/passes/fuse_rms_norm.py
dahlinPL: hmm I guess you're right :)

dahlinPL: I guess it will be easier to raise this change in separate PR, but before I do that, please let me know your opinion

I could keep `serialize/operators/adapters/llama_rmsnorm.py` and `quantization/wrapq/wrappers/ops/quant_rmsnorm.py` as-is (using `circle_custom.rms_norm`) and add `AtenRMSNormVisitor` alongside `RMSNormVisitor`. However, I think it makes sense to simplify and:

- replace all occurrences of `circle_custom.rms_norm` with `aten.rms_norm`
- remove `CircleRMSNorm` from `register_custom_op.py` completely

This gives us a clean workflow:

- normalize to `aten.rms_norm` when needed (e.g. for HuggingFace's LlamaRMSNorm which uses manual pow/mean/rsqrt/mul)
- `serialize/operators/op_rmsnorm.py` has a single pipeline for all RMSNorm ops (regardless of whether they came from `torch.nn.RMSNorm` or were normalized by an adapter)

mhs4670go: Replacing all `circle_custom.rms_norm` users with `aten.rms_norm` and removing `CircleRMSNorm` sounds like a good cleanup. But it touches the HF adapter, quant wrapper, custom-op registration/fake impl, and tests. It also requires synthesizing `normalized_shape` from `weight.shape` and preserving the same restrictions (`weight is not None`, `len(normalized_shape) == 1`, `eps is not None`). 

Actually, I think that would be better handled in a follow-up PR or when we really need it.

For this PR, just following codes looks enough.

```python
# tico/serialize/operators/op_rmsnorm.py

from tico.serialize.circle_mapping import extract_shape
from tico.utils.errors import NotYetSupportedError
from tico.utils.validate_args_kwargs import CircleRMSNormArgs, RMSNormArgs

@register_node_visitor
class RMSNormVisitor(NodeVisitor):
    target = [
        torch.ops.circle_custom.rms_norm.default,
        torch.ops.aten.rms_norm.default,
    ]

    def _parse_args(self, node):
        if node.target == torch.ops.aten.rms_norm.default:
            args = RMSNormArgs(*node.args, **node.kwargs)

            if args.weight is None:
                raise NotYetSupportedError("RMSNorm without weight is not supported")

            if len(args.normalized_shape) != 1:
                raise NotYetSupportedError(
                    "Only 1-D normalized_shape RMSNorm is supported"
                )

            if list(extract_shape(args.weight)) != list(args.normalized_shape):
                raise NotYetSupportedError(
                    "RMSNorm weight shape should match normalized_shape"
                )

            eps = args.eps
            if eps is None:
                raise NotYetSupportedError("RMSNorm eps=None is not supported yet")

            return args.input, args.weight, eps

        args = CircleRMSNormArgs(*node.args, **node.kwargs)
        return args.input, args.weight, args.eps
```

dahlinPL: implemented in https://github.com/Samsung/TICO/pull/766

dahlinPL:
delivered with https://github.com/Samsung/TICO/pull/766

PR #755, opened 2026-06-03
[interpreter] Fix Circle inference input binding

mhs4670go:
Use `ModelInputSpec.bind` in `tico/interpreter/infer.py` to
 produce `user_inputs` so `*args`/`**kwargs` are bound and
ordered according to the Circle model input signature.

TICO-DCO-1.0-Signed-off-by: seongwoo mhs4670go@naver.com


stamalakhov (approved):
LGTM!

PR #756, opened 2026-06-03
[github] Update CI runners to Ubuntu 24.04

mhs4670go:
This commit updates ubuntu version.


zetwhite (approved):
LGTM 👍 

Just out of curiosity, does TICO only support 24.04 officially?

mhs4670go:
@zetwhite No. I just upgrade the version since 22.04 is quite old now. Also, TICO is pure python library. Therefore, there's no dependency on ubuntu version.

PR #760, opened 2026-06-04
[quantization] Introduce Quantize/Dequantize for MX

stamalakhov:
This PR adds definition for Quantize/Dequantize stubs for MX format and adds tests for it.

Draft: #436

### DIFF
--- a/tico/utils/register_custom_op.py
+++ b/tico/utils/register_custom_op.py
@@ -705,6 +705,54 @@ def CircleQuantizeMX():
         return input_
 
 
+def CircleQuantizeMXDecomposed():
+    @custom_op("circle_custom::quantize_mx_decomposed", mutates_args=())
+    def quantize_mx(
+        input_: torch.Tensor,
+        elem_format: str,
+        axis: int,
+        shared_exp_method: str = "max",
+        round: str = "nearest",
+    ) -> torch.Tensor:
+        # this op should be fake one, so please consider different quantization scheme in case it failed here
+        assert False
+        return input_.clone()
+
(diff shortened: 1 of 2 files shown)


tico/utils/register_custom_op.py
```diff
@@ -705,6 +705,54 @@ def _(
...
+        axis: int,
+        shared_exp_method: str = "max",  # Fixed
+        round: str = "nearest",  # Fixed
+    ) -> torch.Tensor:
+        return input_
+
+
+def CircleDeQuantizeMXDecomposed():
+    @custom_op("circle_custom::dequantize_mx_decomposed", mutates_args=())
+    def quantize_mx(
```
mhs4670go:
```suggestion
    def dequantize_mx(
```

tico/utils/register_custom_op.py
```diff
@@ -705,6 +705,54 @@ def _(
...
+        input_: torch.Tensor,
+        elem_format: str,
+        axis: int,
+        shared_exp_method: str = "max",
+        round: str = "nearest",
+    ) -> torch.Tensor:
+        # this op should be fake one, so please consider different quantization scheme in case it failed here
+        assert False
+        return input_.clone()
```
mhs4670go: How about throwing an error instead?

```suggestion
        raise RuntimeError(
    "circle_custom::quantize_mx_decomposed is a fake-only op and must not be executed with real tensors"
    )
    return input_.new_empty(input_.size())
```

stamalakhov:
@mhs4670go 
Fixed. 
Please take another look.

PR #762, opened 2026-06-06
Canonicalize MX fake quant export through Q-DQ

mhs4670go:
Introduce a separate MX fake-quant frontend op and lower it to a logical quantize_mx/dequantize_mx pair before Circle export.

Related: #436

### DIFF
--- a/tico/serialize/circle_mapping.py
+++ b/tico/serialize/circle_mapping.py
@@ -63,20 +63,9 @@ def str_to_circle_dtype(
         "int64": circle.TensorType.TensorType.INT64,
         "bool": circle.TensorType.TensorType.BOOL,
         "uint4": circle.TensorType.TensorType.UINT4,
-        # TODO Add more dtypes
+        "mxint8": circle.TensorType.TensorType.MXINT8,
+        "mxfp4": circle.TensorType.TensorType.MXFP4,
     }
-    optional_dtypes = {
-        "mxint8": "MXINT8",
-        "mxfp4": "MXFP4",
-        "mxfp6_e3m2": "MXFP6_E3M2",
-        "mxfp6_e2m3": "MXFP6_E2M3",
-        "mxfp8_e4m3": "MXFP8_E4M3",
-        "mxfp8_e5m2": "MXFP8_E5M2",
-    }
-    for dtype, circle_name in optional_dtypes.items():
-        circle_dtype = getattr(circle.TensorType.TensorType, circle_name, None)
-        if circle_dtype is not None:
-            dmap[dtype] = circle_dtype
     if str_dtype not in dmap:
         raise RuntimeError(f"Unsupported dtype {str_dtype}")
     circle_type = dmap[str_dtype]
--- a/tico/serialize/circle_mapping.py
+++ b/tico/serialize/circle_mapping.py
@@ -47,10 +47,10 @@ def to_circle_dtype(
     return circle_type
 
 
-# Convert str dtype used in QuantParam to circle dtype
 def str_to_circle_dtype(
     str_dtype: str,
 ) -> int:
+    """Convert a QuantParam dtype string to a Circle tensor dtype."""
     dmap = {
         "float32": circle.TensorType.TensorType.FLOAT32,
         "float": circle.TensorType.TensorType.FLOAT32,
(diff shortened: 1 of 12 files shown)


mhs4670go:
@stamalakhov 

Thanks for the draft. I slighty changed a bit the codes that introduce `mx_fake_quantize`.

Then `DecomposeFakeQuantize` lowers `mx_fake_quantize` to `quantize_mx -> dequantize_mx`, and `FoldQuantOps` folds that Q-DQ pair into the producer node's `QPARAM` metadata, just like affine quantization already does. This avoids overloading `quantize_mx` with two different meanings. 

Please feel free to give your opinions.

tico/utils/register_custom_op.py
```diff
@@ -664,43 +665,93 @@ def _(
...
-    # This operator conducts fake-quantization of microscaling
-    # NOTE Why using "quantize"_mx not "fake_quantize"_mx?
-    # To align with function name of microxcaling repo.
-    # https://github.com/microsoft/microxcaling/blob/v1.1.0/mx/mx_ops.py#L173
-    @custom_op("circle_custom::quantize_mx", mutates_args=())
-    def quantize_mx(
+def CircleMXFakeQuantize():
+    """Register the eager MX fake-quantization custom operator."""
+
+    @custom_op("circle_custom::mx_fake_quantize", mutates_args=())
```
stamalakhov, line 671: 👍

tico/serialize/circle_mapping.py
```diff
@@ -65,10 +65,20 @@ def str_to_circle_dtype(
         "uint4": circle.TensorType.TensorType.UINT4,
         # TODO Add more dtypes
     }
-
+    optional_dtypes = {
```
stamalakhov: :+1: 
Although i'm not sure why these:
```
 "mxint8": "MXINT8",
 "mxfp4": "MXFP4",
```
 can not be inserted directly into `dmap`.

stamalakhov:
> @stamalakhov
> 
> Thanks for the draft. I slighty changed a bit the codes that introduce `mx_fake_quantize`.
> 
> Then `DecomposeFakeQuantize` lowers `mx_fake_quantize` to `quantize_mx -> dequantize_mx`, and `FoldQuantOps` folds that Q-DQ pair into the producer node's `QPARAM` metadata, just like affine quantization already does. This avoids overloading `quantize_mx` with two different meanings.
> 
> Please feel free to give your opinions.

@mhs4670go 
I've tried to keep existing codes untouched. So that they can be refactored later. I planned to introduce `Q-DQ`  after #760, I I supposed to introduce serialization of `MX` in another PR, as it is contained in another module. [Draft](https://github.com/Samsung/TICO/pull/436) contains all of these features. Should i approve #762?

mhs4670go:
@stamalakhov 

Thanks for the clarification. I understand your intention: #760 introduces the MX Q/DQ stubs first, and the Q-DQ canonicalization and serialization could be layered on top in later PRs. I'm sorry if my comment came across as dismissing your code. That was not my intention at all.

Even though it can be refactored later, my preference was let #762 supersede #760 if you are comfortable with the scope, because the fake-quant op, Q-DQ decomposition, and folding logic are tightly coupled and are easier to review as one coherent export flow.

From the design perspective, this keeps the MX export path consistent with the affine fake-quant path: fake quant API -> Q-DQ decomposition -> qparam folding into producer metadata -> Circle qmodel export.

If you're okay with it, I'd prefer to proceed directly on top of the refactored code, even if it may require a bit of extra work.

tico/serialize/circle_mapping.py
mhs4670go: Good point. I used the indirect `getattr` mapping only to avoid breaking environments where `circle_schema` does not define MX tensor types yet. In that case, non-MX dtype conversion can still work and MX dtype conversion fails only when requested.

However, since this PR is adding MX export support, it is reasonable to require a schema version that already has `MXINT8`/`MXFP4`. I agree direct entries in `dmap` are simpler and clearer. I will update it that way.

stamalakhov:
> If you're okay with it, I'd prefer to proceed directly on top of the refactored code, even if it may require a bit of extra work.

I'm ok with it.

stamalakhov (approved):
LGTM!

PR #763, opened 2026-06-08
[evaluation]  Continue video-mme dataset support for qwen3-vl adapter

Torrero:
This commit introduces:
- continue video-mme dataset support for qwen3-vl adapter
- automatic video frame resizing to fit within the static context budget of PTQ-quantized models
- refactoring videomme arguments
- tests

### DIFF
--- a/tico/quantization/evaluation/lmms_eval_utils.py
+++ b/tico/quantization/evaluation/lmms_eval_utils.py
@@ -700,9 +700,9 @@ def evaluate_vlm_on_tasks(
         eval_kwargs["limit"] = limit
     if use_cache is not None:
         eval_kwargs["use_cache"] = use_cache
-    if max_new_tokens is not None:
-        gen_kwargs = f"max_new_tokens={max_new_tokens}"
-        eval_kwargs["gen_kwargs"] = gen_kwargs
+
+    gen_kwargs = f"max_new_tokens={_effective_max_new_tokens}"
+    eval_kwargs["gen_kwargs"] = gen_kwargs
 
     # Auto-register custom task directories (e.g. videomme_mini) by
     # creating a ``TaskManager`` with ``include_path`` pointing to the
--- a/tico/quantization/evaluation/lmms_eval_utils.py
+++ b/tico/quantization/evaluation/lmms_eval_utils.py
@@ -438,6 +438,82 @@ def _patch_from_pretrained_to_reuse_model(model, processor):
     return _restore()
 
 
+def _patch_video_frame_budget(
+    *,
+    max_pixels: int,
+    min_pixels: int,
+    max_num_frames: int,
+):
+    """Patch the lmms-eval ``Qwen3_VL`` wrapper to enforce the video frame budget.
+
+    This is a safety net that overrides ``max_pixels``, ``min_pixels``, and
+    ``max_num_frames`` on the ``Qwen3_VL`` wrapper after instantiation.
+    While the primary mechanism is passing these values via ``model_args``,
+    this patch ensures the budget is enforced even if the model_args parsing
+    has edge cases.
+
+    The patch works by monkey-patching ``__init__`` to set the instance
+    attributes after the original initialization.
+
+    The patch is automatically undone when the returned context manager exits.
+
+    Args:
+        max_pixels: Maximum pixels per video frame.
+        min_pixels: Minimum pixels per video frame.
+        max_num_frames: Maximum number of video frames.
+
+    Returns:
+        A context manager that restores the original ``__init__`` on exit.
+    """
+    import contextlib
+
+    _patched_classes = []
+
+    # Patch both the simple and chat Qwen3_VL wrappers
+    for _module_path, _class_name in [
+        ("lmms_eval.models.simple.qwen3_vl", "Qwen3_VL"),
+        ("lmms_eval.models.chat.qwen3_vl", "Qwen3_VL"),
+    ]:
+        try:
+            import importlib
+
+            _module = importlib.import_module(_module_path)
+            _cls = getattr(_module, _class_name)
+            _patched_classes.append(_cls)
+        except (ImportError, AttributeError):
+            continue
+
+    if not _patched_classes:
+        return contextlib.nullcontext()
+
+    _original_inits = {cls: cls.__init__ for cls in _patched_classes}
+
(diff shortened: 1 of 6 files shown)


tico/quantization/recipes/adapters/qwen3_vl.py
```diff
@@ -421,6 +422,20 @@ def evaluate(self, ctx: RecipeContext) -> None:
...
 
+        videomme = eval_cfg.get("videomme", {})
+        if videomme.get("enabled", False):
+            evaluate_and_print_video_mme(
+                model=ctx.model,
+                processor=ctx.processor,
+                device=str(ctx.device),
+                batch_size=int(videomme.get("batch_size", 1)),
+                max_new_tokens=int(videomme.get("max_new_tokens", 30)),
+                n_samples=int(videomme.get("n_samples", -1)),
```
mhs4670go: `n_samples=-1` is used in the config as the full dataset sentinel, but this value is forwarded to `evaluate_vlm_on_tasks()` as `limit=-1`. Since `evaluate_vlm_on_tasks()` treats any non-None limit as partial Video-MME evaluation, this switches the task to videomme_mini and only downloads a limited set of chunks.

Is this what you intended?

tico/quantization/wrapq/examples/quantize_qwen3_vl_with_gptq.py
```diff
@@ -489,28 +489,28 @@ def parse_args():
...
+        "E.g. --videomme_n_samples 10 evaluates only the first 10 samples.",
     )
     parser.add_argument(
-        "--video_mme_max_num_frames",
+        "--videomme_max_num_frames",
         type=int,
-        default=5,
+        default=32,
         help="Maximum number of frames to sample from each video for Video-MME evaluation. "
         "Default is 5 for fitting within max_position_embeddings (2048).",
```
mhs4670go: Just curiosity, why can we change the default value if the initial 5 was for fitting within 2048?
```suggestion
        "Default is 32 for fitting within max_position_embeddings (2048).",
```

tico/quantization/recipes/adapters/qwen3_vl.py
```diff
@@ -421,6 +422,20 @@ def evaluate(self, ctx: RecipeContext) -> None:
...
+                max_num_frames=int(videomme.get("max_num_frames", -1)),
```
mhs4670go: The adapter defaults `max_num_frames` to -1, while the Video-MME helper defaults it to 32. If a config only sets `videomme.enabled=true`, the budget computation can receive `max_num_frames=-1`, which produces nonsensical adjusted frame counts; 0 would also raise a division error. Please default to 32 in the adapter and validate `max_num_frames > 0` before budget computation.

```python
videomme = eval_cfg.get("videomme", {})
if videomme.get("enabled", False):
    max_num_frames = int(videomme.get("max_num_frames", 32))
    if max_num_frames <= 0:
        raise ValueError(
            "evaluation.videomme.max_num_frames must be a positive integer."
        )

    evaluate_and_print_video_mme(
        model=ctx.model,
        processor=ctx.processor,
        device=str(ctx.device),
        batch_size=int(videomme.get("batch_size", 1)),
        max_new_tokens=int(videomme.get("max_new_tokens", 30)),
        n_samples=int(videomme.get("n_samples", -1)),
        max_num_frames=max_num_frames,
        use_cache=videomme.get("use_cache", None),
        verbose=bool(videomme.get("verbose", eval_cfg.get("verbose", False))),
    )
```

tico/quantization/recipes/adapters/qwen3_vl.py
Torrero: Thank you for pointing out this case, I fixed it,

tico/quantization/wrapq/examples/quantize_qwen3_vl_with_gptq.py
Torrero: > Just curiosity, why can we change the default value if the initial 5 was for fitting within 2048?

The estimated score of the float model is about 50%, I think 5 frames per video is not enough to correctly evaluate the dataset, it becomes something like guessing the correct answer for both the quantized and float models. Thus, I introduced a decrease in the number of pixels per frame with increasing the maximum number of frames.

tico/quantization/recipes/adapters/qwen3_vl.py
Torrero: fixed it

tico/quantization/evaluation/lmms_eval_utils.py
```diff
@@ -556,8 +633,28 @@ def evaluate_vlm_on_tasks(
...
-        model_name, model_args_dict = _build_model_args(inner_model, max_num_frames)
+
+        # Detect if the model is a PTQ wrapper with a static context budget.
+        # Quantized models have a static causal mask of size
+        # max_position_embeddings that cannot be exceeded at runtime.
+        # When evaluating video benchmarks, the total token count
+        # (text + visual tokens from all frames) must fit within this
+        # budget, so we compute max_pixels/min_pixels per frame.
+        _max_pos_emb = _get_max_position_embeddings(inner_model)
+        _effective_max_new_tokens = max_new_tokens if max_new_tokens is not None else 30
```
mhs4670go, line 644: There is still a mismatch between the token budget computation and the actual generation settings when `max_new_tokens` is left as `None`.

In `evaluate_vlm_on_tasks()`, the budget calculation falls back to `30` tokens:

```python
_effective_max_new_tokens = max_new_tokens if max_new_tokens is not None else 30
````

However, `gen_kwargs` is only passed to `simple_evaluate()` when `max_new_tokens is not None`. So in the default CLI path, the video frame budget is computed as if generation reserves 30 tokens, but lmms-eval's Qwen3_VL wrapper can still use its own default generation length. This means the computed `max_pixels` may be too large for the actual runtime context budget.

Could we make the default explicit and keep both paths consistent? For example:

```python
_effective_max_new_tokens = max_new_tokens if max_new_tokens is not None else 30
...
eval_kwargs["gen_kwargs"] = f"max_new_tokens={_effective_max_new_tokens}"
```

Torrero: fixed it, thank you for pointing this out

Torrero:
@mhs4670go PTAL

mhs4670go (approved):
LGTM

PR #766, opened 2026-06-09
[serialize] Implement RMSNorm

dahlinPL:
This commit implements RMSNorm

TICO-DCO-1.0-Signed-off-by: Marcin Słowiński <m.slowinski2@samsung.com>

### DIFF
--- a/tico/serialize/operators/op_rmsnorm.py
+++ b/tico/serialize/operators/op_rmsnorm.py
@@ -63,7 +63,7 @@ class RMSNormVisitor(NodeVisitor):
             return args.input, args.weight, eps
 
         circle_args = CircleRMSNormArgs(*node.args, **node.kwargs)
-        return circle_args.input, args.weight, args.eps
+        return circle_args.input, circle_args.weight, circle_args.eps
 
     def define_node(
         self,
--- a/tico/serialize/operators/op_rmsnorm.py
+++ b/tico/serialize/operators/op_rmsnorm.py
@@ -21,29 +21,55 @@ import torch
 from circle_schema import circle
 
 from tico.serialize.circle_graph import CircleSubgraph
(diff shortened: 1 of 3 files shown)


dahlinPL:
tested with below script:

```python
import torch
import torch.nn as nn

import tico


class RMSNormModel(nn.Module):
    """Simple model with RMSNorm for LLaMA/Qwen-style normalization."""

    def __init__(self):
        super().__init__()
        self.rms_norm = nn.RMSNorm(normalized_shape=64, eps=1e-6)

    def forward(self, x):
        return self.rms_norm(x)


if __name__ == "__main__":
    torch_module = RMSNormModel()
    example_inputs = (torch.randn(1, 8, 64, dtype=torch.float32),)

    circle_model = tico.convert(torch_module.eval(), example_inputs)
    circle_model.save("rmsnorm.circle")

    print("Circle model saved to: rmsnorm.circle")
```

result from main:
image


result from this PR:
image

tico/serialize/operators/op_rmsnorm.py
```diff
@@ -21,29 +21,55 @@
...
+                )
+
+            eps = args.eps
+            if eps is None:
+                raise NotYetSupportedError("RMSNorm eps=None is not supported yet")
+
+            return args.input, args.weight, eps
+
+        circle_args = CircleRMSNormArgs(*node.args, **node.kwargs)
+        return circle_args.input, args.weight, args.eps
```
mhs4670go:
```suggestion
        return circle_args.input, circle_args.weight, circle_args.eps
```

dahlinPL: omg 🤦

mhs4670go (approved):
LGTM

PR #770, opened 2026-06-10
[utils] Suppress "isinstance(treespec, LeafSpec) is deprecated" warning

arkq:
This PR suppresses warning caused by PyTorch:

```
/usr/lib/python3.14/copyreg.py:104: FutureWarning: `isinstance(treespec, LeafSpec)` is deprecated, use `isinstance(treespec, TreeSpec) and treespec.is_leaf()` instead.
  return cls.__new__(cls, *args)
```

### DIFF
--- a/tico/utils/convert.py
+++ b/tico/utils/convert.py
@@ -143,7 +143,7 @@ def traced_run_decompositions(exported_program: ExportedProgram):
             if op in _decomp_table:
                 del _decomp_table[op]
 
-        with SuppressWarning(FutureWarning, ".*LeafSpec*"):
+        with SuppressWarning(FutureWarning, ".*LeafSpec"):
             ep = ep.run_decompositions(decomp_table=_decomp_table)
         return ep
 
@@ -361,7 +361,7 @@ def convert(
             mod, args, kwargs, dynamic_shapes=dynamic_shapes, strict=strict
         )
 
-    with SuppressWarning(FutureWarning, ".*LeafSpec*"):
+    with SuppressWarning(FutureWarning, ".*LeafSpec"):
         circle_binary = convert_exported_module_to_circle(
             exported_program, config=config
         )
--- a/tico/utils/convert.py
+++ b/tico/utils/convert.py
@@ -143,7 +143,8 @@ def traced_run_decompositions(exported_program: ExportedProgram):
             if op in _decomp_table:
                 del _decomp_table[op]
 
-        ep = ep.run_decompositions(decomp_table=_decomp_table)
+        with SuppressWarning(FutureWarning, ".*LeafSpec*"):
+            ep = ep.run_decompositions(decomp_table=_decomp_table)
         return ep
 
     if torch.__version__.startswith("2.5"):
--- a/tico/utils/utils.py
+++ b/tico/utils/utils.py
@@ -41,19 +41,18 @@ def get_fake_mode(exported_program: ExportedProgram):
 
 
 class SuppressWarning:
-    def __init__(self, warning_category: type[Warning], regex):
-        self.warning_category = warning_category
+    def __init__(self, category: type[Warning], regex):
+        self.ctx = warnings.catch_warnings()
+        self.category = category
         self.regex = regex
 
     def __enter__(self):
-        warnings.filterwarnings(
-            "ignore", category=self.warning_category, message=self.regex
-        )
+        self.ctx.__enter__()
+        warnings.filterwarnings("ignore", self.regex, category=self.category)
+        return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
-        warnings.filterwarnings(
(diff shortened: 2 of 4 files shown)


tico/utils/convert.py
```diff
@@ -143,7 +143,8 @@ def run_decompositions(ep: ExportedProgram):
             if op in _decomp_table:
                 del _decomp_table[op]
 
-        ep = ep.run_decompositions(decomp_table=_decomp_table)
+        with SuppressWarning(FutureWarning, ".*LeafSpec*"):
+            ep = ep.run_decompositions(decomp_table=_decomp_table)
```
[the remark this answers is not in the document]
arkq: > message is a string containing a regular expression that the start of the warning message must match

`".*LeafSpec"` should be enough

mhs4670go:
@arkq 

One concern here is that `SuppressWarning.__exit__()` does not restore the previous warning filter state. Instead, it adds a new `"default"` filter for the same warning category/message.

https://github.com/Samsung/TICO/blob/2c4db21e05c24d205e97c9c37a842204b24e59e0/tico/utils/utils.py#L53-L56

This can be fragile when `SuppressWarning` contexts are nested. For example, `convert()` already wraps `convert_exported_module_to_circle()` with:

```python
with SuppressWarning(FutureWarning, ".*LeafSpec"):
    circle_binary = convert_exported_module_to_circle(
        exported_program, config=config
    )
```

This PR also adds another suppression inside `run_decompositions()`.

The inner `__exit__()` may re-enable the same warning while the outer suppression scope is still active. This makes the behavior dependent on warning filter ordering and can also leave global warning filters different from the state before entering the context.

How about making `SuppressWarning` restore the previous warning filter stack using `warnings.catch_warnings()`?

```python
class SuppressWarning:
    def __init__(self, warning_category: type[Warning], regex):
        self.warning_category = warning_category
        self.regex = regex
        self._catcher = None

    def __enter__(self):
        self._catcher = warnings.catch_warnings()
        self._catcher.__enter__()
        warnings.filterwarnings(
            "ignore",
            category=self.warning_category,
            message=self.regex,
        )
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        assert self._catcher is not None
        return self._catcher.__exit__(exc_type, exc_val, exc_tb)
```

This way, nested `SuppressWarning` scopes remain safe, and the warning filter state is restored exactly after leaving each context.

arkq:
> How about making SuppressWarning restore the previous warning filter stack using `warnings.catch_warnings()`

Yes, indeed nesting was broken in current implementation. I've fixed that with the suggested usage of `warnings.catch_warnings()`

tico/utils/utils.py
```diff
@@ -41,19 +41,18 @@ def get_fake_mode(exported_program: ExportedProgram):
 
 
 class SuppressWarning:
```
mhs4670go, line 43: Thanks for updating this to use `warnings.catch_warnings()`. This addresses the nested `SuppressWarning` concern.

One small nit: would it be slightly safer to create the `catch_warnings()` context in `__enter__()` instead of `__init__()`? With the current implementation, reusing the same `SuppressWarning` instance would try to re-enter the same `catch_warnings()` object. Most usages here instantiate it inline, so this is probably not an immediate issue, but creating the catcher in `__enter__()` makes the context manager more robust.

```python
class SuppressWarning:
    def __init__(self, category: type[Warning], regex):
        self.category = category
        self.regex = regex
        self.ctx = None

    def __enter__(self):
        self.ctx = warnings.catch_warnings()
        self.ctx.__enter__()
        warnings.filterwarnings("ignore", self.regex, category=self.category)
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        assert self.ctx is not None
        return self.ctx.__exit__(exc_type, exc_val, exc_tb)

Comment (arkq) [reply tico/utils/utils.py:43]: > With the current implementation, reusing the same SuppressWarning instance would try to re-enter the same catch_warnings() object.

Why would you like to reuse a context manager in Python? Technically it's possible, but reusing context manages in Python is not advisable. Most context manages are single-use only and using them twice might lead to resources leakage or a throw in the best case....

Just imagine what would happen here (the most common with usage in Python):

```python
f = open("example.txt", "w")
with f:
    f.write("one")
with f:
    f.write("two")
```

EDIT:

> reusing context manages in Python is not advisable.

Of course there is one exception to that rule:
```python
lock = threading.Lock()
with lock:
    print("Foo")
with lock:
    print("Bar")
```

mhs4670go: Fair point. Since all current usages instantiate `SuppressWarning` inline, reusing the same context manager instance is not a practical concern here.

My comment was only a robustness nit, not a blocker. I'm fine with keeping this as is. Thanks!

mhs4670go (approved):
LGTM

PR #774, opened 2026-06-11
[quantization] Add verbose and progress control for VQA

mhs4670go:
This commit adds verbose and progress control for VQA evaluation.


Torrero (approved):
LGTM

PR #775, opened 2026-06-12
[quantization] Fail fast on missing COCO eval dependencies

mhs4670go:
This commit fails fast on missing COCO eval dependencies
 before inference.


Torrero (approved):
LGTM

PR #776, opened 2026-06-12
[quantization] Fix gemma4_text_attention Circle export

mhs4670go:
- suppress_tico_warnings: yield inside try/except caught export exceptions and re-yielded, triggering "generator didn't stop after throw()". Separated import guard from yield to fix propagation.

- Gemma4TextAttentionBaseCase: shared_kv_states={} in export kwargs caused IndexError in torch.export strict mode (pytree tensor-count mismatch). Override export_input to strip it before export.


shs-park (commented):
LGTM!
=)

_Reviewed by AI via [oh-my-agents](https://github.com/shs-park/oh-my-agents)_

shs-park (approved):

PR #778, opened 2026-06-12
[quantization] Use circle_custom.rms_norm in QuantGemma4RMSNorm

mhs4670go:
This commit uses circle_custom.rms_norm in QuantGemma4RMSNorm.

### DIFF
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_rmsnorm.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_rmsnorm.py
@@ -39,30 +39,41 @@ class QuantGemma4RMSNorm(QuantModuleBase):
         self.eps = float(getattr(fp, "eps", getattr(fp, "variance_epsilon", 1e-6)))
         self.with_scale = hasattr(fp, "weight") and getattr(fp, "weight") is not None
 
+        self.obs_weight = self._make_obs("weight")
         self.obs_act_in = self._make_obs("act_in")
         self.obs_act_out = self._make_obs("act_out")
-        self.obs_weight = self._make_obs("weight") if self.with_scale else None
 
     def enable_calibration(self) -> None:
-        """Enable activation calibration and collect the static RMSNorm weight range."""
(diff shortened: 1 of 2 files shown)


tico/quantization/wrapq/wrappers/gemma4/quant_rmsnorm.py
```diff
@@ -39,30 +39,41 @@ def __init__(
...
+    def _raw_weight_for_rms_norm(self, x: torch.Tensor) -> torch.Tensor:
+        if self.with_scale:
+            return self.module.weight
+
+        return torch.ones((x.shape[-1],), dtype=x.dtype, device=x.device)
+
+    def _weight_for_rms_norm(self, x: torch.Tensor) -> torch.Tensor:
+        """Return a calibrated or fake-quantized RMSNorm scale tensor."""
+        weight = self._raw_weight_for_rms_norm(x)
+        if self._mode is Mode.QUANT:
```
shs-park, line 61: `with_scale=False` never collects stats for the synthetic unit `weight`, but quant mode still fake-quantizes it.

A scale-less RMSNorm can quantize the all-ones multiplier from empty observer stats and turn the normalized output into near-zero values instead of preserving unit scale.

```suggestion
        if self._mode is Mode.CALIB:
            self.obs_weight.collect(weight)
        elif self._mode is Mode.QUANT:
            weight = self.obs_weight.fake_quant(weight)
```

shs-park (commented):
AI found one or more actionable issues in this PR.
Please check the inline comments when present.

_Reviewed by AI via [oh-my-agents](https://github.com/shs-park/oh-my-agents)_

tico/quantization/wrapq/wrappers/gemma4/quant_rmsnorm.py
mhs4670go: If `with_scale` is False, `_raw_weight_for_rms_norm` returns `torch.ones`. Therefore, I think it's okay.

shs-park (approved):

PR #781, opened 2026-06-15
[quantization] Add a text decoder layer wrapper for gemma4

mhs4670go:
This commit adds a text decoder layer wrapper for gemma4.


shs-park (approved):
LGTM!
=)

PR #786, opened 2026-06-16
[serialize] Support Gather as Circle GatherNd

arkq:
Add support for PyTorch gather operator by converting it into Circle GATHER_ND.

Since the `torch.gather` and `circle.gather_nd` semantics are different, the conversion involves creating full indices matrix for `circle.gather_nd` based on the `index` and `dim` `torch.gather` arguments. The final conversion looks like this (for different gathering dimension in the torch semantic):

Gathering over 0 axis:
circle-0

Gathering over 1st axis:
circle-1

Gathering over 2nd axis:
circle-2

Gathering over 3rd axis:
circle-3

PR #790, opened 2026-06-22
[serialize] Support torch.gather export via Circle GatherNd

mhs4670go:
This commit supports torch.gather export via Circle GatherNd.

### DIFF
--- a/test/modules/op/gather.py
+++ b/test/modules/op/gather.py
@@ -1,0 +1,118 @@
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Modules that exercise torch.gather value tests through Circle export."""
+
+import torch
+
+from test.modules.base import TestModuleBase
+from test.support.tag import skip
+
+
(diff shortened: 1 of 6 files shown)


mhs4670go:
@arkq 

I agree with the overall lowering direction in #786: general `torch.gather` is not a good fit for Circle `GATHER`, and constructing full-coordinate indices for Circle `GATHER_ND` matches the expected semantics.

However, I am concerned about the PR's implementation strategy of keeping the original `aten.gather` node and changing its `index` input to a full-coordinate tensor, then relying on a metadata flag to make the serializer interpret it as `GATHER_ND`.

After this pass, the node **no longer has valid `aten.gather` semantics.** In other words, the graph still says `aten.gather(input, dim, index)`, but `index` is now shaped like `[..., rank]` and is meant to be consumed by `GATHER_ND`. This makes the intermediate IR misleading and invalid from a PyTorch/ATen point of view. It also means later passes, meta propagation, debugging, or future optimizations may accidentally treat this node as a normal `aten.gather`, even though its actual meaning has already changed.

I think it would be cleaner and safer to materialize an explicit internal GatherNd-like representation instead:

```
aten.gather(input, dim, index)
  -> build full-coordinate indices
  -> internal_gather_nd(input, full_indices)
  -> serialize internal_gather_nd as Circle GATHER_ND
```

This would make the semantic boundary clear:

- before the pass: the graph has PyTorch gather semantics
- after the pass: the graph has explicit Circle/TFLite-style GATHER_ND semantics
- the serializer no longer needs to infer a semantic change from a metadata flag

Because of this, I would prefer not to merge #786 in its current form. Instead, I suggest we keep the same high-level idea—lowering `torch.gather` to Circle `GATHER_ND`, which is why I posted this PR.

Could you review this PR instead?

arkq (approved):
> I am concerned about the PR's implementation strategy of keeping the original aten.gather node and changing its index input to a full-coordinate tensor, then relying on a metadata flag to make the serializer interpret it as GATHER_ND.
> After this pass, the node no longer has valid aten.gather semantics.

Yes, I totally agree with you. I did not know how to implement `torch.gather` to `circle.gather_nd` conversion while preserving torch graph validity - I was not aware that there is `torch.ops.circle_custom.gather_nd`. Your approach seems to be better indeed. The code looks OK.

PR #791, opened 2026-06-22
[quantization] Add a gemma wrapper for TextModel

mhs4670go:
This commit adds a wrapper for gemma text model.

```bash
python -m tico.quantization.examples.inspect \
  --config tico/quantization/examples/configs/wrapper_smoke.yaml \
  --mode wrapper-smoke \
  --case gemma4_text_model \
  --strict
┌───────────── Wrapper Smoke Summary ─────────────
│ Case             : gemma4_text_model
│ Status           : PASS
│ Mean |diff|      : 0.079053
│ Max |diff|       : 0.483883
│ PEIR             : 0.083751
│ Shape match      : True
│ Quant finite     : True
└─────────────────────────────────────────────────
    ┌────────────────────────────────────────────┐
 3.7┤                                            │
    │                                            │
    │                                     •   •  │
 2.6┤                                            │
    │                                 •          │
    │                              •  ••         │
    │                             ••••           │
 1.6┤                            •••             │
    │                        •••••               │
    │                      •••••                 │
 0.5┤                    •••••                   │
    │                   •••••                    │
    │                ••••••                      │
    │              •••••                         │
-0.5┤            ••••••                          │
    │          •••••                             │
    │         ••••                               │
-1.6┤      ••••                                  │
    │    ••••••                                  │
    │   •••                                      │
    │  •••                                       │
-2.7┤                                            │
    └┬──────────┬──────────┬─────────┬──────────┬┘
   -2.7       -1.1        0.5       2.1       3.7 
```

### DIFF
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
@@ -22,6 +22,7 @@ import torch
 import torch.nn as nn
 
 from tico.quantization.config.ptq import PTQConfig
+from tico.quantization.wrapq.mode import Mode
 from tico.quantization.wrapq.utils.utils import get_model_arg, join_name
 from tico.quantization.wrapq.wrappers.gemma4.utils import assert_gemma4_e2b_no_moe
 from tico.quantization.wrapq.wrappers.ptq_wrapper import PTQWrapper
@@ -600,14 +601,14 @@ class QuantGemma4TextModel(QuantModuleBase):
     ) -> torch.Tensor:
         """Recover input ids from exact main embeddings for PLE compatibility."""
         embedding = self._unwrap_fp_module(self.embed_tokens)
-        weight = embedding.weight.to(
-            device=inputs_embeds.device, dtype=inputs_embeds.dtype
-        )
+        weight = embedding.weight
         scale = getattr(embedding, "embed_scale", None)
         if scale is not None:
-            weight = weight * scale.to(
-                device=inputs_embeds.device, dtype=inputs_embeds.dtype
-            )
+            weight = weight * scale.to(device=weight.device, dtype=weight.dtype)
+        weight = weight.to(
+            device=inputs_embeds.device,
+            dtype=inputs_embeds.dtype,
+        )
 
         with torch.no_grad():
             matches = (inputs_embeds[:, :, None, :] == weight[None, None, :, :]).all(
@@ -722,6 +723,20 @@ class QuantGemma4TextModel(QuantModuleBase):
             )
         if per_layer_inputs is not None and not self.hidden_size_per_layer_input:
             raise ValueError("per_layer_inputs was provided, but PLE is disabled.")
+        # In QUANT mode, inputs_embeds may already be fake-quantized before PLE
+        # construction. Recovering token ids by exact floating-point comparison
+        # against the raw embedding table is therefore not a valid quantized-path
+        # contract. Require callers to pass explicit per_layer_inputs instead.
+        if (
+            self._mode is Mode.QUANT
+            and input_ids is None
+            and self.hidden_size_per_layer_input
+            and per_layer_inputs is None
+        ):
+            raise ValueError(
+                "Gemma4 PLE requires explicit per_layer_inputs when "
+                "inputs_embeds is provided in QUANT mode."
+            )
 
         return_shared_kv_states = bool(kwargs.pop("return_shared_kv_states", False))
         output_hidden_states = bool(output_hidden_states)
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
@@ -12,14 +12,17 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+"""PTQ wrapper for the dense Gemma4 E2B text model."""
+
 from collections import UserDict
-from typing import Iterable, Optional
+from collections.abc import Mapping
+from typing import Any, Iterable, Optional
 
 import torch
 import torch.nn as nn
 
 from tico.quantization.config.ptq import PTQConfig
-from tico.quantization.wrapq.utils.utils import join_name
+from tico.quantization.wrapq.utils.utils import get_model_arg, join_name
 from tico.quantization.wrapq.wrappers.gemma4.utils import assert_gemma4_e2b_no_moe
 from tico.quantization.wrapq.wrappers.ptq_wrapper import PTQWrapper
 from tico.quantization.wrapq.wrappers.quant_module_base import QuantModuleBase
@@ -28,7 +31,26 @@ from tico.quantization.wrapq.wrappers.registry import try_register
 
 @try_register("transformers.models.gemma4.modeling_gemma4.Gemma4TextModel")
 class QuantGemma4TextModel(QuantModuleBase):
-    """PTQ wrapper skeleton for the Gemma4 E2B text model."""
+    """Quantization wrapper for the dense Gemma4 E2B text model.
+
+    Full and sliding causal masks are built from bounded static templates owned
+    by this wrapper. The forward path does not call Hugging Face mask factories.
+    Masked entries use ``PTQConfig.attention_mask_fill_value`` instead of the
+    minimum finite value of the activation dtype, which keeps affine observer
+    ranges usable around unmasked attention logits.
+
+    The static template capacity can be selected with
+    ``PTQConfig.model_args["text"]["max_seq"]``. When it is omitted, the
+    wrapper uses the smaller of the model context limit and 2048 tokens to avoid
+    allocating an unexpectedly large quadratic mask.
+
+    The E2B scope intentionally rejects MoE. Cache allocation and writes,
+    sampling, and CPU/NPU orchestration remain runtime responsibilities.
+    """
+
+    force_export: bool = False
+    _DEFAULT_STATIC_MAX_SEQ = 2048
+    _SUPPORTED_LAYER_TYPES = frozenset(("full_attention", "sliding_attention"))
 
     def __init__(
         self,
@@ -39,12 +61,21 @@ class QuantGemma4TextModel(QuantModuleBase):
     ):
         assert_gemma4_e2b_no_moe(fp_model)
         super().__init__(qcfg, fp_name=fp_name)
+
         self.module = fp_model
         self.config = fp_model.config
-        self.unique_layer_types = set(self.config.layer_types)
-        self.hidden_size_per_layer_input = getattr(
-            fp_model, "hidden_size_per_layer_input", None
+        self.padding_idx = getattr(fp_model, "padding_idx", None)
+        self.vocab_size = getattr(fp_model, "vocab_size", self.config.vocab_size)
+        self.unique_layer_types = tuple(sorted(set(self.config.layer_types)))
+        self.hidden_size_per_layer_input = int(
+            getattr(fp_model, "hidden_size_per_layer_input", 0) or 0
         )
+
+        unsupported = set(self.unique_layer_types) - self._SUPPORTED_LAYER_TYPES
+        if unsupported:
+            raise ValueError(
+                "Unsupported Gemma4 text layer types: " f"{sorted(unsupported)}."
+            )
 
         self.embed_tokens = PTQWrapper(
             fp_model.embed_tokens,
@@ -67,6 +98,9 @@ class QuantGemma4TextModel(QuantModuleBase):
             fp_name=join_name(fp_name, "norm"),
         )
         self.rotary_emb = fp_model.rotary_emb
+        self.gradient_checkpointing = bool(
+            getattr(fp_model, "gradient_checkpointing", False)
+        )
 
         self.embed_tokens_per_layer: Optional[nn.Module] = None
         self.per_layer_model_projection: Optional[nn.Module] = None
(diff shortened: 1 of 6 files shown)


dvsav:
# Note for myself

## What

This commit adds a full PTQ wrapper (`QuantGemma4TextModel`) for the dense Gemma4 E2B text model, enabling quantization of the complete `Gemma4TextModel` architecture. The wrapper replaces a previous stub/skeleton implementation with a production-ready version that constructs bounded static attention masks and RoPE templates internally (without calling HuggingFace mask factories), supports Gemma4-specific Per-Layer Embeddings (PLE), shared KV states, and is compatible with the `torch.export`-based conversion pipeline. The wrapper is also activated in the module registry.

## Why

The Gemma4 model family requires a text-model-level wrapper to enable end-to-end PTQ quantization and conversion to Circle format. Previously, only the sub-component wrappers (attention, MLP, decoder layer) were registered, meaning the full text model could not be prepared as a single quantizable unit. This commit fills that gap by providing the top-level `QuantGemma4TextModel` wrapper that orchestrates the sub-layer wrappers and handles Gemma4-specific concerns:

- **Static mask construction**: NPU export requires precomputed bounded masks rather than dynamic HuggingFace mask factory calls. The wrapper builds full causal and sliding-window causal mask templates as registered buffers, sliced at runtime.
- **Finite mask fill values**: Using `PTQConfig.attention_mask_fill_value` (a finite negative number) instead of `-inf` keeps affine observer ranges usable during calibration.
- **Static RoPE templates**: Precomputed cos/sin tables per layer type avoid dynamic RoPE computation during `torch.export` tracing.
- **PLE support**: Gemma4's Per-Layer Embeddings mechanism (token-identity + context-aware projection) needs special handling in the quantized path, including reverse input-id lookup from embeddings for multimodal use cases.

## Key Design Decisions

1. **Bounded static mask templates over HF mask factories**: The wrapper pre-builds full causal and sliding-window causal masks as buffers during `__init__`, then slices them at runtime. This eliminates the dependency on HuggingFace's `create_causal_mask` / `create_sliding_window_causal_mask` during the forward pass, which is essential for `torch.export` compatibility and NPU deployment.

2. **Finite fill value instead of `-inf`**: Masked positions use `PTQConfig.attention_mask_fill_value` (a large finite negative number) rather than the minimum finite value of the activation dtype. This prevents affine observers from seeing extreme outlier values that would distort quantization parameter computation.

3. **Static RoPE templates with dynamic fallback**: Position embeddings are precomputed and registered as buffers for the export/static path, but the wrapper falls back to calling `self.rotary_emb()` dynamically when not in export context (and position_ids were explicitly provided). This balances export compatibility with calibration flexibility.

4. **No `as_export_module` on TextModel**: The complete `QuantGemma4TextModel` is intentionally not the static NPU export unit. Export happens at the decoder-layer granularity. The wrapper explicitly raises `NotImplementedError` if static inputs are required without a pre-provided mask mapping.

5. **MoE rejection**: The E2B scope explicitly rejects Mixture-of-Experts configs via `assert_gemma4_e2b_no_moe`, keeping the wrapper focused on dense decoder layers only.

6. **Reverse input-id lookup for PLE**: When only `inputs_embeds` is provided (typical in multimodal VLM pipelines), the wrapper can reverse-lookup `input_ids` from the embedding matrix to compute per-layer token-identity embeddings. This avoids requiring callers to provide both.

7. **`force_export` flag + `torch.compiler.is_compiling()` detection**: The wrapper detects whether it's running inside a `torch.export` context and automatically switches to static template paths. The `force_export` flag allows explicit override.

## Changes

- **`tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py`** - Major expansion from a stub to a full 848-line implementation of `QuantGemma4TextModel`. Key additions include:
  - Bounded static mask template construction (`_build_full_attention_mask_template`, `_build_sliding_attention_mask_template`) and registration as buffers
  - Static RoPE cos/sin template precomputation and registration
  - Position ID creation/validation with static template support (`_make_position_ids`)
  - Attention mask normalization and construction from static templates (`_normalize_attention_mask_for_layer`, `_create_attention_mask_mapping`, `_observe_attention_mask_mapping`)
  - Per-Layer Embeddings (PLE) support: `get_per_layer_inputs`, `project_per_layer_inputs`, `_reverse_input_ids_from_embeddings`
  - Full forward method with DynamicCache support, shared KV states, `return_dict`/`output_hidden_states`/`output_attentions`/`return_shared_kv_states` output options
  - Comprehensive observer collection (`_all_observers`) including attention masks, position embeddings, and PLE observers
  - Input validation for invalid argument combinations
  - `_requires_static_inputs()` detection for torch.export context
  - `_unwrap_layer_output` helper for decoder layer output extraction
  - `_output_cls()` returning `Gemma4TextModelOutputWithPast` for HF-compatible output

- **`test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py`** - New 372-line test file with 9 unit tests covering:
  - Registry-based prepare wrapping
  - No-quant forward parity with HuggingFace (input_ids path)
  - Static CPU-provided mask mapping path
  - PLE path with input_ids
  - PLE path with inputs_embeds and explicit per_layer_inputs (multimodal-style)
  - Shared KV states return
  - Input validation error messages
  - Absence of `as_export_module` on TextModel
  - Mode transitions and observer collection lifecycle
  - MoE config rejection

- **`tico/quantization/recipes/debug/wrapper_smoke/cases/gemma4.py`** - Added `Gemma4TextModelCase` smoke case class and included it in the `GEMMA4_CASES` tuple. The case builds a tiny Gemma4TextModel with sliding+full attention, provides calibration/eval inputs with `input_ids` and `attention_mask`, and uses `return_dict=True`.

- **`tico/quantization/wrapq/wrappers/registry.py`** - Uncommented the `quant_text_model` module line in `_CORE_MODULES`, activating the `QuantGemma4TextModel` wrapper in the registry so it is discovered and used by `prepare()`.

## Tests

The changes are tested through two complementary mechanisms:

1. **Unit tests** (`test_quant_text_model.py`): 9 test cases covering the core functionality:
   - Registry integration: verifies `prepare()` produces a `PTQWrapper` containing `QuantGemma4TextModel`
   - Numerical parity: no-quant wrapper output matches HuggingFace reference for both `input_ids` and `inputs_embeds` paths (using `torch.allclose` with `atol=1e-5, rtol=1e-5`)
   - Static mask mapping: verifies the CPU-provided mask dict path matches HF output
   - PLE paths: both token-identity (input_ids) and multimodal-style (inputs_embeds + explicit per_layer_inputs)
   - Shared KV states: verifies output contains shared KV states when requested
   - Validation: checks error messages for invalid input combinations (both input_ids and inputs_embeds, per_layer_inputs with input_ids, static export without mask dict)
   - Export contract: confirms TextModel does not own `as_export_module`
   - Lifecycle: calibration → freeze → quantization mode transitions with observer verification
   - Scope: MoE configs are explicitly rejected

   All tests are gated by `@unittest.skipUnless(_has_gemma4(), ...)` to skip gracefully when transformers doesn't provide Gemma4.

```
$ python -m pytest test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py -v
============================================================================== test session starts ===============================================================================
cachedir: .pytest_cache
configfile: pyproject.toml
plugins: anyio-4.13.0
collected 10 items                                                                                                                                                               

test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_00_prepare_wraps_text_model_when_registered PASSED                        [ 10%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_mode_transitions_and_observer_collection PASSED                           [ 20%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_moe_text_model_is_rejected_for_e2b_scope PASSED                           [ 30%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_no_quant_forward_matches_hf_text_model_with_input_ids PASSED              [ 40%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_ple_path_matches_hf_with_input_ids PASSED                                 [ 50%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_ple_path_matches_hf_with_inputs_embeds_and_explicit_per_layer_inputs PASSED [ 60%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_shared_kv_text_model_returns_shared_state_when_requested PASSED           [ 70%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_static_attention_mask_mapping_matches_hf_text_model PASSED                [ 80%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_text_model_wrapper_does_not_own_export_adapter_hook PASSED                [ 90%]
test/quantization/wrapq/wrappers/gemma4/test_quant_text_model.py::TestQuantGemma4TextModel::test_validation_errors_match_expected_contract PASSED                          [100%]

============================================================================== 10 passed in 11.32s ===============================================================================

```

2. **Smoke test case** (`gemma4.py`): `Gemma4TextDecoderLayerSlidingPrefillCase` added to the `GEMMA4_CASES` tuple for integration-level wrapper smoke testing.

```
$ python -m tico.quantization.examples.inspect \
    --config tico/quantization/examples/configs/wrapper_smoke.yaml \
    --mode wrapper-smoke \
    --case gemma4_text_decoder_layer_sliding_prefill \
    --export circle \
    --output-dir ./out/wrapper_smoke
┌───────────── Wrapper Smoke Summary ─────────────
│ Case             : gemma4_text_decoder_layer_sliding_prefill
│ Status           : PASS
│ Mean |diff|      : 0.116630
│ Max |diff|       : 1.095015
│ PEIR             : 0.118720
│ Shape match      : True
│ Quant finite     : True
└─────────────────────────────────────────────────
Artifacts:
  - circle: out/wrapper_smoke/gemma4_text_decoder_layer_sliding_prefill.q.circle
    ┌────────────────────────────────────────────┐
 4.9┤                                            │
    │                                       ••   │
    │                                      •• •  │
 3.2┤                                    •••     │
    │                                  ••••      │
    │                               ••••         │
    │                             ••••           │
 1.5┤                           ••••             │
    │                        •••••••             │
    │                      •••••                 │
-0.2┤                     ••••                   │
    │                 •••••••                    │
    │                 ••••                       │
    │             ••••••                         │
-1.9┤            •••••                           │
    │        •  ••••                             │
    │         •••                                │
-3.6┤       •••                                  │
    │     •••                                    │
    │     ••                                     │
    │  ••                                        │
-5.3┤                                            │
    └┬──────────┬──────────┬─────────┬──────────┬┘
   -5.3       -2.7       -0.2       2.4       4.9 
```

Torrero (dismissed):
LGTM

tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
```diff
@@ -89,99 +123,726 @@ def __init__(
...
+        scale = getattr(embedding, "embed_scale", None)
+        if scale is not None:
+            weight = weight * scale.to(
+                device=inputs_embeds.device, dtype=inputs_embeds.dtype
+            )
+
+        with torch.no_grad():
+            matches = (inputs_embeds[:, :, None, :] == weight[None, None, :, :]).all(
+                dim=-1
+            )
```
dvsav, line 616: I suspect that the exact equality comparison `inputs_embeds[...] == weight[...]` may fail:

1. __In QUANT mode__, `inputs_embeds` is fake-quantized before the reverse lookup is called, so it no longer exactly matches any row in the raw weight table.
2. __With dtype casting__, floating-point non-associativity breaks the comparison: `(weight * scale).to(fp16) != weight.to(fp16) * scale.to(fp16)`, because the multiplication happens in different precision contexts on each side.

dvsav:
🤔 

## Gemma4TextModelCase

```
$ python -m tico.quantization.examples.inspect \
    --config tico/quantization/examples/configs/wrapper_smoke.yaml \
    --mode wrapper-smoke \
    --case gemma4_text_model \
    --export circle \
    --output-dir ./out/wrapper_smoke
┌───────────── Wrapper Smoke Summary ─────────────
│ Case             : gemma4_text_model
│ Status           : FAIL
│ Mean |diff|      : 0.079053
│ Max |diff|       : 0.483881
│ PEIR             : 0.083751
│ Shape match      : True
│ Quant finite     : True
└─────────────────────────────────────────────────
Messages:
  - Circle export failed: Observed exception
  Explanation: Dynamo found no exception handler at the top-level compiled function when encountering an exception. Exception will propagate outside the compiled region.
  Hint: Your code may result in an error when running in eager. Please double check that your code doesn't contain a similar error when actually running eager/uncompiled. You can do this by removing the `torch.compile` call, or by using `torch.compiler.set_stance("force_eager")`. 
  Hint: It may be possible to write Dynamo tracing rules for this code. Please report an issue to PyTorch if you encounter this graph break often and it is causing performance issues.

  Developer debug context: raised exception NotImplementedError([ConstantVariable(str: 'QuantGemma4TextModel static/export mode requires static masks as a dict keyed by layer type.')])

 For more details about this graph break, please visit: https://meta-pytorch.github.io/compile-graph-break-site/gb/gb0088.html

from user code:
   File "myenv/lib/python3.10/site-packages/torch/_dynamo/functional_export.py", line 221, in forward
    res = self._export_root(*args, **kwargs)
  File "TICO/tico/quantization/wrapq/wrappers/ptq_wrapper.py", line 46, in forward
    return self.wrapped(*args, **kwargs)
  File "TICO/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py", line 761, in forward
    mask_mapping = self._create_attention_mask_mapping(
  File "TICO/tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py", line 532, in _create_attention_mask_mapping
    raise NotImplementedError(

Set TORCHDYNAMO_VERBOSE=1 for the internal stack trace (please do this especially if you're reporting a bug to PyTorch). For even more developer context, set TORCH_LOGS="+dynamo"

    ┌────────────────────────────────────────────┐
 3.7┤                                            │
    │                                         •  │
 2.6┤                                     •      │
    │                                 ••         │
    │                             •••••          │
 1.6┤                            •••             │
    │                        •••••               │
 0.5┤                    ••••••                  │
    │                  ••••••                    │
    │              •••••• •                      │
-0.5┤            ••••••                          │
    │          •••••                             │
-1.6┤      ••••••                                │
    │    ••••••                                  │
    │  •••                                       │
-2.7┤                                            │
    └┬──────────┬──────────┬─────────┬──────────┬┘
   -2.7       -1.1        0.5       2.1       3.7
```

dvsav (changes requested):

tico/quantization/wrapq/wrappers/gemma4/quant_text_model.py
mhs4670go: Good catch. The reverse lookup is inherently a floating-point exact-match operation against the raw embedding table as you said, so it does not fit well with the quantization path.

I changed the QUANT path so that, when PLE is enabled and `inputs_embeds` is provided, callers must also provide explicit `per_layer_inputs`. This prevents the reverse lookup from running on fake-quantized embeddings.

mhs4670go:
@dvsav 

The export failure is expected because `Gemma4TextModelCase` currently validates only PTQ preparation, calibration, conversion, and numerical parity. Exporting the full text model is not considered here. It can be re-consdiered later we design a runtime.

To make this limitation clear, I explicitly marked Circle export as unsupported for this case so that an export request fails early with a descriptive message, rather than surfacing as an unexpected conversion error.

```bash
python -m tico.quantization.examples.inspect \
    --config tico/quantization/examples/configs/wrapper_smoke.yaml \
    --mode wrapper-smoke \
    --case gemma4_text_model \
    --export circle \
    --output-dir ./out/wrapper_smoke
┌───────────── Wrapper Smoke Summary ─────────────
│ Case             : gemma4_text_model
│ Status           : FAIL
│ Mean |diff|      : 0.079053
│ Max |diff|       : 0.483883
│ PEIR             : 0.083751
│ Shape match      : True
│ Quant finite     : True
└─────────────────────────────────────────────────
Messages:
  - This case validates PTQ numerical parity only. Full Gemma4TextModel Circle export requires a dedicated static adapter.
    ┌────────────────────────────────────────────┐
 3.7┤                                            │
    │                                            │
    │                                     •   •  │
 2.6┤                                            │
    │                                 •          │
    │                              •  ••         │
    │                             ••••           │
 1.6┤                            •••             │
    │                        •••••               │
    │                      •••••                 │
 0.5┤                    •••••                   │
    │                   •••••                    │
    │                ••••••                      │
    │              •••••                         │
-0.5┤            ••••••                          │
    │          •••••                             │
    │         ••••                               │
-1.6┤      ••••                                  │
    │    ••••••                                  │
    │   •••                                      │
    │  •••                                       │
-2.7┤                                            │
    └┬──────────┬──────────┬─────────┬──────────┬┘
   -2.7       -1.1        0.5       2.1       3.7
```

dvsav (approved):
👍 LGTM. Thank you!

Torrero (approved):
LGTM

PR #807, opened 2026-07-06
[quantization] Align Gemma4 PTQ builder with HF module scope

mhs4670go:
Previously, `build_gemma4_e2b_ptq_config()` always placed text overrides under
`model.language_model.*`, which matches `Gemma4ForConditionalGeneration` but not
`Gemma4ForCausalLM`. In Hugging Face `Gemma4ForCausalLM`, model is already a
`Gemma4TextModel`, so overrides such as embed_tokens, layers, and norm were not
reached when preparing the causal LM wrapper.

Update the Gemma4 PTQ override tree to follow the original HF module layout for both `CausalLM` and `ConditionalGeneration` paths, so the existing builder API can be reused without silently dropping custom text-model overrides.

### DIFF
--- a/tico/quantization/config/gemma4_builders.py
+++ b/tico/quantization/config/gemma4_builders.py
@@ -60,6 +60,17 @@ def _linear_tree_override(
     return root
 
 
+def _deep_merge(base: Mapping[str, Any], overlay: Mapping[str, Any]) -> Dict[str, Any]:
+    """Return ``base`` recursively merged with ``overlay``."""
+    out: Dict[str, Any] = copy.deepcopy(dict(base))
+    for key, value in overlay.items():
+        if isinstance(value, Mapping) and isinstance(out.get(key), Mapping):
+            out[key] = _deep_merge(out[key], value)
+        else:
+            out[key] = copy.deepcopy(value)
+    return out
+
+
 def _gemma4_text_attention_override(
     linear_weight: Optional[QuantSpec],
     norm_weight: Optional[QuantSpec],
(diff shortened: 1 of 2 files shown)


tico/quantization/config/gemma4_builders.py
```diff
@@ -85,46 +228,51 @@ def build_gemma4_e2b_ptq_config(
...
 
-    model_overrides["embed_vision"] = {
-        "embedding_projection": _weight_override(linear_weight),
+    overrides: Dict[str, Any] = {
+        **copy.deepcopy(text_model_overrides),
+        **copy.deepcopy(gemma4_model_overrides),
+        "model": {
+            **copy.deepcopy(text_model_overrides),
+            **copy.deepcopy(gemma4_model_overrides),
+        },
```
dvsav: The `**` spread does a *shallow merge* at the top level. If `text_model_overrides` and `gemma4_model_overrides` share any top-level keys, the second one silently overwrites the first.
Quick example:
```
>>> a = { 1: "1", 2: "2" }
>>> b = { 1: "one" }
>>> c = { **a, **b }
>>> c
{1: 'one', 2: '2'}
```
Currently `text_model_overrides` and `gemma4_model_overrides` don't overlap, but relying on that may be fragile.
How about applying `deep_merge` (`tico/quantization/recipes/config.py`)?
```
    overrides: Dict[str, Any] = {
        **deep_merge(text_model_overrides, gemma4_model_overrides),
        "model": deep_merge(text_model_overrides, gemma4_model_overrides),
```

tico/quantization/config/gemma4_builders.py
```diff
@@ -60,6 +60,149 @@ def _linear_tree_override(
...
+        vision["encoder"]["layers"][str(idx)] = {
+            "self_attn": _linear_tree_override(
+                linear_weight,
+                ("q_proj", "k_proj", "v_proj", "o_proj"),
+            ),
+            "mlp": _linear_tree_override(
+                linear_weight,
+                ("gate_proj", "up_proj", "down_proj"),
+            ),
+        }
```
dvsav: Vision encoder layers' overrides specify only `self_attn` and `mlp`.
But `Gemma4VisionEncoderLayer` and `QuantGemma4VisionEncoderLayer` also have:
- `input_layernorm` (RMSNorm)
- `post_attention_layernorm` (RMSNorm)
- `pre_feedforward_layernorm` (RMSNorm)
- `post_feedforward_layernorm` (RMSNorm)

These norm weights are not overridden. If `norm_weight` is provided in `build_gemma4_e2b_ptq_config`, the vision encoder norms will use the default weight spec instead of the specified `norm_weight`. This seems to be inconsistent with the text layer override, which does include all norm modules.

tico/quantization/config/gemma4_builders.py
```diff
@@ -60,6 +60,149 @@ def _linear_tree_override(
...
+            "input_proj": _weight_override(vision_patch_embed_weight or linear_weight)
+        },
+        "encoder": {"layers": {}},
+    }
+    for idx in range(num_vision_layers):
+        vision["encoder"]["layers"][str(idx)] = {
+            "self_attn": _linear_tree_override(
+                linear_weight,
+                ("q_proj", "k_proj", "v_proj", "o_proj"),
+            ),
```
dvsav: The text attention override includes `q_norm`, `k_norm`, `v_norm`:
```python
# Text attention
for name in ("q_norm", "k_norm", "v_norm"):
    result[name] = copy.deepcopy(norm_override)
```

But the vision attention override only includes the linear projections `("q_proj", "k_proj", "v_proj", "o_proj")`.
`QuantGemma4VisionAttention` also has `q_norm`, `k_norm` and `v_norm`, so would you consider adding the respective overrides?

dvsav (changes requested):

tico/quantization/config/gemma4_builders.py
```diff
@@ -85,46 +228,51 @@ def build_gemma4_e2b_ptq_config(
...
+        vision_patch_embed_weight=vision_patch_embed_weight,
+    )
 
-    model_overrides = overrides["model"]
-    model_overrides["language_model"] = {
-        "embed_tokens": _weight_override(embedding_weight),
-        "layers": {},
-        "norm": _weight_override(norm_weight),
+    multimodal_projection_overrides = {
+        "embedding_projection": _weight_override(linear_weight),
```
dvsav: The multimodal embedder (`QuantGemma4MultimodalEmbedder`) has two submodules:
- `embedding_pre_projection_norm` (RMSNorm)
- `embedding_projection` (Linear)

The builder only overrides `embedding_projection`:
```python
multimodal_projection_overrides = {
    "embedding_projection": _weight_override(linear_weight),
}
```

mhs4670go:
@dvsav Thanks for the reviews. It was really helpful. PTAL:)

dvsav (approved):
LGTM, Thank you!

PR #808, opened 2026-07-06
[quantization] Add configurable Qwen3-VL text attention layout

mhs4670go:
Introduce Qwen3-VL attention options to switch between the NPU-export-friendly unrolled path and a HF-like batched path.


Torrero (approved):
LGTM

Issue #768, opened 2026-06-10
[quantization] Support Gemma4


mhs4670go:
## Goal

Add wrapper-based PTQ support for **Gemma4 E2B** and make it usable in a **static-shape NPU inference flow**. The runtime should follow the same high-level principle as the existing Llama static runtime: CPU owns dynamic/control-heavy orchestration, while NPU executes fixed-shape, quantized tensor-compute subgraphs.

This issue focuses on implementation work needed for:

- activation quantization and weight quantization
- static-shape prefill and decode inference
- image-text Gemma4 E2B first
- CPU/NPU partitioning
- wrapper smoke tests and evaluation flow
- later comparison against the existing Qwen3-VL PTQ path

### In scope for v0

- Gemma4 E2B only.
- Dense text decoder path only.
- No MoE path.
- Batch size fixed to `1` for the first static runtime.
- Static `max_seq`, for example `2048`.
- Static image input shape and static visual-token count.
- One fixed visual segment layout, for example:

  ```text
  [text prefix][image soft-token slots][text suffix][padding]
  ```

- Static prefill graph and static decode graph separated.
- Qwen3-VL style evaluation compatibility.

## Key design decision

Do **not** try to export the full Hugging Face Gemma4 forward as one NPU graph.

Instead, follow the Llama static runtime pattern:

1. prepare/wrap model modules for PTQ
2. expose static export adapters for subgraphs
3. let CPU runtime manage tokenization, layout validation, RoPE, masks, cache state, sampling, and loop control
4. let NPU run only fixed-shape tensor compute with quantized weights and quantized activations

This is necessary because Gemma4 full forward contains dynamic/control-heavy behavior:

- placeholder mask construction
- image/video/audio `masked_scatter` fusion
- dynamic mask creation through Hugging Face mask utilities
- optional cache object handling
- shared KV state bookkeeping
- optional PLE path
- optional multimodal paths

These behaviors are better handled by a static runtime wrapper rather than exported as-is.

## Proposed directory structure

```text
tico/
└── quantization/
    ├── wrapq/
    │   ├── wrappers/
    │   │   ├── registry.py                         # MODIFY: register Gemma4 wrapper modules
    │   │   └── gemma4/                              # ADD
    │   │       ├── __init__.py
    │   │       ├── utils.py
    │   │       ├── export_adapters.py
    │   │       │
    │   │       ├── quant_clippable_linear.py
    │   │       ├── quant_rmsnorm.py
    │   │       │
[...]

PR #811, opened 2026-07-06
[evaluation] Wire LLaVA-Bench judge evaluation in Gemma4Adapter

dvsav:
## What

This PR wires the `Gemma4Adapter` into the TICO recipe pipeline and implements its `evaluate()` method to dispatch LLaVA-Bench judge evaluation. Previously, the `Gemma4Adapter` class existed but was not registered in the adapter registry, causing `KeyError: "Unknown model family: gemma4"` when running any Gemma4 config. Additionally, `evaluate()` raised `NotImplementedError`, blocking evaluation runs.

## Why

The [Support Gemma4](https://github.com/Samsung/TICO/issues/768) task specification requires Qwen3-VL style evaluation compatibility for Gemma4, including a LLaVA-Bench judge config. The `Gemma4Adapter` class and the `gemma4_e2b_llava_bench_judge.yaml` config already existed, but the adapter was not registered in `_ADAPTERS`, and `evaluate()` was a stub. This PR completes the wiring so that `python tico/quantization/examples/quantize.py --config .../gemma4_e2b_ptq_only.yaml` and `python tico/quantization/examples/evaluate.py --config .../gemma4_e2b_llava_bench_judge.yaml` both work.

## Key Design Decisions

1. **Direct copy of Qwen3-VL dispatch pattern** — The `evaluate()` implementation is a direct copy of `Qwen3VLAdapter.evaluate()`'s LLaVA-Bench dispatch block. The underlying `evaluate_and_print_llava_bench_judge()` function is model-agnostic (it uses `model.generate()` via the HF `GenerationMixin`), so no model-specific logic was needed.

2. **Three dispatch paths** — The `evaluate()` method supports three LLaVA-Bench modes: `judge` (LLM judge scoring), `legacy` (COCO-style CIDEr/BLEU), and boolean `true` (legacy with deprecation warning). Unknown modes raise `ValueError`. This matches the Qwen3-VL adapter exactly.

3. **Stub `export()` method** — The abstract base class `ModelAdapter` requires `export()`. A stub was added that checks `export.enabled` and returns early, raising `NotImplementedError` only if export is explicitly enabled. This allows PTQ-only and evaluation configs to run without implementing full Circle export.

4. **Image sizing fields in YAML** — Added `image_max_pixels`, `image_min_pixels`, `resized_height`, `resized_width`, `visual_token_margin` to `gemma4_e2b_llava_bench_judge.yaml` to match the Qwen3-VL config. These control image resizing before processor tokenization to fit the static token budget.

## Changes

- **`tico/quantization/recipes/adapters/__init__.py`** — Added `Gemma4Adapter` import and registered `"gemma4": Gemma4Adapter()` in the `_ADAPTERS` dict.
- **`tico/quantization/recipes/adapters/gemma4.py`** — Added imports for `evaluate_and_print_llava_bench_judge`, `evaluate_llava_bench`, `print_coco_score_results`; replaced the `NotImplementedError` stub in `evaluate()` with full LLaVA-Bench judge/legacy dispatch logic; added stub `export()` method.
- **`tico/quantization/examples/configs/gemma4_e2b_llava_bench_judge.yaml`** — Added `image_max_pixels`, `image_min_pixels`, `resized_height`, `resized_width`, `visual_token_margin` fields for token budget control.
- **`test/quantization/recipes/test_gemma4_adapter_llava_bench_judge.py`** (new) — 5 unit tests for the evaluate dispatch.

## Tests

All tests were run with `python -m pytest` from `test/quantization/recipes/`:

- `test_evaluate_dispatches_nested_llava_bench_judge_config` — Verifies that a nested `llava_bench` config with `mode: judge` calls `evaluate_and_print_llava_bench_judge` with the correct kwargs (model, processor, device, llava_cfg, model_cfg, runtime_cfg, default_n_samples, default_max_seq_len).
- `test_evaluate_dispatches_legacy_llava_bench_boolean_path` — Verifies that `llava_bench: true` (boolean) falls through to the legacy COCO path with a deprecation warning.
- `test_evaluate_dispatches_nested_legacy_mode` — Verifies that `mode: legacy` calls `evaluate_llava_bench` with per-config `n_samples` and `max_seq_len`.
- `test_evaluate_rejects_unknown_llava_bench_mode` — Verifies that unknown modes raise `ValueError`.
- `test_evaluate_noop_when_disabled` — Verifies that `evaluation.enabled: false` returns without error.

The existing 4 Qwen3-VL adapter tests (`test_qwen_adapter_llava_bench_judge.py`) were also run to confirm no regressions. All 9 tests pass.

```bash
cd /home/d.savchenkov/TICO/test/quantization/recipes && \
/home/d.savchenkov/myenv/bin/python -m pytest test_gemma4_adapter_llava_bench_judge.py -v

======================================================= test session starts =======================================================
cachedir: .pytest_cache
configfile: pyproject.toml
plugins: anyio-4.13.0
collected 5 items                                                                                                                 

test_gemma4_adapter_llava_bench_judge.py::TestGemma4AdapterLlavaBenchJudge::test_evaluate_dispatches_legacy_llava_bench_boolean_path PASSED [ 20%]
test_gemma4_adapter_llava_bench_judge.py::TestGemma4AdapterLlavaBenchJudge::test_evaluate_dispatches_nested_legacy_mode PASSED [ 40%]
test_gemma4_adapter_llava_bench_judge.py::TestGemma4AdapterLlavaBenchJudge::test_evaluate_dispatches_nested_llava_bench_judge_config PASSED [ 60%]
test_gemma4_adapter_llava_bench_judge.py::TestGemma4AdapterLlavaBenchJudge::test_evaluate_noop_when_disabled PASSED         [ 80%]
test_gemma4_adapter_llava_bench_judge.py::TestGemma4AdapterLlavaBenchJudge::test_evaluate_rejects_unknown_llava_bench_mode PASSED [100%]

======================================================== 5 passed in 7.50s ========================================================
```

## Example Scripts

No new example scripts were added. The existing config `gemma4_e2b_llava_bench_judge.yaml` can now be used with the following command line:

### FP baseline evaluation

```
$ python tico/quantization/examples/quantize.py --config tico/quantization/examples/configs/gemma4_e2b_llava_bench_judge.yaml
=== Config ===
Model                 : google/gemma-4-e2b-it
Device                : cuda
DType                 : float32
Seed                  : 42
GPTQ enabled          : False
GPTQ lm_head enabled  : False
PTQ enabled           : False
SpinQuant enabled     : False
CLE enabled           : False
Activation            : not set
Linear weight         : not set
Embedding weight      : not set
LM head weight        : not set
Spin rotation weight  : disabled
Calibration samples   : 1
Calibration seq length: 2048
Max seq length        : 2048
Profile               : not set

=== Loading model ===
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1951/1951 [00:04<00:00, 474.52it/s]
=== Building calibration inputs ===
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 68/68 [00:00<00:00, 13399.07it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 36/36 [00:00<00:00, 13560.39it/s]
Resolving data files: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 143/143 [00:00<00:00, 13349.33it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 68/68 [00:00<00:00, 13955.02it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 36/36 [00:00<00:00, 13444.48it/s]
Resolving data files: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 143/143 [00:00<00:00, 14032.04it/s]
[info] Loaded dataset: lmms-lab/VQAv2 (testdev, streaming), size=1
=== Running quantization pipeline ===
=== Evaluation ===
README.md: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1.64k/1.64k [00:00<00:00, 3.50MB/s]
[1] question_id=0 answer='Based on the image, which shows a dramatic volcanic cone, a coastline with turquoise water, and a road winding along the edge, this is very likely a view of **Mount Fuji** in Japan, or a similar volcanic landscape.\n\nHowever, without more specific identifying markers, it is difficult to name it with absolute certainty. If you can provide any additional context (like the location or the source of the photo), I might be able to give you a more precise answer.\n\n**If you are referring to a famous volcanic landscape in Japan, it strongly resembles views of Mount Fuji.**'
[2] question_id=1 answer="This is an aerial photograph showcasing a dramatic coastal landscape...
```

### PTQ + evaluation

```
$ python tico/quantization/examples/quantize.py \
  --config tico/quantization/examples/configs/gemma4_e2b_ptq_only.yaml \
  --set evaluation.enabled=true \
  --set evaluation.llava_bench.enabled=true \
  --set evaluation.llava_bench.mode=judge
=== Config ===
Model                 : google/gemma-4-e2b-it
Device                : cuda
DType                 : float32
Seed                  : 42
GPTQ enabled          : False
GPTQ lm_head enabled  : False
PTQ enabled           : True
SpinQuant enabled     : False
CLE enabled           : False
Activation            : int16
Linear weight         : uint4
Vision patch weight   : uint8
Embedding weight      : uint8
LM head weight        : uint8
Spin rotation weight  : disabled
Calibration samples   : 8
Calibration seq length: 2048
Max seq length        : 2048
Profile               : not set

=== Loading model ===
Loading weights: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1951/1951 [00:04<00:00, 470.47it/s]
=== Building calibration inputs ===
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 68/68 [00:00<00:00, 13385.24it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 36/36 [00:00<00:00, 13179.27it/s]
Resolving data files: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 143/143 [00:00<00:00, 13160.41it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 68/68 [00:00<00:00, 12881.07it/s]
Resolving data files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 36/36 [00:00<00:00, 13227.77it/s]
Resolving data files: 100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 143/143 [00:00<00:00, 13621.58it/s]
[info] Loaded dataset: lmms-lab/VQAv2 (testdev, streaming), size=8
=== Running quantization pipeline ===
Wrapping model with PTQ wrappers …
[Warn] GPTQ quantizers were not found; PTQ weight observers will use PTQ statistics.
Gemma4 PTQ calibration:   0%|                                                                                                                                 | 0/8 [00:00<?, ?it/s]/home/d.savchenkov/TICO/tico/quantization/wrapq/wrappers/gemma4/utils.py:141: UserWarning: Visual token count mismatch: expected 256, got 266. Using actual count. If this is unexpected, check image resolution settings.
  warnings.warn(
Gemma4 PTQ calibration:  38%|█████████████████████████████████████████████▍                                                                           | 3/8 [00:03<00:04,  1.03it/s]/home/d.savchenkov/TICO/tico/quantization/wrapq/wrappers/gemma4/utils.py:141: UserWarning: Visual token count mismatch: expected 256, got 273. Using actual count. If this is unexpected, check image resolution settings.
  warnings.warn(
Gemma4 PTQ calibration:  75%|██████████████████████████████████████████████████████████████████████████████████████████▊                              | 6/8 [00:05<00:01,  1.07it/s]/home/d.savchenkov/TICO/tico/quantization/wrapq/wrappers/gemma4/utils.py:141: UserWarning: Visual token count mismatch: expected 256, got 272. Using actual count. If this is unexpected, check image resolution settings.
  warnings.warn(
Gemma4 PTQ calibration: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8/8 [00:07<00:00,  1.05it/s]
=== Evaluation ===
[1] question_id=0 answer="Based on the image, which shows a dramatic volcanic landscape with a steep, eroded cone, a coastline, and lush vegetation, this is very likely a view of **Mount Ruapehu** or another prominent volcano in the **North Island of New Zealand**.\n\nThe specific features—the steep, dark slopes, the coastal view, and the general topography—are highly characteristic of New Zealand's volcanic regions.\n\nIf you can provide more context or a closer view, a more definitive identification might be possible. However, without further information, the most probable answer is a **volcanic landscape in New Zealand**."
[2] question_id=1 answer='This is an aerial photograph showcasing a dramatic coastal landscape...'
```

### DIFF
--- a/tico/quantization/recipes/adapters/gemma4.py
+++ b/tico/quantization/recipes/adapters/gemma4.py
@@ -23,6 +23,13 @@ from tico.quantization.config.gemma4_builders import build_gemma4_e2b_ptq_config
 from tico.quantization.recipes.adapters.base import ModelAdapter
 from tico.quantization.recipes.context import RecipeContext
 from tico.quantization.recipes.data.vlm import build_vlm_calibration_inputs
+from tico.quantization.recipes.evaluation.llava_bench_judge import (
+    evaluate_and_print_llava_bench_judge,
+)
+from tico.quantization.recipes.evaluation.vlm import (
+    evaluate_llava_bench,
+    print_coco_score_results,
+)
 from tico.quantization.recipes.utils import (
     move_to_device,
     quant_spec_from_config,
@@ -53,7 +60,12 @@ class Gemma4Adapter(ModelAdapter):
         cache_dir = model_cfg.get("cache_dir")
         device_map = runtime_cfg.get("device_map")
         if device_map is None:
-            device_map = "auto" if ctx.device.type != "cpu" else "cpu"
+            if ctx.device.type == "cpu":
+                device_map = "cpu"
+            else:
+                # Use specific device index instead of "auto" to avoid
+                # multi-GPU model split during PTQ calibration.
+                device_map = str(ctx.device)
 
         ctx.processor = AutoProcessor.from_pretrained(
             name,
@@ -174,11 +186,68 @@ class Gemma4Adapter(ModelAdapter):
         return convert(prepared, inplace=True)
 
     def evaluate(self, ctx: RecipeContext) -> None:
-        """Evaluate Gemma4 E2B.
-
-        TODO: Reuse the Qwen3-VL LLaVA-Bench judge evaluator once the adapter is
-        wired into the shared evaluation entry point.
-        """
-        if not ctx.cfg.get("evaluation", {}).get("enabled", False):
+        """Evaluate Gemma4 E2B via LLaVA-Bench judge and other VLM tasks."""
+        eval_cfg = ctx.cfg.get("evaluation", {})
+        if not eval_cfg.get("enabled", False):
             return
-        raise NotImplementedError("Gemma4 E2B evaluation adapter is not wired yet.")
+
+        max_seq_len = eval_cfg.get("max_seq_len")
+        n_samples = int(eval_cfg.get("n_samples", 50))
+
+        llava_bench_cfg = eval_cfg.get("llava_bench", False)
+        if isinstance(llava_bench_cfg, Mapping):
+            if llava_bench_cfg.get("enabled", False):
+                mode = str(llava_bench_cfg.get("mode", "judge")).lower()
+                if mode in {"judge", "llm_judge"}:
+                    evaluate_and_print_llava_bench_judge(
+                        model=ctx.model,
+                        processor=ctx.processor,
+                        device=str(ctx.device),
+                        llava_cfg=llava_bench_cfg,
+                        model_cfg=ctx.cfg.get("model", {}),
+                        runtime_cfg=ctx.cfg.get("runtime", {}),
+                        default_n_samples=n_samples,
+                        default_max_seq_len=max_seq_len,
+                    )
+                elif mode in {"legacy", "coco", "caption"}:
+                    llava_results = evaluate_llava_bench(
+                        model=ctx.model,
+                        processor=ctx.processor,
+                        device=str(ctx.device),
+                        n_samples=int(llava_bench_cfg.get("n_samples", n_samples)),
+                        max_seq_len=llava_bench_cfg.get("max_seq_len", max_seq_len),
+                    )
+                    print_coco_score_results(
+                        "\n=== LLaVA Bench Legacy COCO-style Evaluation ===",
+                        llava_results,
+                    )
+                else:
+                    raise ValueError(
+                        "evaluation.llava_bench.mode must be one of "
+                        "{'judge', 'llm_judge', 'legacy', 'coco', 'caption'}, "
+                        f"got {mode!r}."
+                    )
+        elif llava_bench_cfg:
+            print(
+                "[WARNING] evaluation.llava_bench=true uses the legacy "
+                "COCO-style CIDEr/BLEU path. Prefer the nested judge config: "
+                "evaluation.llava_bench.enabled=true, mode=judge."
+            )
+            llava_results = evaluate_llava_bench(
+                model=ctx.model,
+                processor=ctx.processor,
+                device=str(ctx.device),
+                n_samples=n_samples,
+                max_seq_len=max_seq_len,
+            )
+            print_coco_score_results("\n=== Llava Bench Evaluation ===", llava_results)
+
+    def export(self, ctx: RecipeContext) -> None:
+        """Export Gemma4 E2B artifacts.
+
+        TODO: Implement Circle export via the static runtime submodule adapters.
+        """
+        export_cfg = ctx.cfg.get("export", {})
+        if not export_cfg.get("enabled", False):
+            return
+        raise NotImplementedError("Gemma4 E2B export adapter is not wired yet.")
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_model.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_model.py
@@ -93,10 +93,12 @@ class QuantGemma4Model(QuantModuleBase):
             qcfg=qcfg.child("embed_vision") if qcfg else None,
             fp_name=join_name(fp_name, "embed_vision"),
         )
-        if getattr(fp_model, "audio_tower", None) is not None:
-            raise NotImplementedError(
-                "Gemma4 E2B skeleton does not implement audio static runtime."
-            )
+        # Leave audio tower in FP (not quantized) - audio is not supported yet.
+        self.audio_tower = (
+            fp_model.audio_tower
+            if getattr(fp_model, "audio_tower", None) is not None
+            else None
+        )
 
         self.visual_start_idx = int(
             self.qcfg.model_args.get("vision", {}).get("visual_start_idx", 0)
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_text_attention.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_text_attention.py
@@ -20,7 +20,7 @@ import torch.nn as nn
 import torch.nn.functional as F
 
 from tico.quantization.config.ptq import PTQConfig
-from tico.quantization.wrapq.utils.utils import join_name
+from tico.quantization.wrapq.utils.utils import get_model_arg, join_name
 from tico.quantization.wrapq.wrappers.ptq_wrapper import PTQWrapper
 from tico.quantization.wrapq.wrappers.quant_module_base import QuantModuleBase
 from tico.quantization.wrapq.wrappers.registry import try_register
@@ -75,7 +75,7 @@ class QuantGemma4TextAttention(QuantModuleBase):
         self.attention_dropout = float(
             getattr(fp_attn, "attention_dropout", 0.0) or 0.0
         )
-        self.max_seq = int(getattr(self.config, "max_position_embeddings"))
+        self.max_seq = self._resolve_max_seq()
 
         self.q_proj = PTQWrapper(
             fp_attn.q_proj,
@@ -165,6 +165,53 @@ class QuantGemma4TextAttention(QuantModuleBase):
         )
         mask.triu_(1)
         self.register_buffer("causal_mask_template", mask, persistent=False)
+
+    _DEFAULT_STATIC_MAX_SEQ = 2048
+
+    def _resolve_max_seq(self) -> int:
+        """Resolve the causal mask template capacity from config or model_args.
+
+        Resolution precedence:
+        1. ``calibration.seq_len`` (from PTQ config, same as Llama/Qwen adapters)
+        2. ``model_args["text"]["max_seq"]``
+        3. ``model_args["max_seq"]``
+        4. ``min(max_position_embeddings, 2048)``
+
+        This avoids allocating a ``(1, 1, max_position_embeddings, max_position_embeddings)``
+        buffer when the model supports a very large context window (e.g. 128K)
+        but the actual calibration/export sequence length is much smaller.
+        """
+        # Try calibration.seq_len first (same as Llama/Qwen adapters)
+        from tico.quantization.recipes.config import get_by_path
+
+        try:
+            calib_seq_len = get_by_path(self.qcfg, "calibration.seq_len")
+            if calib_seq_len is not None:
+                return int(calib_seq_len)
+        except (KeyError, TypeError):
+            pass
+
+        # Fall back to model_args
+        configured = get_model_arg(self.qcfg, "text", "max_seq", default=None)
+        if configured is None:
+            configured = get_model_arg(self.qcfg, "max_seq", default=None)
+
+        model_capacity = int(getattr(self.config, "max_position_embeddings"))
+        max_seq = (
+            min(model_capacity, self._DEFAULT_STATIC_MAX_SEQ)
+            if configured is None
+            else int(configured)
+        )
+        if max_seq <= 0:
+            raise ValueError(
+                f"Gemma4 attention max_seq must be positive, got {max_seq}."
+            )
+        if max_seq > model_capacity:
+            raise ValueError(
+                "Gemma4 attention max_seq exceeds max_position_embeddings: "
+                f"max_seq={max_seq}, model_capacity={model_capacity}."
+            )
+        return max_seq
 
     @staticmethod
     def _expand_rope_table(table: torch.Tensor) -> torch.Tensor:
--- a/tico/quantization/wrapq/wrappers/gemma4/utils.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/utils.py
@@ -132,9 +132,18 @@ def fixed_slot_fuse(
         )
 
     visual_len = int(visual_embeds.shape[1])
+    # Use actual visual token count - config num_visual_tokens is a hint for
+    # static export, but calibration images may produce different token counts.
     expected_len = visual_len if num_visual_tokens is None else int(num_visual_tokens)
     if visual_len != expected_len:
-        raise ValueError(f"Expected {expected_len} visual tokens, got {visual_len}.")
+        # Warn instead of raising error - use actual token count for flexibility
+        import warnings
+
+        warnings.warn(
+            f"Visual token count mismatch: expected {expected_len}, got {visual_len}. "
+            "Using actual count. If this is unexpected, check image resolution settings."
+        )
+        expected_len = visual_len
 
     end = int(visual_start_idx) + expected_len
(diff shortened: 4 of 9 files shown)


tico/quantization/wrapq/wrappers/gemma4/quant_text_attention.py
```diff
@@ -75,7 +75,7 @@ def __init__(
         self.attention_dropout = float(
             getattr(fp_attn, "attention_dropout", 0.0) or 0.0
         )
-        self.max_seq = int(getattr(self.config, "max_position_embeddings"))
+        self.max_seq = self._resolve_max_seq()
```
dvsav, line 78: **Problem:** The causal mask template allocation used `max_position_embeddings` directly, causing OOM for models with large context windows (e.g., 128K) even when calibration uses much shorter sequences.

**Fix:** Added `_resolve_max_seq()` method that resolves the mask capacity from `model_args["text"]["max_seq"]` or `model_args["max_seq"]`, falling back to `min(max_position_embeddings, 2048)`. This prevents unnecessary large buffer allocations.

tico/quantization/examples/inspector.py
dvsav, line 1: **Problem:** The file `tico/quantization/examples/inspect.py` shadowed Python's built-in `inspect` module, causing `ImportError: cannot import name 'signature' from partially initialized module 'inspect'` when third-party libraries (like `tqdm`) tried to import from the built-in module.

**Fix:** Renamed `inspect.py` to `inspector.py` and updated all import references in test files.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
```diff
@@ -93,10 +93,12 @@ def __init__(
...
-        if getattr(fp_model, "audio_tower", None) is not None:
-            raise NotImplementedError(
-                "Gemma4 E2B skeleton does not implement audio static runtime."
-            )
+        # Leave audio tower in FP (not quantized) - audio is not supported yet.
+        self.audio_tower = (
+            fp_model.audio_tower
+            if getattr(fp_model, "audio_tower", None) is not None
+            else None
+        )
```
dvsav, line 101: **Problem:** `Gemma4ForConditionalGeneration` models from HF include an `audio_tower` attribute, but the PTQ wrapper raised `NotImplementedError` when encountering it, blocking all PTQ runs.

**Fix:** Changed the audio tower handling to leave it in FP (passthrough) instead of raising an error. This allows vision+text PTQ to proceed while documenting that audio quantization is not yet supported.

tico/quantization/recipes/adapters/gemma4.py
```diff
@@ -53,7 +60,12 @@ def load_model(self, ctx: RecipeContext) -> RecipeContext:
         cache_dir = model_cfg.get("cache_dir")
         device_map = runtime_cfg.get("device_map")
         if device_map is None:
-            device_map = "auto" if ctx.device.type != "cpu" else "cpu"
+            if ctx.device.type == "cpu":
+                device_map = "cpu"
+            else:
+                # Use specific device index instead of "auto" to avoid
+                # multi-GPU model split during PTQ calibration.
+                device_map = str(ctx.device)
```
dvsav, line 68: **Problem:** Using `device_map="auto"` caused the model to split across multiple GPUs, but calibration batches were sent to a single device, resulting in `RuntimeError: Expected all tensors to be on the same device, but got index is on cuda:0, different from other tensors on cuda:1`.

**Fix:** Changed `device_map` logic to use the specific device index (e.g., `"cuda:0"`) instead of `"auto"`, ensuring the entire model loads on a single GPU during PTQ calibration.

tico/quantization/wrapq/wrappers/gemma4/utils.py
```diff
@@ -132,9 +132,18 @@ def fixed_slot_fuse(
...
     if visual_len != expected_len:
-        raise ValueError(f"Expected {expected_len} visual tokens, got {visual_len}.")
+        # Warn instead of raising error - use actual token count for flexibility
+        import warnings
+
+        warnings.warn(
+            f"Visual token count mismatch: expected {expected_len}, got {visual_len}. "
+            "Using actual count. If this is unexpected, check image resolution settings."
+        )
+        expected_len = visual_len
```
dvsav, line 146: **Problem:** The `fixed_slot_fuse()` function raised `ValueError` when calibration images produced a different number of visual tokens than the configured `num_visual_tokens` (e.g., 266 vs 256), blocking PTQ calibration.

**Fix:** Changed the behavior to warn instead of raising an error, using the actual visual token count from the vision tower. The `num_visual_tokens` config value is now treated as a hint for static export rather than a strict requirement for calibration.

**Test Update:** Updated `test_fixed_slot_fuse_rejects_wrong_visual_length` to `test_fixed_slot_fuse_warns_on_wrong_visual_length` to verify the warning behavior instead of asserting an error.

tico/quantization/wrapq/wrappers/gemma4/quant_text_attention.py
Torrero: Maybe we can use existing mechanism of `max_position_embeddings` restriction which has already been implemented for LLama and Qwen?  https://github.com/Samsung/TICO/blob/dc30b0f5cf7a2079448a14dc5200a53cf5275d2f/tico/quantization/recipes/adapters/llama.py#L186 https://github.com/Samsung/TICO/blob/dc30b0f5cf7a2079448a14dc5200a53cf5275d2f/tico/quantization/recipes/adapters/qwen3_vl.py#L107

tico/quantization/examples/inspector.py
Torrero: I think that changing the name of the module should be in all references, not only in tests, but also in utils, READMEs.

mhs4670go: I agree on renaming the module name to `inspector`. But, it can be split into another PR. I'll make a PR for this issue.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
mhs4670go: Just curiosity, llava-bench needs audio/video support?

tico/quantization/wrapq/wrappers/gemma4/utils.py
mhs4670go: I think we should consider a slightly different structure for the Gemma4 wrapper before solving the visual-token mismatch by relaxing `fixed_slot_fuse()`.

My main concern is that the current approach makes the normal PTQ/calibration path depend on the static export layout too early. In the current design, `fixed_slot_fuse()` is used directly in the forward path, so calibration, eager execution, and static/export behavior are all tied to the same fixed visual-token contract. That works for a narrow static-runtime prototype, but it makes the wrapper less robust as a general Gemma4 quantization wrapper.

A better long-term design would be closer to the Qwen3-VL wrapper pattern:

```
eager / PTQ calibration path:
  use dynamic placeholder-mask fusion, matching HF semantics

static / export path:
  use fixed-slot fusion with strict layout validation

static-runtime calibration profile:
  resize/preprocess images to the target static layout
  validate visual token count, placeholder span, and sequence length
```

In other words, I would prefer:

```
dynamic semantic path for calibration
+
strict static validation for export/runtime
```

rather than

```
fixed-slot fusion everywhere
or
fixed-slot fusion with warning-only mismatch handling
```

The reason is that PTQ calibration should ideally observe activations from a path that matches the original model semantics. If the input contains image placeholder tokens at certain positions, the eager/calibration path should fuse image embeddings into those actual placeholder positions. That makes the wrapper more faithful to the HF model and avoids baking static runtime assumptions into ordinary calibration.

At the same time, I agree that the static runtime needs a strict fixed layout. For that path, `fixed_slot_fuse()` should remain strict. If the static config says `num_visual_tokens = 256`, then a sample that produces 266 or 273 visual tokens should not silently continue. That is a real static-layout mismatch and should be caught either during preprocessing or validation.

I'll post a draft for this approach soon. Please review it. This is a little more refactoring up front, but it gives us a cleaner and safer architecture than trying to make a static export helper flexible enough to handle calibration mismatches.

mhs4670go: I've posted [a draft](https://github.com/Samsung/TICO/pull/817). That being said, you can merge this PR anyway and re-consider the refactor later.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
dvsav: > Just curiosity, llava-bench needs audio/video support?

No, as long as I know, LLaVA-Bench (lmms-lab/llava-bench-in-the-wild) doesn't include audio or video inputs. Just images and text.

That said, the fix above is required regardless of LLaVA-Bench content. It's required because the `audio_tower` attribute exists in HF's `Gemma4ForConditionalGeneration` class.

tico/quantization/wrapq/wrappers/gemma4/quant_text_attention.py
dvsav: 👍 Done
Updated `_resolve_max_seq()` resolution precedence:
   1. `calibration.seq_len` (from PTQ config) - **NEW, highest priority**
   2. `model_args["text"]["max_seq"]`
   3. `model_args["max_seq"]`
   4. `min(max_position_embeddings, 2048)` (fallback)

tico/quantization/wrapq/wrappers/gemma4/utils.py
dvsav: @mhs4670go shall I roll back this to raising `ValueError` in this PR or shall I do that after [your draft](https://github.com/Samsung/TICO/pull/817) gets merged?

Torrero (approved):
LGTM

tico/quantization/wrapq/wrappers/gemma4/utils.py
mhs4670go: Let me refactor the wrapper first then. I'll rebase the draft soon. Thank you.

mhs4670go (approved):
LGTM

PR #817, opened 2026-07-08
[quantization] Split Gemma4 dynamic and static fusion paths

mhs4670go:
Refactor the Gemma4 wrapper to keep the eager/PTQ calibration path close to the original HF semantics while preserving strict fixed-slot fusion for static export/runtime.

- Add dynamic placeholder-mask fusion for eager/PTQ execution
- Keep fixed_slot_fuse strict for static/export paths
- Add static calibration image-size alignment in Gemma4Adapter

### DIFF
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_model.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_model.py
@@ -22,7 +22,9 @@ from tico.quantization.wrapq.mode import Mode
 from tico.quantization.wrapq.utils.utils import join_name
 from tico.quantization.wrapq.wrappers.gemma4.utils import (
     assert_gemma4_e2b_no_moe,
+    dynamic_placeholder_fuse,
     fixed_slot_fuse,
+    validate_static_visual_layout,
 )
 from tico.quantization.wrapq.wrappers.ptq_wrapper import PTQWrapper
 from tico.quantization.wrapq.wrappers.quant_module_base import QuantModuleBase
@@ -64,6 +66,10 @@ def _get_placeholder_mask(
 class QuantGemma4Model(QuantModuleBase):
     """PTQ wrapper skeleton for image-text Gemma4 E2B."""
 
+    # This boolean flag is used by unit tests to exercise the static fusion path
+    # without running torch.export.
+    force_export: bool = False
+
     def __init__(
         self,
         fp_model: nn.Module,
@@ -100,11 +106,11 @@ class QuantGemma4Model(QuantModuleBase):
             else None
         )
 
-        self.visual_start_idx = int(
-            self.qcfg.model_args.get("vision", {}).get("visual_start_idx", 0)
-        )
-        self.num_visual_tokens = int(
-            self.qcfg.model_args.get("vision", {}).get("num_visual_tokens", 0)
+        vision_args = self.qcfg.model_args.get("vision", {})
+        self.visual_start_idx = int(vision_args.get("visual_start_idx", 0))
+        self.num_visual_tokens = int(vision_args.get("num_visual_tokens", 0))
+        self.validate_static_layout = bool(
+            vision_args.get("validate_static_layout", self.num_visual_tokens > 0)
         )
         self.obs_mm_fusion = self._make_obs("mm_fusion")
         self.obs_per_layer_inputs = self._make_obs("per_layer_inputs")
@@ -123,6 +129,44 @@ class QuantGemma4Model(QuantModuleBase):
         )
         return self.embed_vision(vision_outputs.last_hidden_state)
 
+    def _uses_static_fusion(self) -> bool:
+        """Return whether the current call should use static fixed-slot fusion."""
+        return bool(torch.compiler.is_compiling() or self.force_export)
+
+    def _validate_runtime_multimodal_masks(
+        self,
+        *,
+        image_mask: torch.Tensor,
+        video_mask: torch.Tensor,
+        audio_mask: torch.Tensor,
+        pixel_values: Optional[torch.Tensor],
+    ) -> None:
+        """Reject unsupported or incomplete multimodal eager inputs."""
+        if video_mask.any() or audio_mask.any():
+            raise NotImplementedError(
+                "Gemma4 PTQ wrapper supports image-text inputs only. "
+                "Video and audio placeholder tokens are not supported."
+            )
+        if pixel_values is None and image_mask.any():
+            raise ValueError(
+                "pixel_values must be provided when input_ids contain image placeholders."
+            )
+
+    def _validate_static_image_layout(
+        self,
+        image_mask: torch.Tensor,
+        seq_len: int,
+    ) -> None:
+        """Validate image placeholders against the configured static layout."""
+        if not self.validate_static_layout:
+            return
+        validate_static_visual_layout(
+            image_mask,
+            visual_start_idx=self.visual_start_idx,
+            num_visual_tokens=self.num_visual_tokens,
+            seq_len=seq_len,
+        )
+
     def forward(
         self,
         input_ids: Optional[torch.Tensor] = None,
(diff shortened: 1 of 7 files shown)


dvsav:
## Note for me

### Problem

The original [skeleton commit](https://github.com/Samsung/TICO/commit/eb371415b0790ec8c09d7880fdd2ca22fade43cf) had a single multimodal fusion path: **`fixed_slot_fuse`** was used unconditionally in `QuantGemma4Model.forward()`. This function writes visual embeddings into a *hardcoded contiguous position range* `[visual_start_idx, visual_start_idx + num_visual_tokens)` in the text embedding tensor.

This created a fundamental tension between two execution contexts:

1. **Eager/PTQ calibration** — During calibration, the model runs in eager mode (no `torch.export`). The HuggingFace processor produces `input_ids` where image placeholder tokens appear at *arbitrary positions* determined by the tokenizer — they are **not** guaranteed to be in a contiguous block at `visual_start_idx`. The original HF model uses `masked_scatter` to place visual embeddings at the actual placeholder positions. By forcing `fixed_slot_fuse` during calibration, the wrapper would write visual features to the wrong token positions, producing **incorrect activation statistics** and therefore **wrong quantization parameters**.

2. **Static export/runtime** — For `torch.export` and NPU deployment, the graph must be fully static with no data-dependent control flow. `masked_scatter` (boolean indexing) is not exportable. The `fixed_slot_fuse` approach (slice assignment at a known position) is the correct strategy here, but it requires that the input layout actually matches the configured static span.

In short: **one fusion function cannot serve both contexts**. The eager path needs dynamic, mask-based fusion to match HF semantics; the export path needs static, position-based fusion to be graph-capturable.

---

### The Solution Design

The commit introduces a **dual-path fusion architecture** that dispatches based on execution context:

#### 1. New `dynamic_placeholder_fuse()` function (`utils.py`)

A new fusion function that replicates HF's `masked_scatter` semantics using boolean indexing:

```python
fused[batch_idx, image_mask[batch_idx], :] = visual_embeds[batch_idx]
```

- Takes an `image_mask` tensor `(B, S)` that marks the actual image placeholder positions
- Writes visual embeddings to those exact positions, wherever they are
- Uses `torch.clone()` + per-batch boolean indexing — **not exportable** but correct for eager calibration
- Includes thorough validation: rank checks, batch/hidden dim compatibility, placeholder count must match visual token count

#### 2. `validate_static_visual_layout()` function (`utils.py`)

A validator that checks whether the image placeholder mask **exactly matches** the configured static visual-token span:

```python
expected = torch.zeros_like(mask, dtype=torch.bool)
expected[:, start:end] = True
assert torch.equal(mask, expected)
```

This is the **bridge** between the two paths: it verifies that the dynamic calibration inputs are layout-compatible with the static export contract. If calibration data passes this validation, then `dynamic_placeholder_fuse` and `fixed_slot_fuse` will produce **identical results** — which is explicitly tested.

#### 3. Dispatch logic in `QuantGemma4Model.forward()` (`quant_model.py`)

The forward method now branches:

```python
if self._uses_static_fusion():
    # torch.export or force_export → fixed_slot_fuse (static, exportable)
    inputs_embeds = fixed_slot_fuse(...)
else:
    # eager/PTQ calibration → dynamic_placeholder_fuse (correct HF semantics)
    inputs_embeds = dynamic_placeholder_fuse(...)
```

The dispatch key is `_uses_static_fusion()`:
```python
return bool(torch.compiler.is_compiling() or self.force_export)
```

- `torch.compiler.is_compiling()` — True during `torch.export`, False during eager
- `force_export` — class-level flag for unit testing the static path without running export

#### 4. Static calibration image-size alignment (`Gemma4Adapter`)

The adapter now reads `model_args.vision.image_height` and `image_width` from config and **temporarily patches the image processor** to resize all calibration images to that fixed size:

```python
@contextmanager
def _fixed_image_processor_size(processor, image_size):
    # Set do_resize=True, size={height, width}, crop_size={height, width}
    # ... yield ...
    # Restore original attributes
```

This ensures that every calibration image produces the **same number of visual tokens** (since patch count = image_area / patch_size²), which is a prerequisite for the static layout contract (`num_visual_tokens` must be fixed).

#### 5. Runtime multimodal input validation

New `_validate_runtime_multimodal_masks()` method rejects:
- Video/audio placeholder tokens (v0 scope is image-text only)
- Image placeholders without corresponding `pixel_values`

This validation only runs in eager mode (`not torch.compiler.is_compiling()`), avoiding graph breaks during export.

---

### Key Design Properties

| Property | Dynamic path (eager/PTQ) | Static path (export/runtime) |
|---|---|---|
| Fusion function | `dynamic_placeholder_fuse` | `fixed_slot_fuse` |
| Position source | `image_mask` (from `input_ids`) | `visual_start_idx` (config) |
| Exportable? | ❌ (boolean indexing) | ✅ (slice assignment) |
| HF semantics? | ✅ (matches `masked_scatter`) | ✅ (if layout validated) |
| Layout validation | `validate_static_visual_layout` | `validate_static_visual_layout` |

The critical invariant: **if `validate_static_visual_layout` passes, both paths produce identical outputs.** This is verified by the test `test_dynamic_and_fixed_fusion_match_for_valid_static_layout`.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
```diff
@@ -239,12 +291,38 @@ def forward(
...
+            else:
+                if image_mask is None:
+                    raise ValueError(
+                        "input_ids must be provided with pixel_values when running "
+                        "Gemma4 eager/PTQ dynamic placeholder fusion."
+                    )
+                self._validate_static_image_layout(
+                    image_mask,
+                    seq_len=int(inputs_embeds.shape[1]),
+                )
```
dvsav: ### Do we need to validate static image layout in the dynamic (eager) path?

The eager path always calls `_validate_static_image_layout()`, which calls `validate_static_visual_layout()`. This validator checks that image placeholders are exactly in `[visual_start_idx, visual_start_idx + num_visual_tokens)`.

But the whole point of the dynamic path is to handle calibration data where image placeholders may not be in the static span. If the calibration dataset produces prompts where image tokens are at different positions (which is the normal HF behavior for variable-length text prefixes), this validation will reject all calibration data.

The `validate_static_layout` flag defaults to `True` when `num_visual_tokens > 0` (`QuantGemma4Model.__init__`):
```
self.validate_static_layout = bool(
    vision_args.get("validate_static_layout", self.num_visual_tokens > 0)
)
```

So if a user configures `num_visual_tokens` (which they must for the static path), the eager calibration path will enforce that calibration data already has the static layout.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
```diff
@@ -64,6 +66,10 @@ def _get_placeholder_mask(
 class QuantGemma4Model(QuantModuleBase):
     """PTQ wrapper skeleton for image-text Gemma4 E2B."""
 
+    # This boolean flag is used by unit tests to exercise the static fusion path
+    # without running torch.export.
+    force_export: bool = False
```
dvsav: Although this flag is intended specifically for unit tests, unit tests that might use it were not added.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
dvsav, line 312: The adapter patches the image processor to produce fixed-size images, which should produce a fixed number of visual tokens. But there is no assertion that the actual number of visual tokens produced by the vision tower matches `config.model_args.vision.num_visual_tokens`. If the image size is misconfigured (e.g., wrong patch size), the vision tower will produce a different number of tokens, and the error will only surface deep in `fixed_slot_fuse` with a confusing "Visual token count mismatch" error.

**Suggestion:** Add an early assertion in `get_image_features()` or at the start of the fusion step:
```python
actual_visual_tokens = image_embeds.shape[1]
if actual_visual_tokens != self.num_visual_tokens:
    raise ValueError(
        f"Vision tower produced {actual_visual_tokens} visual tokens but "
        f"config expects {self.num_visual_tokens}. Check image_height/image_width."
    )
```

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
mhs4670go: Good catch. I'll add a test.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
mhs4670go: You are right that static layout validation does not belong in a generic dynamic/eager path. I'll add more detail documentation for reducing confusion.

tico/quantization/wrapq/wrappers/gemma4/quant_model.py
mhs4670go: Partially agreed. Because the dynamic/eager path should allow variable visual token counts as long as they match the actual image placeholder count.

Therefore, the check is useful for the static/export path and for explicit static-layout calibration. I will add a static-only visual token count validator before fixed_slot_fuse so misconfigured image_height / image_width / num_visual_tokens produce a clearer error message.

dvsav (approved):
LGTM

Issue #812, opened 2026-07-07
Static-shape multi-turn support with append-prefill


mhs4670go:
## Proposal

This is a follow-up to the current static-shape prefill/decode design described in:

https://github.com/Samsung/TICO/issues/586#issuecomment-4161647429

The target NPU only supports static shapes. Also, handling `attention_mask` and `position_embeddings` inside the NPU graph is inefficient, so these are intentionally prepared by the runtime.

Therefore, the multi-turn design should keep the same principle as the existing static prefill/decode design:

- exported decoder-layer graphs use fully static tensor shapes
- KV cache is managed externally by the runtime
- `attention_mask` is prepared by the runtime
- `position_embeddings = (cos, sin)` are prepared by the runtime
- the exported graph does not compute dynamic positions
- the exported graph does not build causal masks
- the exported graph does not own or update the full KV cache

The missing part is efficient multi-turn cache reuse.

Today, after the first prefill and assistant decode, a new user turn can be handled in two simple ways:

1. Re-run prefill on the full conversation history.
2. Feed the new user turn token-by-token through the decode path.

Both are functionally possible, but neither is ideal:

- full-history prefill discards the benefit of the existing KV cache
- decode-loop append requires one graph invocation per user token

For multi-turn serving, we should add a third path:

```text
append-prefill / chunk-prefill
```

This path consumes multiple new user-turn tokens at once and appends their KV tensors to the existing external cache.

---

## High-level runtime flow

The intended multi-turn flow is:

```text
1. prefill(initial prompt)
   - no past KV
   - returns prompt KV
   - runtime writes prompt KV into external cache

2. decode assistant tokens
   - single-token decode
   - runtime updates external cache one token at a time

3. append-prefill(next user turn)
   - consumes multiple new user tokens
   - uses existing external cache as past KV
   - returns only delta KV for the new user tokens
   - runtime appends the delta KV into the external cache

4. decode next assistant tokens
   - continues from the updated cache
```

So the static graph set becomes:

```text
prefill:
  first prompt / full-context recompute path

append-prefill:
  multi-token cache append path for new user turns

decode:
  single-token generation path
```

`append-prefill` should be a separate export mode instead of overloading the existing prefill or decode contract.

---

## Terminology

Let:

```text
B       = batch size
D       = hidden size
Hkv     = number of KV heads
Hd      = head dimension
K       = fixed attention key length, usually max_seq
P       = number of valid tokens already stored in external KV cache
[...]

PR #818, opened 2026-07-08
[quantization] Support multi-turn workflows

mhs4670go:
This commit supports multi-turn workflows.

```bash
python -m tico.quantization.recipes.debug.static_llama_runtime \
  --model /mnt/sdb/model/Llama-3.2-1B-Instruct \
  --max-seq 256 \
  --device cpu \
  --no-run-baseline \
  --run-multiturn \
  --append-bucket 16


python -m tico.quantization.recipes.debug.static_llama_runtime \
  --model /mnt/sdb/model/Llama-3.2-1B-Instruct \
  --max-seq 256 \
  --padding-side right \
  --device cpu \
  --no-run-baseline \
  --run-multiturn \
  --multiturn-first-prompt $'User: What is the capital of France?\nAssistant:' \
  --multiturn-second-turn $'\nUser: What country is that city in?\nAssistant:' \
  --multiturn-decode-steps-before-append 4 \
  --multiturn-decode-steps-after-append 4 \
  --append-bucket 16 \
  --compare-append-prefill-with-decode-loop

```

```bash
...
====================================================================================================
Multi-turn post-append decode logits step 1
mean|diff| = 0.02939297
 max|diff| = 0.10098839
PEIR       = 0.340664 %
====================================================================================================
Multi-turn post-append decode logits step 2
mean|diff| = 0.03495336
 max|diff| = 0.17215163
PEIR       = 0.569002 %
====================================================================================================
Multi-turn post-append decode logits step 3
mean|diff| = 0.02905156
 max|diff| = 0.19826031
PEIR       = 0.583857 %
====================================================================================================
Multi-turn post-append decode logits step 4
mean|diff| = 0.01603191
 max|diff| = 0.09334183
PEIR       = 0.341133 %
====================================================================================================
Multi-turn generated text:
User: What is the capital of France?
Assistant: The capital of France
User: What country is that city in?
Assistant: France
User:

```

Related: #812

### DIFF
--- a/tico/quantization/wrapq/wrappers/llama/export_adapters.py
+++ b/tico/quantization/wrapq/wrappers/llama/export_adapters.py
@@ -323,6 +323,65 @@ class LlamaAttentionDecodeExportAdapter(nn.Module):
         return hidden, new_k, new_v
 
 
+class LlamaAttentionAppendPrefillExportAdapter(nn.Module):
+    """
+    Export adapter for append-prefill attention.
+
+    Append-prefill consumes a fixed-size block of new tokens and attends to a
+    runtime-managed static past buffer. The runtime must prepare the additive
+    attention mask and RoPE position embeddings outside the exported graph.
+
+    Input contract
+    --------------
+    - hidden_states:        (B, Q, D)
+    - position_embeddings:  (B, Q, head_dim)
+    - attention_mask:       (B, Q, K)
+    - past_key_value:       (B, num_kv_heads, K - Q, head_dim)
+
+    Return contract
+    ---------------
+    - return_kv=True:
+        (hidden_states, (new_key, new_value))
+      where new_key/new_value are the KV delta for the appended block with
+      shape `(B, num_kv_heads, Q, head_dim)`.
+
+    - return_kv=False:
+        hidden_states
+    """
+
+    def __init__(self, wrapped, *, return_kv: bool = True):
+        super().__init__()
+        self.wrapped = wrapped
+        self.return_kv = return_kv
+
+    def forward(
+        self,
+        hidden_states: torch.Tensor,
+        position_embeddings: Tuple[torch.Tensor, torch.Tensor],
+        attention_mask: Optional[torch.Tensor] = None,
+        past_key_value: Optional[Tuple[torch.Tensor, torch.Tensor]] = None,
+        **kwargs,
+    ):
+        """Run append-prefill attention and return delta-only K/V tensors."""
+        outputs = self.wrapped(
+            hidden_states=hidden_states,
+            position_embeddings=position_embeddings,
+            attention_mask=attention_mask,
+            past_key_value=past_key_value,
+            use_cache=self.return_kv,
+            cache_output_mode="delta",
+            **kwargs,
+        )
+
+        hidden = outputs[0]
+
+        if not self.return_kv:
+            return hidden
+
+        new_k, new_v = outputs[2]
+        return hidden, new_k, new_v
+
+
 class LlamaDecoderLayerPrefillExportAdapter(nn.Module):
     """
     Export adapter for prefill.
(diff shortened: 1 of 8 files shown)


tico/quantization/wrapq/wrappers/llama/export_adapters.py
```diff
@@ -323,6 +323,65 @@ def forward(
         return hidden, new_k, new_v
 
 
+class LlamaAttentionAppendPrefillExportAdapter(nn.Module):
```
dvsav: Couldn't find any references to this class.

tico/quantization/wrapq/wrappers/llama/export_adapters.py
```diff
@@ -426,6 +485,69 @@ def forward(
         return hidden, new_k, new_v
 
 
+class LlamaDecoderLayerAppendPrefillExportAdapter(nn.Module):
```
dvsav: This class looks absolutely identical to `LlamaDecoderLayerDecodeExportAdapter`. Can we simply reuse the latter one?

tico/quantization/wrapq/wrappers/llama/export_adapters.py
```diff
@@ -323,6 +323,65 @@ def forward(
...
+    --------------
+    - hidden_states:        (B, Q, D)
+    - position_embeddings:  (B, Q, head_dim)
+    - attention_mask:       (B, Q, K)
+    - past_key_value:       (B, num_kv_heads, K - Q, head_dim)
+
+    Return contract
+    ---------------
+    - return_kv=True:
+        (hidden_states, (new_key, new_value))
```
dvsav: The actual code returns a flat tuple `(hidden_states, new_key, new_value)`, not a nested tuple.

dvsav:
### (Optional) Add unit tests

The PR adds a lot of new functionality (new export mode, new adapters, multi-turn runtime with attention mask construction, cache management, chunked append) but includes no unit tests. The only validation is the debug script `static_llama_runtime.py`. Key areas needing tests:
- `_build_append_prefill_attention_mask` correctness (causal within new block, masking of gap, padded rows).
- `append_prefill` cache write-back correctness.
- `append_prefill_chunked` with chunks smaller than bucket size.
- Export of append-prefill graphs at various bucket sizes.

tico/quantization/wrapq/wrappers/llama/export_adapters.py
```diff
@@ -426,6 +485,69 @@ def forward(
...
+      where new_key/new_value are the delta KV tensors for the current block.
+
+    - return_kv=False:
+        hidden_states
+    """
+
+    def __init__(self, wrapped, *, return_kv: bool = True):
+        super().__init__()
+        self.wrapped = wrapped
+        self.wrapped.return_type = "tuple"
```
dvsav, line 401: ### (Optional) `return_type = "tuple"` mutation in `__init__` is fragile

All three decoder-layer adapters set `self.wrapped.return_type = "tuple"` in `__init__`. Since the runtime creates all three adapters from the same wrapped module, the last `__init__` wins. This works today because all set the same value, but it's a latent bug if a future adapter sets a different `return_type`. Consider setting `return_type` in `forward()` or making it a parameter of the adapter rather than mutating the wrapped module.

dvsav:
### (Optional) Implement bucket selection logic

The [issue specification](https://github.com/Samsung/TICO/issues/812) describes runtime bucket selection: "The runtime selects the smallest usable bucket satisfying: `T <= Q` and `P + Q <= K`." The current `append_prefill_chunked` uses a **fixed** `bucket_size` for all chunks. A production runtime should select the smallest bucket that fits each chunk, and handle the case where no bucket fits (context-length policy). The current implementation (`_normalize_append_prefill_buckets`) raises `ValueError` when capacity is exceeded, which is acceptable for a first version but should be documented as a limitation.

dvsav (changes requested):

tico/quantization/wrapq/wrappers/llama/export_adapters.py
mhs4670go: Agreed. I'll remove the duplicate class and reuse it while keeping `append_prefill` as a separate export mode.

mhs4670go:
@dvsav PTAL:)

dvsav (dismissed):
👍 LGTM. Thank you!

Torrero (dismissed):
LGTM

mhs4670go:
@dvsav @Torrero I've just rebased the PR. PTAL

dvsav (approved):
👍

PR #823, opened 2026-07-10
[quantization] Printing override policies and model after PTQ

Torrero:
This PR prints the configuration of the override policies and model after PTQ using `print_model` option in yaml file.

It introduces two helper functions:

- __`format_override_policy(policy)`__ — Formats a single PTQ override policy into a compact, human-readable string.
- __`print_override_policies(ptq_stage)`__ — Iterates over the `override_policies` list from the PTQ stage config and prints each policy under a `--- Override policies ---` header.

Co-Authored-By: Cline

### DIFF
--- a/tico/quantization/recipes/runner.py
+++ b/tico/quantization/recipes/runner.py
@@ -148,6 +148,62 @@ def _max_seq_len(cfg: Mapping[str, Any]) -> Any:
     return "not set"
 
 
+def _format_override_policy(policy: Mapping[str, Any]) -> str:
+    """Return a compact, human-readable description of one override policy."""
+    name = policy.get("name", "<unnamed>")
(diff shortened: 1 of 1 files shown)


mhs4670go:
There are a few comments about this PR.

- The summary currently prints only the spec reference, e.g. spec=mx_int8_activation, while the actual kind, elem_format, axis, and rounding settings live under ptq.specs. Changing the named spec definition therefore produces the same summary even though the quantization behavior changes.
- The printed override information should include `raw_overrides` as well.
- We might as well move this diagnostic into PTQStage.run() after apply_ptq_override_policies() as each logic should be inside its module;ptq.

Since the change would be big, I'll post dedicated PR for this change. Could you review the PR instead?

PR #826, opened 2026-07-13
[examples] Update config for LLaMA

mhs4670go:
This commit updates configs to make it easier to use.

Current configuration doesn't reflects our flows enough. I've simply reduced them into three steps.

### LLaMA benchmark flow

Use this three-step loop for the common LLaMA post-quantization benchmark flow.

1. Quantize the model and save the configured checkpoint.

   ```bash
   python -m tico.quantization.examples.quantize \
     --config tico/quantization/examples/configs/llama_quantize.yaml
   ```

2. Evaluate the saved checkpoint with the configured LLM evaluation suite.

   ```bash
   python -m tico.quantization.examples.evaluate \
     --config tico/quantization/examples/configs/llama_eval_suite.yaml \
     --checkpoint ./out/llama_quantized/quantized_model.pt
   ```

3. Export per-layer Circle artifacts from the saved checkpoint.

   ```bash
   python -m tico.quantization.examples.export \
     --config tico/quantization/examples/configs/llama_export.yaml \
     --checkpoint ./out/llama_quantized/quantized_model.pt
   ```


Torrero (approved):
LGTM

PR #805, opened 2026-07-06
[DRAFT] Export fcircle

mhs4670go:
This draft exports fcircle.

```bash
python tico/quantization/wrapq/examples/quantize_full_qmodel_with_gptq.py \
  --model ~/models/Llama-3.2-3B-Instruct/ --max_seq_len 2048 \
  --output_dir . --save circle_per_layer --decode_calibration_steps 1
```

TICO-DCO-1.0-Signed-off-by: seongwoo <mhs4670go@naver.com>

PR #827, opened 2026-07-14
[quantization] Support export with float dtype

mhs4670go:
This commit supports an export with float dypte.

Related: #805 

### DIFF
--- a/tico/quantization/examples/export.py
+++ b/tico/quantization/examples/export.py
@@ -22,6 +22,8 @@ from tico.quantization.recipes.config import load_recipe_config, save_effective_
 from tico.quantization.recipes.context import RecipeContext
 from tico.quantization.recipes.utils import set_seed, torch_dtype_from_name
 
+_EXPORT_SOURCES = ("model", "checkpoint")
+
 
 def _parse_artifacts(value: str) -> str:
     artifacts = [item.strip() for item in value.split(",") if item.strip()]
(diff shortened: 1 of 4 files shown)


tico/quantization/examples/export.py
```diff
@@ -79,21 +126,30 @@ def main() -> None:
...
-        args.checkpoint,
-        map_location=device,
-        weights_only=False,
-    )
-    if hasattr(checkpoint, "to"):
-        checkpoint = checkpoint.to(device)
-    ctx.model = checkpoint.eval()
+    if source == "model":
+        ctx = RecipeContext(cfg=cfg, adapter=adapter)
+        ctx = adapter.load_model(ctx)
```
Torrero: Minor issue,
In this branch `RecipeContext` is created without `device`/`dtype`, but in opposite branch  `device` and `dtype` are used for `RecipeContext` creation.

tico/quantization/examples/export.py
```diff
@@ -58,11 +98,18 @@ def parse_args() -> argparse.Namespace:
         ),
     )
     parser.add_argument("--set", action="append", default=[], metavar="KEY=VALUE")
-    return parser.parse_args()
+    args = parser.parse_args()
+    try:
+        _resolve_source(args.source, args.checkpoint)
```
Torrero: Minor issue, 
`_resolve_source()` is called twice, maybe it can be resolved as follows:

```    
...
    try:
        args.source = _resolve_source(args.source, args.checkpoint)
    except ValueError as exc:
        parser.error(str(exc))
    return args
...

def main() -> None:
    args = parse_args()
    source = args.source 
```

tico/quantization/examples/export.py
mhs4670go: Thanks. The model path delegates device/dtype initialization to `adapter.load_model()`, while the checkpoint path initializes them explicitly because it skips model loading.

tico/quantization/examples/export.py
mhs4670go: Good catch. I normalized `args.source` in `parse_args()` and removed the second `_resolve_source()` call from `main()`.

Torrero (approved):
LGTM

PR #832, opened 2026-07-21
[examples] Simplify Qwen3-VL configs

mhs4670go:
This commit simplifies Qwen3-VL configs and consolidate evaluation presets.

@Torrero Please check if it helps your evaluation flows.


Torrero:
Thank you! Now the Qwen3-VL configs are clearer.

tico/quantization/examples/configs/qwen3_vl_eval_suite.yaml
```diff
@@ -107,16 +67,21 @@ evaluation:
...
-      answers: ./out/llava_bench/qwen3_vl_2b.answers.jsonl
-      reviews: ./out/llava_bench/qwen3_vl_2b.llama3_2_3b.reviews.jsonl
-      summary: ./out/llava_bench/qwen3_vl_2b.llama3_2_3b.summary.json
+      dir: ./out/qwen3_vl_eval_suite/llava_bench
+      answers: ./out/qwen3_vl_eval_suite/llava_bench/qwen3_vl_4b.answers.jsonl
+      reviews: ./out/qwen3_vl_eval_suite/llava_bench/qwen3_vl_4b.llama3_2_3b.reviews.jsonl
+      summary: ./out/qwen3_vl_eval_suite/llava_bench/qwen3_vl_4b.llama3_2_3b.summary.json
     judge:
       enabled: true
       model_id: meta-llama/Llama-3.2-3B-Instruct
```
Torrero, line 87: We use `unsloth/Llama-3.2-3B-Instruct` as judge model, maybe we can set it up as default model.

Torrero (approved):
Overall LGTM

tico/quantization/examples/configs/qwen3_vl_eval_suite.yaml
mhs4670go: Yeah, actually I thought it should be official one. But, it would be better to use what we often use. I'll modify it in a follow-up PR.

PR #834, opened 2026-07-21
[quantization] Add token embedding, vision prefill, and mm_fusion verification steps in Gemma4 static runtime

dvsav:

## What

This PR adds three new side-by-side verification steps to the `StaticGemma4Runtime` for the Gemma4 E2B model: `token_embedding`, `vision_prefill`, and `mm_fusion`. Each step independently validates a runtime adapter's output against the corresponding HF `transformers` reference implementation, ensuring numerical equivalence before the static runtime is used for actual inference. It also refactors all `verify_step_*` functions to return their computed results, eliminating redundant recomputation across steps.

## Why

The `StaticGemma4Runtime` is being built incrementally, one workflow step at a time. Before proceeding to prefill, decode, and generation, we need confidence that each upstream component (token embedding, vision tower, multimodal fusion) produces identical results to the HF reference. Side-by-side validation catches subtle bugs early — such as dtype mismatches, incorrect embedding scaling, or wrong fusion semantics — rather than debugging them through the full prefill pipeline. The refactoring to return results (instead of `bool`) eliminates redundant `token_embedding` and `vision_prefill` calls that were previously duplicated in the `mm_fusion` step.

## Key Design Decisions

1. **Return computed results, not `bool`**: Since verification failures raise exceptions (`assert_close`, `raise ValueError`), returning `True` was meaningless. Each `verify_step_*` now returns its runtime-computed tensor (or batch dict), allowing downstream steps to reuse it without recomputation.

2. **`vision_prefill` uses PEIR, not exact match, for HF comparison**: The runtime's vision adapter wraps a quantized vision tower (with fake Q-DQ ops), so exact equality against the FP HF model is not expected. PEIR (Peak-Error-to-Interval Ratio) quantifies the deviation. An additional exact-match assertion against the quantized model's `get_image_features` ensures the adapter itself introduces no error beyond quantization.

3. **`mm_fusion` compares `fixed_slot_fuse` vs HF `masked_scatter`**: The runtime uses a static `torch.cat`-based fusion (fixed contiguous slot), while HF uses dynamic `masked_scatter`. The verification re-runs the processor to recover raw `input_ids` with image tokens, builds the image mask, and applies `masked_scatter` to confirm both paths produce identical fused embeddings.

4. **`pixel_values` cast to model dtype**: The HF processor outputs float32, but the quantized vision tower weights are BFloat16. The `vision_prefill` step casts `pixel_values` to `model.dtype` before passing to both runtime and HF reference sides.

5. **`mm_fusion` accepts `text_embeds` and `visual_embeds` as parameters** rather than recomputing them internally, following the "reuse results" principle.

## Changes

- `tico/quantization/recipes/debug/static_gemma4_runtime.py`
  - `verify_step_build_static_inputs`: Changed return type from `bool` to `dict[str, torch.Tensor]`, returns the `batch` dict.
  - Added `verify_step_token_embedding`: Validates `Gemma4TokenEmbeddingExportAdapter` output against HF `Gemma4TextScaledWordEmbedding`, including a sanity check that `sqrt(hidden_size)` embedding scale is applied. Returns `text_embeds` tensor.
  - Added `verify_step_vision_prefill`: Validates `Gemma4VisionPrefillExportAdapter` output against HF `model.get_image_features().pooler_output` (with PEIR metric) and quantized model's `get_image_features` (exact match). Casts `pixel_values` to model dtype. Returns `visual_embeds` tensor.
  - Added `verify_step_mm_fusion`: Validates `Gemma4MMFusionExportAdapter` (fixed_slot_fuse) against HF `masked_scatter` reference. Accepts `text_embeds` and `visual_embeds` as parameters. Returns `fused_embeds` tensor.
  - `run_static_gemma4_runtime`: Updated to chain return values — `batch` from Step 1, `text_embeds` from Step 2, `visual_embeds` from Step 3, passed to Step 4. Removed redundant `build_static_inputs` call.

## Tests

- All 11 existing tests in `test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py` continue to pass (tests `_normalize_valid_token_mask` and `_validate_padding_layout` helper functions).
- Manual end-to-end verification via the inspector script confirms all 4 steps pass:
  - Step 1 (`build_static_inputs`): All checks passed.
  - Step 2 (`token_embedding`): All checks passed.
  - Step 3 (`vision_prefill`): PEIR = 7.142857e-03, all checks passed.
  - Step 4 (`mm_fusion`): All checks passed.

## Example Script

```bash
$ python ./tico/quantization/examples/inspector.py \
Loading weights: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 1951/1951 [00:00<00:00, 4382.96it/s]
[verify_step_vision_prefill] PEIR = 7.142857e-03
[run_static_gemma4_runtime] SKIP: prefill (not implemented in this PR)
[run_static_gemma4_runtime] SKIP: decode (not implemented in this PR)
```

### DIFF
--- a/tico/quantization/recipes/debug/static_gemma4_runtime.py
+++ b/tico/quantization/recipes/debug/static_gemma4_runtime.py
@@ -41,7 +41,6 @@ from tico.quantization.wrapq.wrappers.gemma4.utils import (
     StaticGemma4Layout,
 )
 
-
 # =============================================================================
 # CPU Helper Functions (pure Python, no model needed)
 # =============================================================================
@@ -482,7 +481,7 @@ def verify_step_build_static_inputs(
     runtime: StaticGemma4Runtime,
     prompt: str,
     image,
-) -> bool:
+) -> dict[str, torch.Tensor]:
     """Side-by-side validation of ``build_static_inputs`` against HF reference.
 
     This function re-derives each sub-step of ``build_static_inputs`` using the
@@ -496,8 +495,9 @@ def verify_step_build_static_inputs(
     5. ``valid_length`` — correct unpadded sequence length
     6. ``padding`` — right-padded layout with pad_token_id fill
 
-    Returns ``True`` if all checks pass.
+    Returns the batch dict from ``build_static_inputs``.
     """
+
     import torch.testing
 
     layout = runtime.layout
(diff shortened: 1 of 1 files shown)


tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -587,15 +588,244 @@ def verify_step_build_static_inputs(
...
+        raise ValueError(
+            "vision_prefill shape mismatch: "
+            f"runtime={tuple(rt_visual_embeds.shape)}, "
+            f"reference={tuple(hf_visual_embeds.shape)}"
+        )
+
+    # --- PEIR (Peak-Error-to-Interval Ratio) ---
+    from tico.quantization.evaluation.metric import compute_peir
+
+    peir = compute_peir(rt_visual_embeds, hf_visual_embeds)
```
mhs4670go: The first argument should be reference value: hf_visual_embeds.

tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -587,15 +588,244 @@ def verify_step_build_static_inputs(
...
+    runtime: StaticGemma4Runtime,
+    batch: dict[str, torch.Tensor],
+) -> torch.Tensor:
+    """Side-by-side validation of ``vision_prefill`` against HF reference.
+
+    The runtime's ``Gemma4VisionPrefillExportAdapter`` runs the vision tower
+    followed by the ``embed_vision`` projection.  This function feeds
+    ``pixel_values`` and ``image_position_ids`` through both the runtime
+    adapter and HF's ``model.get_image_features(...)`` and asserts that the
+    projected visual soft tokens match.
```
Torrero: There is mismatch in the description. 
Actually, you test this assertion between runtime and quantized model not HF model.

tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -587,15 +588,244 @@ def verify_step_build_static_inputs(
...
+    image_mask = raw_input_ids == image_token_id  # (1, seq_len_raw)
+
+    # Pad the mask to max_seq to match the runtime's static shape
+    max_seq = layout.max_seq
+    seq_len_raw = raw_input_ids.shape[1]
+    if seq_len_raw > max_seq:
+        raise ValueError(f"Raw sequence length {seq_len_raw} exceeds max_seq {max_seq}")
+
+    padded_mask = torch.zeros((1, max_seq), dtype=torch.bool, device=runtime.device)
+    padded_mask[:, :seq_len_raw] = image_mask
```
Torrero, line 830: This verification does not use `visual_start_idx` and `num_visual_tokens`,  this differs from `runtime.mm_fusion`.

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: 👍 Thank you for noticing! The docstring was updated to accurately describe the two-tier verification - PEIR (informational) against the HF FP model, and `assert_close` (hard assertion) against the quantized model - rather than claiming the function asserts exact match against HF.

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: 👍 Added an explicit check that the image token positions in `raw_input_ids` match the static layout (`visual_start_idx` and `num_visual_tokens`). Previously, the `masked_scatter` reference path implicitly assumed the image tokens were contiguous at `visual_start_idx` (validated only by Step 1's `build_static_inputs`). Without this check, a layout mismatch would produce a confusing "mm_fusion mismatch" assertion error instead of a clear diagnostic. The new check raises a descriptive `ValueError` if the start position or token count doesn't match the static layout.

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: 👍 Thanks for catching that! Fixed the `compute_peir` argument order in `verify_step_vision_prefill`. The `compute_peir(base, target)` function is asymmetric — `base` is the reference used for both the numerator and the interval denominator, while `target` is the value being compared (used only in the numerator). The original code had the arguments swapped (`compute_peir(rt_visual_embeds, hf_visual_embeds)`), which incorrectly computed the interval denominator from the quantized runtime output instead of the HF FP reference. The fix swaps them to `compute_peir(hf_visual_embeds, rt_visual_embeds)`, ensuring the interval is computed from the FP reference as the metric's design intends.

Torrero (approved):
LGTM, thank you!

mhs4670go (approved):
LGTM

PR #836, opened 2026-07-22
[quantization] Implement prefill masks and RoPE verification in StaticGemma4Runtime

dvsav:

## What

Implements real per-layer-type attention masks and RoPE (Rotary Position Embedding) generation in `StaticGemma4Runtime`, replacing the previous placeholder implementation. Adds **Step 5** of the side-by-side verification flow (`verify_step_masks_and_rope`) that validates the runtime's masks and RoPE against Hugging Face `transformers` reference outputs.

## Why

The static Gemma4 runtime requires CPU-owned, pre-computed attention masks and RoPE tensors that will be fed to NPU-executed decoder layers. Gemma4 E2B has a **mixed layer-type architecture** — a repeating pattern of 5 sliding-attention layers followed by 1 full-attention layer — each requiring distinct mask patterns and distinct RoPE configurations:

- **`full_attention`**: proportional RoPE (`partial_rotary_factor=0.25`, `global_head_dim=512`, `rope_theta=1_000_000`) with a standard causal mask.
- **`sliding_attention`**: default RoPE (`head_dim=256`, `rope_theta=10_000`) with a sliding-window causal mask (`sliding_window=512`).

The previous `build_prefill_masks_and_rope` returned shape-compatible zero/one placeholders. This PR replaces them with correct implementations and adds verification to ensure they exactly match HF's `Gemma4TextRotaryEmbedding` and `create_masks_for_generate`.

## Key Design Decisions

1. **Per-layer-type RoPE via HF module**: The runtime calls the model's own `Gemma4TextRotaryEmbedding` module (`self.model.model.language_model.rotary_emb`) directly, passing the `layer_type` argument. This guarantees the runtime uses the exact same `inv_freq` buffers and `attention_scaling` that the HF model uses, avoiding reimplementation errors.

2. **Key-only padding masking (matching HF convention)**: Both mask builders (`_build_full_attention_mask` and `_build_sliding_window_attention_mask`) apply padding only via `key_valid` (column masking). Query rows for padding positions are **not** zeroed out — they retain their causal/sliding-window pattern. This matches HF's `create_masks_for_generate`, which only controls which **keys** are visible, not which **queries** are active. Padding query outputs are simply discarded downstream. An earlier version incorrectly included `query_valid`, which truncated padding rows to all-False and caused a pattern mismatch.

3. **Boolean pattern comparison for masks**: HF may use different `mask_value` representations or return `None` depending on the attention implementation. The verification compares **boolean attention patterns** (allowed vs. forbidden) rather than exact float values, and temporarily forces `_attn_implementation="eager"` to obtain 4D float masks from HF's `create_masks_for_generate`.

4. **Exact match for RoPE**: Since the runtime and HF reference use the *same* module, RoPE comparison is an exact `torch.testing.assert_close` check.

## Changes

All changes are in **`tico/quantization/recipes/debug/static_gemma4_runtime.py`** (293 insertions, 29 deletions):

- **Added `_build_full_attention_mask`**: Builds a standard causal mask with key-only padding masking. Returns `(B, 1, S, S)` additive bias.
- **Added `_build_sliding_window_attention_mask`**: Builds a sliding-window causal mask (`kv_idx > q_idx - sliding_window`) with key-only padding masking. Returns `(B, 1, S, S)` additive bias.
- **Replaced placeholder `build_prefill_masks_and_rope`**: Now computes `position_ids = arange(seq_len)`, builds both mask types from the text config's `sliding_window`, and computes per-layer-type RoPE via the HF `rotary_emb` module. Returns `(attention_masks, position_embeddings)` dicts keyed by layer type.
- **Added `verify_step_masks_and_rope`**: Side-by-side verification comparing runtime RoPE (exact match) and masks (boolean pattern match) against HF reference. Returns `(rt_masks, rt_rope)` to avoid recomputation in downstream steps.
- **Updated `run_static_gemma4_runtime`**: Calls Step 5 after Step 4.

## Tests

No new unit test files in this PR — verification is done via the example script's side-by-side comparison flow. All 5 verification steps pass:

```
[verify_step_vision_prefill] PEIR = 7.142857e-03
```

## Example Script

```bash
python -m tico.quantization.recipes.debug.static_gemma4_runtime \
  --config tico/quantization/examples/configs/static_gemma4_runtime.yaml
```

### DIFF
--- a/tico/quantization/recipes/debug/static_gemma4_runtime.py
+++ b/tico/quantization/recipes/debug/static_gemma4_runtime.py
@@ -97,6 +97,90 @@ def _validate_padding_layout(
                 raise ValueError("Right padding expected but not found")
 
 
+def _build_full_attention_mask(
+    valid_token_mask: torch.Tensor,
+    device: torch.device,
+    dtype: torch.dtype,
+    mask_value: float,
+) -> torch.Tensor:
+    """Build a standard causal attention mask with padding.
+
+    Args:
+        valid_token_mask: Boolean mask of shape ``(B, S)`` — True for real
+            tokens, False for padding.
+        device: Target device.
+        dtype: Target dtype for the output mask.
+        mask_value: Value used to mask forbidden positions (typically
+            ``torch.finfo(dtype).min``).
+
+    Returns:
+        Additive attention bias of shape ``(B, 1, S, S)``.  ``0.0`` allows
+        attention, ``mask_value`` forbids it.
+    """
+    batch_size, seq_len = valid_token_mask.shape
+    # Causal: kv_idx <= q_idx
+    causal = torch.tril(
+        torch.ones(seq_len, seq_len, device=device, dtype=torch.bool)
+    )  # (S, S)
+    causal = causal.unsqueeze(0).expand(batch_size, -1, -1)  # (B, S, S)
+
+    key_valid = valid_token_mask.to(device).unsqueeze(1).expand(-1, seq_len, -1)
+
+    valid = causal & key_valid
+    mask = torch.zeros(batch_size, seq_len, seq_len, device=device, dtype=dtype)
+    mask = mask.masked_fill(~valid, mask_value)
+    return mask.unsqueeze(1)  # (B, 1, S, S)
+
+
+def _build_sliding_window_attention_mask(
+    valid_token_mask: torch.Tensor,
+    sliding_window: int,
+    device: torch.device,
+    dtype: torch.dtype,
+    mask_value: float,
+) -> torch.Tensor:
+    """Build a sliding-window causal attention mask with padding.
+
+    A token at position ``q`` can attend to position ``k`` if:
+    - ``k <= q`` (causal), AND
+    - ``k > q - sliding_window`` (within the sliding window)
+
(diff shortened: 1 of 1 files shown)


mhs4670go (approved):
LGTM

tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -848,6 +968,146 @@ def verify_step_mm_fusion(
...
+        torch.testing.assert_close(
+            rt_cos,
+            ref_cos,
+            msg=f"RoPE cos mismatch for layer_type={layer_type!r}",
+        )
+        torch.testing.assert_close(
+            rt_sin,
+            ref_sin,
+            msg=f"RoPE sin mismatch for layer_type={layer_type!r}",
+        )
```
Torrero, line 1040: @dvsav 
This comparison is unclear for me, in both cases you use HF's module `rotary_emb` with almost the same artificial `dummy_hidden` input.  Could you please clarify your idea, or maybe I missed something.

dvsav: 👍 I agree, but, I'll fix that in the following PR, if you don't mind.

PR #838, opened 2026-07-23
[serialize] Remove redundant empty buffer

mhs4670go:
This commit removes redundant empty buffer.

Circle always use index-0 buffer for empty buffers. Currently, the tensor creation api always creates a empty buffer, which makes redundant buffers.

### DIFF
--- a/test/unit_test/serialize/operator/fixture.py
+++ b/test/unit_test/serialize/operator/fixture.py
@@ -12,8 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import torch.fx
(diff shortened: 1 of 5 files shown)


shs-park (approved):
LGTM!
=)

_Reviewed by AI via personal agents_

llFreetimell:
As I said offline, buffer[0] is special buffer in ONERT.
(https://github.com/Samsung/ONE/issues/7256#issuecomment-882406047)

If we don't use ONERT, this PR seems okay :)
Unfortunately, I have no authority to ensure that :(

PR #843, opened 2026-07-27
[quantization] Add static decode step for StaticGemma4Runtime

dvsav:

## What

This PR implements the **decode step** for the `StaticGemma4Runtime` in the TICO project, enabling side-by-side validation of single-token autoregressive decoding against Hugging Face's reference implementation for Gemma4 E2B models.

## Why

The static-shape NPU inference flow requires separate prefill and decode paths. While prefill was previously implemented, decode was a skeleton placeholder. This PR adds the full decode implementation to:
- Validate KV cache management across decode steps
- Verify per-layer-type masks and RoPE at decode positions
- Ensure shared-KV bookkeeping works correctly for consumer layers
- Enable end-to-end smoke testing of the static runtime

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| **Slice KV cache to `past_len` before passing to decode layers** | The full `max_seq`-sized cache buffer contains stale padding data. Passing only valid tokens (0..past_len) ensures the attention wrapper's capacity check passes and avoids carrying garbage through attention computation. |
| **Shared-KV store layers write `past_len + 1` entries** | After writing the new token's K/V delta at position `past_len`, the valid range is 0..past_len inclusive. Consumer layers need exactly this range to compute correct attention scores. |
| **Reuse `verify_step_prefill` return value** | The `verify_step_decode` function accepts `prefill_logits` as a parameter to avoid redundant prefill computation, reducing runtime while maintaining test isolation. |
| **HF reference uses `DynamicCache`** | Mirrors real-world HF usage patterns and ensures the runtime's cache management produces equivalent results to HF's standard cache implementation. |
| **Per-layer-type decode masks and RoPE** | Gemma4 has alternating "full_attention" and "sliding_attention" layers with different head dimensions and attention patterns. The decode path mirrors the prefill design for consistency. |
| **Add `same_token_mode` for multi-step decode** | When quantization causes different argmax tokens between runtime and HF, multi-step comparisons become invalid. The `same_token_mode=True` parameter forces both paths to use HF reference tokens, enabling isolated quantization error measurement. |
| **Prefill divergence auto-detection** | If runtime and HF argmax differ at step 0, multi-step verification is automatically limited to 1 step with a warning, preventing meaningless comparisons. |

## Changes

**`tico/quantization/recipes/debug/static_gemma4_runtime.py`**

- **`build_decode_masks_and_rope`**: Replaced placeholder with full implementation generating per-layer-type full/sliding decode masks and RoPE tensors at the current decode position
- **`decode_one`**: Added PLE (Per-Layer Embeddings) support, shared-KV bookkeeping, and proper KV cache slicing (`past_len` instead of full `max_seq`)
- **`verify_step_decode`**: 
  - New validation function comparing runtime decode logits against HF's `DynamicCache` reference with PEIR metrics
  - Added `same_token_mode: bool = False` parameter for controlled multi-step comparison
  - Added prefill divergence detection with auto-limit to 1 step
  - Added per-step divergence warnings
- **`run_static_gemma4_runtime`**: Integrated step 7 (decode verification) into the smoke flow, reusing prefill logits

**`test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py`**

- **`TestBuildDecodeMasksAndRope`**: New test class with 6 tests for decode mask generation:
  - `test_full_attention_mask_shape`: Verifies mask shape is `(batch_size, 1, max_seq)`
  - `test_full_attention_mask_values`: Verifies positions `0..past_len` are `0.0` (allowed)
  - `test_sliding_window_mask_boundary_early_decode`: Verifies window covers `[0, past_len]` when `past_len < sliding_window`
  - `test_sliding_window_mask_boundary_late_decode`: Verifies window covers `[past_len-sw+1, past_len]` when `past_len >= sliding_window`
  - `test_sliding_window_mask_first_step`: Verifies at `past_len=0`, only position 0 is allowed
  - `test_rope_computed_at_past_len`: Verifies RoPE is computed at the correct decode position

## Tests

```
$ python -m pytest test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py -v
=============================================================== test session starts ================================================================
cachedir: .pytest_cache
configfile: pyproject.toml
plugins: anyio-4.13.0
collected 21 items                                                                                                                                 

test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestNormalizeValidTokenMask::test_batched_input PASSED          [  4%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestNormalizeValidTokenMask::test_shape_mismatch_raises PASSED  [  9%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestNormalizeValidTokenMask::test_with_attention_mask PASSED    [ 14%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestNormalizeValidTokenMask::test_without_attention_mask_no_pad_token_id PASSED [ 19%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestNormalizeValidTokenMask::test_without_attention_mask_uses_pad_token_id PASSED [ 23%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_batched_right_padding_invalid PASSED [ 28%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_batched_right_padding_valid PASSED [ 33%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_right_padding_invalid PASSED    [ 38%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_right_padding_no_padding PASSED [ 42%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_right_padding_valid PASSED      [ 47%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestValidatePaddingLayout::test_unsupported_padding_side_raises PASSED [ 52%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestAllocateEmptyCache::test_all_full_attention_layers PASSED   [ 57%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestAllocateEmptyCache::test_all_sliding_layers PASSED          [ 61%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestAllocateEmptyCache::test_no_global_head_dim_falls_back PASSED [ 66%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestAllocateEmptyCache::test_per_layer_type_head_dim PASSED     [ 71%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_full_attention_mask_shape PASSED [ 76%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_full_attention_mask_values PASSED [ 80%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_rope_computed_at_past_len PASSED [ 85%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_sliding_window_mask_boundary_early_decode PASSED [ 90%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_sliding_window_mask_boundary_late_decode PASSED [ 95%]
test/quantization/recipes/integration/test_static_gemma4_runtime_helpers.py::TestBuildDecodeMasksAndRope::test_sliding_window_mask_first_step PASSED [100%]

========================================================== 21 passed, 2 warnings in 7.43s ==========================================================
```

# Example Script

```
$ python -m pdb ./tico/quantization/examples/inspector.py \

Loading weights: 100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████| 1951/1951 [00:00<00:00, 5027.45it/s]
[verify_step_vision_prefill] PEIR = 0.714286 %
[run_static_gemma4_runtime] Step 6: verify prefill
[verify_step_prefill] Runtime logits shape: (1, 262144)
[verify_step_prefill] Reference logits shape: (1, 262144)
[verify_step_prefill] mean|diff| = 0.05616927
[verify_step_prefill]  max|diff| = 0.50000000
[verify_step_prefill] PEIR       = 2.285714 %
[verify_step_prefill] Runtime argmax token: 106
[verify_step_prefill] Reference argmax token: 106
[verify_step_prefill] Argmax tokens MATCH.
[verify_step_prefill] All checks passed.
[run_static_gemma4_runtime] Step 7: verify decode
[verify_step_decode] Step 1:
[verify_step_decode]   Runtime logits shape: (1, 262144)
[verify_step_decode]   mean|diff| = 0.17796150
[verify_step_decode]    max|diff| = 0.75000000
[verify_step_decode]   PEIR       = 2.922374 %
[verify_step_decode]   Runtime argmax token: 106
[verify_step_decode]   Reference argmax token: 106
[verify_step_decode]   Argmax tokens MATCH.
[verify_step_decode] All checks passed.
```

### DIFF
--- a/tico/quantization/recipes/debug/static_gemma4_runtime.py
+++ b/tico/quantization/recipes/debug/static_gemma4_runtime.py
@@ -629,55 +629,137 @@ class StaticGemma4Runtime:
     ) -> tuple[dict[str, torch.Tensor], dict[str, tuple[torch.Tensor, torch.Tensor]]]:
         """Build CPU-owned static masks and RoPE tensors for one decode step.
 
-        This skeleton returns shape-compatible placeholder tensors so the runtime
-        orchestration can be linted and extended independently. The final
-        implementation should create distinct full/sliding decode masks and
-        position-specific RoPE slices for each Gemma4 layer type.
+        Produces per-layer-type decode masks and RoPE:
+
+        - **Full attention mask**: ``(B, 1, max_seq)`` additive bias.
+          Positions ``0..past_len`` are ``0.0`` (allowed), the rest are
+          ``mask_value`` (forbidden).
+
+        - **Sliding attention mask**: ``(B, 1, max_seq)`` additive bias.
+          Only positions within the sliding window
+          ``[max(0, past_len - sliding_window + 1), past_len]`` are ``0.0``.
+
+        - **RoPE**: ``(cos, sin)`` computed via the HF model's
+          ``Gemma4TextRotaryEmbedding`` at the current decode position
+          ``past_len``, per layer type.
         """
-        mask = build_decode_attention_mask(
+        max_seq = self.layout.max_seq
+        mask_value = torch.finfo(dtype).min
+
+        # Full attention decode mask: attend to all past + current token
+        full_mask = build_decode_attention_mask(
             batch_size=batch_size,
             past_len=self.past_len,
-            max_seq=self.layout.max_seq,
+            max_seq=max_seq,
             device=self.device,
             dtype=dtype,
-            mask_value=-120.0,
+            mask_value=mask_value,
         )
-        head_dim = int(
-            getattr(
-                self.text_config,
-                "head_dim",
-                self.text_config.hidden_size // self.text_config.num_attention_heads,
-            )
+
+        # Sliding attention decode mask: only attend within the sliding window
+        sliding_window = int(getattr(self.text_config, "sliding_window", 1024))
+        sliding_mask = torch.full(
+            (batch_size, 1, max_seq), float(mask_value), device=self.device, dtype=dtype
         )
-        cos = torch.ones(batch_size, 1, head_dim, device=self.device, dtype=dtype)
-        sin = torch.zeros_like(cos)
+        # Allowed positions: max(0, past_len - sliding_window + 1) .. past_len
+        window_start = max(0, self.past_len - sliding_window + 1)
+        sliding_mask[:, :, window_start : self.past_len + 1] = 0.0
+
+        attention_masks = {
+            "full_attention": full_mask,
+            "sliding_attention": sliding_mask,
+        }
+
+        # RoPE at the current decode position
+        rotary_emb = self.model.model.language_model.rotary_emb
+        position_ids = torch.tensor(
+            [[self.past_len]], device=self.device, dtype=torch.long
+        )  # (1, 1)
+        dummy_hidden = torch.zeros(batch_size, 1, 1, device=self.device, dtype=dtype)
 
         layer_types = set(getattr(self.text_config, "layer_types", ["full_attention"]))
-        attention_masks = {layer_type: mask for layer_type in layer_types}
-        position_embeddings = {layer_type: (cos, sin) for layer_type in layer_types}
+        position_embeddings: dict[str, tuple[torch.Tensor, torch.Tensor]] = {}
+        for layer_type in layer_types:
+            cos, sin = rotary_emb(dummy_hidden, position_ids, layer_type)
+            position_embeddings[layer_type] = (cos, sin)
+
         return attention_masks, position_embeddings
 
     @torch.no_grad()
     def decode_one(self, input_ids: torch.Tensor) -> torch.Tensor:
-        """Run one static decode step and return next-token logits."""
-        hidden_states = self.token_embedding(input_ids.to(self.device))
+        """Run one static decode step and return next-token logits.
+
+        Handles PLE, per-layer-type masks/RoPE, and shared-KV bookkeeping,
+        mirroring the prefill path for a single-token decode step.
+        """
+        llm_input_ids = input_ids.to(self.device)
+        hidden_states = self.token_embedding(llm_input_ids)
+
         attention_masks, position_embeddings = self.build_decode_masks_and_rope(
             batch_size=hidden_states.shape[0],
             dtype=hidden_states.dtype,
         )
 
+        # Compute PLE for the single decode token if enabled
+        per_layer_inputs = None
+        if self.text_model.hidden_size_per_layer_input:
+            ple_token = self.text_model.get_per_layer_inputs(llm_input_ids, None)
+            per_layer_inputs = self.text_model.project_per_layer_inputs(
+                hidden_states, ple_token
+            )
+
+        # Shared-KV bookkeeping: store layers' full K/V (including the new
+        # decode token) are passed to consumer layers.  We maintain a dict
+        # keyed by layer_type, updated after each store layer writes its delta.
+        shared_kv_states: dict[str, tuple[torch.Tensor, torch.Tensor]] = {}
+
         for layer_idx, layer in enumerate(self.decode_layers):
-            cache = self.layer_caches[layer_idx]
             layer_type = self.text_config.layer_types[layer_idx]
+            per_layer_input = (
+                per_layer_inputs[:, :, layer_idx, :]
+                if per_layer_inputs is not None
+                else None
+            )
+
+            # Check if this layer is a shared-KV consumer
+            attn = layer.wrapped.self_attn.wrapped
+            is_shared = bool(getattr(attn, "is_kv_shared_layer", False))
+            shared_kv = shared_kv_states.get(layer_type, None) if is_shared else None
+
+            cache = self.layer_caches[layer_idx]
+
             out = layer(
                 hidden_states=hidden_states,
                 attention_mask=attention_masks[layer_type],
                 position_embeddings=position_embeddings[layer_type],
-                past_key_value=(cache.past_k, cache.past_v),
+                past_key_value=(
+                    (
+                        cache.past_k[:, :, : self.past_len, :],
+                        cache.past_v[:, :, : self.past_len, :],
+                    )
+                    if not is_shared
+                    else None
+                ),
+                per_layer_input=per_layer_input,
+                shared_key_value=shared_kv,
             )
-            hidden_states, new_k, new_v = out
-            cache.past_k[:, :, self.past_len : self.past_len + 1, :] = new_k
-            cache.past_v[:, :, self.past_len : self.past_len + 1, :] = new_v
+
+            if isinstance(out, tuple) and len(out) == 3:
+                hidden_states, new_k, new_v = out
+                # Write single-token K/V delta into the fixed cache
+                cache.past_k[:, :, self.past_len : self.past_len + 1, :].copy_(new_k)
+                cache.past_v[:, :, self.past_len : self.past_len + 1, :].copy_(new_v)
+                # Update shared_kv_states for consumer layers.
+                # The valid K/V (prefill + all decode deltas so far) occupies
+                # positions 0..past_len inclusive in the cache.  Slice to that
+                # range so consumer layers see the correct k_len = past_len + 1.
+                if bool(getattr(attn, "store_full_length_kv", False)):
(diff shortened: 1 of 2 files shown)


tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -629,55 +629,137 @@ def build_decode_masks_and_rope(
...
+
             out = layer(
                 hidden_states=hidden_states,
                 attention_mask=attention_masks[layer_type],
                 position_embeddings=position_embeddings[layer_type],
-                past_key_value=(cache.past_k, cache.past_v),
+                past_key_value=(
+                    (
+                        cache.past_k[:, :, : self.past_len, :],
+                        cache.past_v[:, :, : self.past_len, :],
```
mhs4670go, line 738: Please use a fixed past cache, for example (B, Hkv, max_seq - 1, D) as in `StaticLlamaRuntime`, compact valid prefill K/V into its prefix, and pass the full buffer on every decode invocation.

tico/quantization/recipes/debug/static_gemma4_runtime.py
```diff
@@ -1381,7 +1647,12 @@ def run_static_gemma4_runtime(cfg: StaticGemma4RuntimeConfig) -> None:
     print("[run_static_gemma4_runtime] Step 6: verify prefill")
     rt_logits = verify_step_prefill(runtime, batch, cfg.prompt, image)
 
-    # --- Steps 7+: decode / generation (not yet implemented) ---
-    print("[run_static_gemma4_runtime] SKIP: decode (not implemented in this PR)")
+    # --- Step 7: decode validation ---
+    print("[run_static_gemma4_runtime] Step 7: verify decode")
+    rt_decode_logits = verify_step_decode(
+        runtime, batch, cfg.prompt, image, prefill_logits=rt_logits, num_decode_steps=1
```
mhs4670go:
```suggestion
        runtime, batch, cfg.prompt, image, prefill_logits=rt_logits, num_decode_steps=cfg.verify_steps,
```

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: > Please use a fixed past cache, for example (B, Hkv, max_seq - 1, D) as in `StaticLlamaRuntime`, compact valid prefill K/V into its prefix, and pass the full buffer on every decode invocation.

Thank you for the suggestion, but... I've investigated adopting the `StaticLlamaRuntime` pattern for `StaticGemma4Runtime.decode_one`, and found that it requires changes beyond the scope of this PR.

**The Issue:**

The Gemma4 text attention wrapper (`quant_text_attention.py`) performs a capacity check in `_build_attention_mask` that raises when `k_len` exceeds the `causal_mask_template` size:

```python
if k_len > self.causal_mask_template.size(-1):
    raise RuntimeError(
        f"Key length exceeds causal mask capacity: k_len={k_len}, capacity={self.causal_mask_template.size(-1)}."
    )
```

When passing the full cache buffer (2048) and concatenating with the new token's K/V, we get `k_len = 2049`, which triggers this error.

**Why StaticLlamaRuntime Works:**

`StaticLlamaRuntime` uses `max_seq - 1` cache capacity AND processes prefill via chunked `append_prefill`, so the cache never contains more than `max_seq - 1` valid tokens. The decode concatenation produces `k_len <= max_seq - 1`, which passes the capacity check.

**StaticGemma4Runtime Difference:**

Gemma4 uses single-shot prefill that returns full `max_seq` K/V tensors. Passing this full buffer to decode and concatenating with the new token produces `k_len = max_seq + 1`, exceeding capacity.

**Required Changes for StaticLlamaRuntime Pattern:**

To support passing the full buffer while only using valid positions, we could do the following:

1. **`quant_text_attention.py`**: Add `cache_valid_length` parameter to `_build_present_key_value` to slice past K/V before concatenation
2. **`quant_text_decoder_layer.py`**: Forward `cache_valid_length` through the decoder layer
3. **`export_adapters.py`**: Update `Gemma4TextDecoderLayerDecodeExportAdapter.forward` to accept and forward `cache_valid_length`

These changes touch core wrappers and, I think, require a separate, focused PR.

**Current Approach:**

For this PR, `decode_one` slices the cache to `past_len` before passing:
```python
past_key_value=(
    cache.past_k[:, :, :self.past_len, :],
    cache.past_v[:, :, :self.past_len, :],
)
```

This ensures `k_len = past_len + 1 <= max_seq`, passing the capacity check while the attention mask (which is sliced to match `k_len` by `_normalize_attention_mask_shape`) correctly indicates valid positions.

**Future Work:**

I can file a follow-up issue to add `cache_valid_length` support to the Gemma4 attention wrapper infrastructure, enabling the cleaner StaticLlamaRuntime pattern for StaticGemma4Runtime. What do you think?

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: 👍 done

tico/quantization/recipes/debug/static_gemma4_runtime.py
mhs4670go: I think there is a misunderstanding about the proposed cache shape.

The suggestion was not to pass the current `max_seq`-sized cache buffer in full. The persistent past cache itself should be allocated with capacity `max_seq - 1`.

With `max_seq = 2048`, the decode shapes would therefore be:

```
past_key_value:  (B, Hkv, 2047, D)
new_key_value:   (B, Hkv,    1, D)
present KV:      (B, Hkv, 2048, D)
```

Therefore, the concatenation does not produce `k_len = 2049`; it produces exactly `k_len = max_seq`, which satisfies the existing capacity check. No `cache_valid_length` argument or wrapper change is needed.

The single-shot prefill output may still have sequence length `max_seq`. The runtime should copy only the valid prompt K/V range into the prefix of the `max_seq - 1` persistent cache, as `StaticLlamaRuntime.prefill()` already
does. `StaticLlamaRuntime` does not require chunked append-prefill for its initial prefill; `append_prefill` is a separate multi-turn path.

For decode, the fixed physical layout is:

```
[past cache: max_seq - 1 slots][current token: 1 slot]
```

Unused past-cache slots are masked. The decode mask should therefore expose `[:past_len]` plus the final slot `[-1]`, rather than placing the current token at physical index `past_len`.

For a shared-KV store layer, the fixed-size shared K/V passed to consumer layers can similarly be constructed from the `max_seq - 1` past cache and the returned one-token K/V delta.

Slicing by a runtime `cache_valid_length` inside the exported wrapper would make the K dimension dynamic again, which is what the fixed-shape runtime should avoid.

If this is still difficult to follow, we can merge this PR as-is, and I can submit a follow-up PR implementing the fixed-shape cache design. Alternatively, you can implement it directly in this PR.

mhs4670go (approved):
LGTM

tico/quantization/recipes/debug/static_gemma4_runtime.py
dvsav: Ok, let me handle that in the next PR)

PR #844, opened 2026-07-27
[examples] Optimize static Llama last-token logit projection

mhs4670go:
Gather the final valid hidden state before the LM head in prefill and append-prefill paths so the runtime does not materialize full-sequence logits.


shs-park (approved):
LGTM!
=)

_Reviewed by AI via personal agents_

PR #851, opened 2026-08-02
[evaluation] [DRAFT] Gemma4 evaluation

Torrero:
This draft PR adds benchmarks evaluation support for Gemma4 with GPTQ quantizer.

Evaluation `google/gemma-4-e2b-it`

| Config ID| PPL | mmlu | vqav2 | COCO (CIDEr/Bleu_4) | hellaswag| MMMU_Pro (vision) | Llava-Bench Rel. | videomme (21 frames) |
|--|--|--|--|--|--|--|--|--|
| FP32                       |  in progress  |   0.484    | 0.738   | 0.355/0.123  | 0.381  | 0.05  | 81.63 |  49.1  |
| PTQ_W4A16_ple8bit   |  in progress  |      | 0.622   |   | 0.418  | 0.00  | 76.01 |    |
| GPTQ_MSE_W4A16_ple8bit   |  in progress  |  0.442     |  0.737  |  0.406/0.141 | 0.367   |  0.061  |  81.39  |  51.8  |

*For the instruction-tuned model (gemma-4-e2b-it) the perplexity evaluation in raw mode produces huge values (>1000), it should be evaluated in the chat-continuation mode, but its behavior requires clarification. (in progress).

[test_Gemma_orig_eval.log](https://github.com/user-attachments/files/30639259/test_Gemma_orig_eval.log)
[gemma_PTQ_W4A16_ple8bit.log](https://github.com/user-attachments/files/30639263/gemma_PTQ_W4A16_ple8bit.log)

[gemma_GPTQ_mse_W4A16_ple8bit.log](https://github.com/user-attachments/files/30837288/gemma_GPTQ_mse_W4A16_ple8bit.log)

<details> <summary>Run command for PTQ_W4A16_ple8bit </summary>

```
python -m tico.quantization.examples.quantize \
  --config tico/quantization/examples/configs/gemma4_e2b_gptq_ple8bit.yaml \
  --set model.name_or_path=google/gemma-4-e2b-it \
  --set calibration.seq_len=2048 \
  --set calibration.datasets.0.n_samples=10 \
  --set export.enabled=false \
  --set pipeline.0.enabled=false \
  --set pipeline.0.verbose=true \
  --set pipeline.0.weight_bits=4 \
  --set pipeline.0.percdamp=0.1 \
  --set pipeline.1.print_model=true \
  --set pipeline.1.enabled=true \
  --set pipeline.1.activation=int16 \
  --set pipeline.1.linear_weight=uint4 \
[...]

PR #852, opened 2026-08-03
[quantization] Add configs for gemma4

mhs4670go:
This commit adds configs for gemma4.

### DIFF
--- a/tico/quantization/recipes/adapters/gemma4.py
+++ b/tico/quantization/recipes/adapters/gemma4.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 from contextlib import contextmanager
+from pathlib import Path
 from typing import Any, Iterator, Mapping, Sequence
 
 import torch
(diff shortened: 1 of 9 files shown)


tico/quantization/recipes/adapters/gemma4.py
```diff
@@ -319,12 +358,97 @@ def evaluate(self, ctx: RecipeContext) -> None:
...
+                n_samples=int(mmmu.get("n_samples", -1)),
+                max_new_tokens=int(mmmu.get("max_new_tokens", 16)),
+                max_seq_len=max_seq_len,
+                temperature=float(mmmu.get("temperature", 0.0)),
+                verbose=bool(mmmu.get("verbose", verbose)),
+            )
+
+        ppl = eval_cfg.get("ppl", {})
+        if ppl.get("enabled", False):
+            ppl_value = evaluate_vlm_text_ppl(
```
Torrero, line 431: I mentioned in my draft that raw PPL evaluation produces huge PPL score for original instruction-tuned model `gemma-4-e2b-it` (https://github.com/Samsung/TICO/pull/851#issue-5044676988). 
As I understand correctly the `chat-continuation` mode should be used for this aim.

Torrero: We can support it in the next PRs

Torrero (approved):
LGTM, thank you!

PR #854, opened 2026-08-04
[quantization] Add identity quant spec for weights-only PTQ

Torrero:
This PR adds identity quant spec for weights-only PTQ.

The weights-only quantization simplifies the debugging process.


mhs4670go:
I think it's very good chage for debugging. But, how about renaming it to `no_quant` instead? I'll refactor the codes in the next PR.

mhs4670go (approved):
LGTM

PR #855, opened 2026-08-04
[quantization] fix QuantGemma4ClippableLinear module

Torrero:
This PR corrects `fp_name` for QuantGemma4ClippableLinear module and introduces `_clippable_linear_tree_override` in `gemma4_builders.py`.

### Why
- The `fp_name` passed to the inner `PTQWrapper` was wrong — it passed the parent's name (e.g., `"layers.0.self_attn.q_proj"`) instead of the child's qualified name (e.g., `"layers.0.self_attn.q_proj.linear"`).
- New `_clippable_linear_tree_override` resolves the problem when vision modules use `Gemma4ClippableLinear` (which wraps an inner `nn.Linear` as `.linear`), and `QuantGemma4ClippableLinear` scopes the child config via `qcfg.child("linear")`. As Example, the builder was placing weight overrides at `gate_proj.weight` instead of `gate_proj.linear.weight`, so `qcfg.child("linear")` found nothing. `test_builders.py` has been corrected accordingly.

### DIFF
--- a/test/quantization/config/test_builders.py
+++ b/test/quantization/config/test_builders.py
@@ -297,7 +297,7 @@ class TestBuildGemma4E2BPtqConfig(unittest.TestCase):
             DType.uint(4),
         )
         self.assertEqual(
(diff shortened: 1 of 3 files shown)


mhs4670go (approved):
LGTM

PR #893, opened 2026-08-13
[quantization] Adding quantizers to the top-level wrappers for Qwen3 & Gemma4

Torrero:
This PR adds quantizers to the top-level wrappers for Qwen3 & Gemma4

**Why**
This fix copies GPTQ quantizers to the top-level of PTQ wrapper, so that [`find_gptq_quantizers(q_model)`](https://github.com/Samsung/TICO/blob/6da61f4a89b02f6cdca5f684fed4ed50330c3cf8/tico/quantization/recipes/stages/ptq.py#L196) can find them at the top level of wrapper and injects it to the PTQ weight observers. Without this fix,  `find_gptq_quantizers` finds quantizers on the Original model not Wrapped model and as a consequence the  [`inject_gptq_qparams`](https://github.com/Samsung/TICO/blob/6da61f4a89b02f6cdca5f684fed4ed50330c3cf8/tico/quantization/recipes/stages/ptq.py#L198)  does not inject the qparams from GPTQ to PTQ observers for Qwen3 and Gemma4 models.

Co-Authored-By: Cline

### DIFF
--- a/tico/quantization/wrapq/wrappers/gemma4/quant_for_causal_lm.py
+++ b/tico/quantization/wrapq/wrappers/gemma4/quant_for_causal_lm.py
@@ -43,6 +43,11 @@ class QuantGemma4ForCausalLM(QuantModuleBase, GenerationMixin):
         super().__init__(qcfg, fp_name=fp_name)
         self.module = fp_model
         self.config = fp_model.config
+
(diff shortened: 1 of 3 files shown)


mhs4670go (approved):
LGTM

PR #897, opened 2026-08-14
[quantization] Refactor the way gptq quantizer is moved

mhs4670go:
This commit refactors the way gptq quantizer is moved.


mhs4670go:
@Torrero 

GPTQ quantizers may live on the original FP model, while the weight observers that need those qparams live in the prepared PTQ model tree. Therefore, the quantizer owner should only be used for discovery, and qparam injection should always traverse the prepared PTQ model. This also avoids requiring model-specific quantizer forwarding in each top-level wrapper.

PR #915, opened 2026-08-18
[quantization] Support GPTQv2 for Qwen3

Torrero:
**This PR supports GPTQv2 for Qwen3-VL model.**

__GPTQv2__ — an improved variant of GPTQ that corrects for upstream quantization error in the Hessian computation. Standard GPTQ computes the Hessian from activations that have already passed through previously quantized layers, introducing error. GPTQv2 addresses this by collecting "native" (full-precision) inputs from a pristine copy of the original model and computing a correction matrix (`dXXT`) that adjusts the weight updates accordingly.

Additionaly added the ability to __cache/save the inputs of the original (unquantized) model__ during GPTQv2 calibration and  __filtering calibration samples__ by their class/category.
 

#### How GPTQv2 Works

1. __Two forward passes__: The quantized model runs normally (collecting H from corrupted inputs). Simultaneously, a pristine copy of the model runs the same calibration data (collecting native FP inputs).

2. __dXXT computation__: For each layer, the difference between native and quantized inputs (`dX = native_inp - inp`) is cross-correlated with the quantized input: `dXXT += dX @ inp^T`. This captures how upstream quantization error correlates with the current layer's activations.

3. __P correction matrix__: During `fasterquant()`, the correction is applied: `P = alpha * (dXXT @ hinv^T).triu() @ hinv`. This adjusts weight updates so that the quantized weights compensate for the input error, rather than treating corrupted inputs as if they were clean.

4. __alpha parameter__: Controls correction strength. `alpha=0` → standard GPTQ. `alpha=0.25` (default) → mild correction, best empirical results. Higher values risk overcorrection.

Introduction of __`normalize_H`__ flag, which controls whether the GPTQ Hessian matrix `H` is accumulated as a __running average__ (normalized by `sqrt(2/N)`) or as a __raw sum__.

**Results of applying GPTQv2 for Qwen3-vl:**

| Config ID| PPL | mmlu | vqav2 | COCO (CIDEr/Bleu_4) | hellaswag | MMMU_Pro (vision) (under review)  | Llava-Bench Rel. | videomme (32 frames) |
|--|--|--|--|--|--|--|--|--|
| FP32                       | 9.46 | 0.735     | 0.895   | 0.335 / 0.079 | 0.642 | 0.286  (prev. version) |104.7|65|
| Qwen3_W4A16 <br> head:8bit<br>Spinquant<br>GPTQ_V2_alpha_0.25_mse_for_gptq_percdamp_0.12_normalize_H=true <br>Calib_samples: <br>vqav2=128 <br>wikitext2=128 <br>MMMU=64 |  9.86  | 0.702 | 0.886 | 0.319/0.079 | 0.617 | 0.234 (prev. version) | 102.32 | 62.6 |
|% degradation| 4.2% | 3.3% | 1% | 4.7%/0% | 2.5% | 5.2% (prev. version) | 2.3% | 2.4% |
| Qwen3_W4A16 <br> head:8bit<br>Spinquant<br>GPTQ_V2_alpha_0.25_mse_for_gptq_percdamp_0.1_normalize_H=true <br>Calib_samples: <br>vqav2=128 <br>wikitext2=128 <br>MMMU=96 <br> mlu=32 <br> videomme-text=32 | 9.88 |	0.71 | 0.884 | 0.283/0.076 | 0.616 | 0.23 | 103.55 | 64.5 |
|% degradation| 4.4% | 2.5%| 1.1%  | 15.5%/ 3.7% | 2.6% | 5.6% (prev. version) | 1.1% | 0.5% |
| Qwen3_W4A16 <br> head:8bit<br>Spinquant<br>GPTQ_V2_alpha_0.2_mse_for_gptq_percdamp_0.01_normalize_H=true <br>Calib_samples: <br>vqav2=128 split=testdev<br>wikitext2=128 split=train <br> mlu=40 split=dev <br>filtered textvqa=70 split=train| 9.93  |	 0.709 | 0.885 | 0.287/0.084 | 0.610  | 0.236 | 104.30 | 63.7 |
|% degradation| 4.9% | 2.6% | 1%  | 14%/0% | 3.2 | 5.6% (prev. version)  | 0.5% | 1.3%  |

Co-Authored-By: Cline

### DIFF
--- a/tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
+++ b/tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
@@ -268,16 +268,20 @@ class GPTQ:
     and then performs blockwise GPTQ quantization of the layer weights.
     """
 
-    def __init__(self, layer: nn.Module):
+    def __init__(self, layer: nn.Module, normalize_H: bool = True):
         """
         Initialize GPTQ state for a single layer.
 
         Args:
             layer: Quantizable layer. Supported types are Linear, Conv1d,
                 Conv2d, Conv3d, and ConvTranspose2d.
+            normalize_H: If True, use running average with sqrt(2/N)
+                normalization for Hessian accumulation.
+                If False, use simple summation.
         """
         self.layer = layer
         self.dev = self.layer.weight.device
+        self.normalize_H = normalize_H
 
         w = layer.weight.data.clone()
         if isinstance(layer, (nn.Conv1d, nn.Conv2d, nn.Conv3d)):
@@ -295,16 +299,34 @@ class GPTQ:
         self.nsamples = 0
         self.quantizer: Quantizer = Quantizer()
 
+        # GPTQv2: for tracking FP vs quantized input difference
+        self.dXXT: Optional[torch.Tensor] = None
+        self.native_inp: Optional[list] = None
+
     @torch.no_grad()
     def add_batch(self, inp: torch.Tensor, out: torch.Tensor) -> None:
         """
         Accumulate Hessian statistics from one calibration batch.
 
+        For GPTQv2, also processes native_inp (FP inputs) and computes dXXT.
+
         Args:
-            inp: Layer input tensor.
+            inp: Layer input tensor (from the model being quantized).
             out: Layer output tensor. Present for interface consistency.
         """
         del out
+
+        # Process native input for GPTQv2 (before reshaping inp)
+        native_inp_raw = None
+        if hasattr(self, "native_inp") and self.native_inp is not None:
+            if len(self.native_inp) == 0:
+                raise RuntimeError(
+                    "GPTQv2: native inputs exhausted. "
+                    "Number of add_batch calls exceeds collected FP inputs."
+                )
+            native_inp_raw = self.native_inp.pop(0)
+            if native_inp_raw is not None:
+                native_inp_raw = native_inp_raw.to(self.dev)
 
         if len(inp.shape) == 2:
             inp = inp.unsqueeze(0)
@@ -359,11 +381,92 @@ class GPTQ:
         elif isinstance(self.layer, nn.Conv3d):
             inp = _conv3d_input_to_unfolded(self.layer, inp)
 
-        self.H *= self.nsamples / (self.nsamples + batch_size)
+        # GPTQv2: Apply the same reshaping to native_inp
+        native_inp = None
+        if native_inp_raw is not None:
+            native_inp = native_inp_raw
+            if len(native_inp.shape) == 2:
+                native_inp = native_inp.unsqueeze(0)
+
+            if isinstance(self.layer, nn.Linear):
+                if len(native_inp.shape) > 2:
+                    native_inp = native_inp.reshape(-1, native_inp.shape[-1])
+                native_inp = native_inp.t()
+
+            elif isinstance(self.layer, nn.Conv2d):
+                padding = _normalize_2d_padding(self.layer.padding)
+                unfold = nn.Unfold(
+                    kernel_size=self.layer.kernel_size,
+                    dilation=self.layer.dilation,
+                    padding=padding,
+                    stride=self.layer.stride,
+                )
+
+                if self.layer.groups != 1:
+                    native_inp = native_inp.reshape(
+                        native_inp.size(0) * self.layer.groups,
+                        native_inp.size(1) // self.layer.groups,
+                        native_inp.shape[2],
+                        native_inp.shape[3],
+                    )
+
+                native_inp = unfold(native_inp).permute(1, 0, 2).flatten(1)
+
+            elif isinstance(self.layer, nn.Conv1d):
+                unfold = nn.Unfold(
+                    kernel_size=(1, self.layer.kernel_size[0]),
+                    dilation=(1, self.layer.dilation[0]),
+                    padding=(0, self.layer.padding[0]),
+                    stride=(1, self.layer.stride[0]),
+                )
+
+                if self.layer.groups != 1:
+                    native_inp = native_inp.reshape(
+                        native_inp.size(0) * self.layer.groups,
+                        native_inp.size(1) // self.layer.groups,
+                        native_inp.shape[2],
+                    )
+
+                native_inp = native_inp.unsqueeze(-2)
+                native_inp = unfold(native_inp).permute(1, 0, 2).flatten(1)
+
+            elif isinstance(self.layer, nn.ConvTranspose2d):
+                native_inp = get_matmul_input_for_convtranspose2d(
+                    self.layer, native_inp
+                )
+
+            elif isinstance(self.layer, nn.Conv3d):
+                native_inp = _conv3d_input_to_unfolded(self.layer, native_inp)
+
         self.nsamples += batch_size
-        inp = math.sqrt(2.0 / self.nsamples) * inp.float()
+
+        if self.normalize_H:
+            # Running average with sqrt(2/N) normalization
+            self.H *= (self.nsamples - batch_size) / self.nsamples
+            if self.dXXT is not None:
+                self.dXXT *= (self.nsamples - batch_size) / self.nsamples
+            inp = math.sqrt(2.0 / self.nsamples) * inp.double()
+        else:
+            inp = inp.double()
+
         assert self.H is not None
         self.H += inp.matmul(inp.t()).to(self.H.device)
+
+        # GPTQv2: Compute dXXT using native (FP) vs processed input difference
+        if native_inp is not None:
+            if self.dXXT is None:
+                self.dXXT = torch.zeros_like(self.H)
+
+            # Scale native_inp the same way as inp
+            if self.normalize_H:
+                native_inp = math.sqrt(2.0 / self.nsamples) * native_inp.double()
+            else:
+                native_inp = native_inp.double()
+
+            dX = native_inp.to(inp.device) - inp
+            self.dXXT += dX.matmul(inp.t()).to(dtype=self.H.dtype)
+            del native_inp_raw, native_inp
+            native_inp_raw = native_inp = None
 
     @torch.no_grad()
     def fasterquant(
@@ -374,6 +477,7 @@ class GPTQ:
         actorder: bool = False,
         static_groups: bool = False,
         verbose: bool = False,
+        alpha: float = 0.25,
     ) -> None:
         """
         Run blockwise GPTQ quantization on the layer weights.
@@ -419,6 +523,10 @@ class GPTQ:
         h[dead, dead] = 1
         w[:, dead] = 0
 
+        # GPTQv2: Zero out dead elements in dXXT
+        if self.dXXT is not None:
+            self.dXXT[:, dead] = 0
+
         if groupsize != -1 and self.quantizer.mse in {"mse_for_gptq", "smse_for_gptq"}:
             raise ValueError(
                 "GPTQ-adjusted MSE currently does not support groupsize != -1"
@@ -440,6 +548,8 @@ class GPTQ:
             w = w[:, perm]
             h = h[perm][:, perm]
             invperm = torch.argsort(perm)
+            if self.dXXT is not None:
+                self.dXXT = self.dXXT[perm][:, perm]
 
         losses = torch.zeros_like(w)
         q_all = torch.zeros_like(w)
@@ -452,6 +562,12 @@ class GPTQ:
         h = torch.linalg.cholesky(h, upper=True)
         hinv = h
 
+        # GPTQv2: Compute P correction matrix from dXXT
+        P = None
+        if self.dXXT is not None:
+            dXXT = self.dXXT.to(hinv.dtype)
+            P = alpha * ((dXXT @ hinv.T).triu(diagonal=1)) @ hinv
+
         self.quantizer.update(w, hinv, perm)
 
         for i1 in range(0, self.columns, blocksize):
@@ -463,6 +579,7 @@ class GPTQ:
             err1 = torch.zeros_like(w1)
             losses1 = torch.zeros_like(w1)
             hinv1 = hinv[i1:i2, i1:i2]
+            P1 = P[i1:i2, i1:i2] if P is not None else None
 
             for i in range(count):
                 w_col = w1[:, i]
@@ -493,11 +610,17 @@ class GPTQ:
 
                 cur_err = (w_col - q_col) / d
                 w1[:, i:] -= cur_err.unsqueeze(1).matmul(hinv1[i, i:].unsqueeze(0))
+                # GPTQv2: Apply P correction
+                if P1 is not None:
+                    w1[:, i:] += w_col.unsqueeze(1).matmul(P1[i, i:].unsqueeze(0))
                 err1[:, i] = cur_err
 
             q_all[:, i1:i2] = q1
             losses[:, i1:i2] = losses1 / 2
             w[:, i2:] -= err1.matmul(hinv[i1:i2, i2:])
+            # GPTQv2: Apply P correction to remaining weights
+            if P is not None:
+                w[:, i2:] += w1.matmul(P[i1:i2, i2:])
 
         if torch.cuda.is_available():
             torch.cuda.synchronize()
@@ -559,5 +682,7 @@ class GPTQ:
         self.H = None
         self.Losses = None
         self.Trace = None
+        self.dXXT = None
+        self.native_inp = None
         if torch.cuda.is_available():
             torch.cuda.empty_cache()
--- a/tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
+++ b/tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
@@ -268,16 +268,20 @@ class GPTQ:
     and then performs blockwise GPTQ quantization of the layer weights.
     """
 
-    def __init__(self, layer: nn.Module):
+    def __init__(self, layer: nn.Module, normalize_H: bool = True):
         """
         Initialize GPTQ state for a single layer.
 
         Args:
             layer: Quantizable layer. Supported types are Linear, Conv1d,
                 Conv2d, Conv3d, and ConvTranspose2d.
+            normalize_H: If True, use running average with sqrt(2/N)
+                normalization for Hessian accumulation.
+                If False, use simple summation.
         """
         self.layer = layer
         self.dev = self.layer.weight.device
+        self.normalize_H = normalize_H
 
         w = layer.weight.data.clone()
         if isinstance(layer, (nn.Conv1d, nn.Conv2d, nn.Conv3d)):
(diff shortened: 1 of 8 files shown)


Torrero:
**Results of applying GPTQv2 for Qwen3-vl with correct calibration datasets:**


| Config ID| PPL | mmlu | vqav2 | COCO (CIDEr/Bleu_4) | hellaswag | MMMU_Pro (vision) (under review)  | Llava-Bench Rel. | videomme (32 frames) |
|--|--|--|--|--|--|--|--|--|
| FP32                       | 9.46 | 0.735     | 0.895   | 0.335 / 0.079 | 0.642 | 0.286  (prev. version)/ 0.18/0.231* |104.7|65|
| Qwen3_W4A16 <br> head:8bit<br>Spinquant<br>GPTQ_V2_alpha_0.2_mse_for_gptq_percdamp_0.01 <br>Calib_samples: <br>vqav2=128 split=train<br>wikitext2=128 split=train <br> mlu=64 split=auxiliary_train <br>filtered textvqa=70 split=train | 9.77   | 0.708	 | 0.883  | 0.310/0.082 | 0.626  | 0.223  (prev. version)/0.144/ 0.202 | 102.91 | 63.7 |
|% degradation| 3.2%  | 2.7%  |  1.2%  |  7.4%/0%  | 1.6%  |  6.3%  (prev. version) /3.6%/2.9% | 1.8%  |  1.3%   |

* for MMMU_Pro (vision) evaluation used:  previous version / [mmmu prompt_mode=official_direct, max_new_tokens=50](https://github.com/Samsung/TICO/pull/929#issuecomment-5531617288)  / [mmmu prompt_mode=official_direct, max_new_tokens=256](https://github.com/Samsung/TICO/pull/929#issuecomment-5531617288) 

<details> 
<summary> Run command:  Qwen3_W4A16  head:8bit Spinquant GPTQ_V2_alpha_0.2_mse_for_gptq_percdamp_0.01</summary>


**Quantization:**

```
python -m tico.quantization.examples.quantize \
  --config tico/quantization/examples/configs/qwen3_vl_eval_suite.yaml \
  --set model.name_or_path=Qwen/Qwen3-VL-4B-Instruct \
  --set pipeline.0.enabled=true \
  --set pipeline.1.enabled=false \
  --set pipeline.2.mse=mse_for_gptq \
  --set pipeline.2.gptq_v2=true \
  --set pipeline.2.gptq_v2_alpha=0.2 \
  --set pipeline.2.percdamp=0.01 \
  --set pipeline.2.normalize_H=true \
  --set pipeline.2.c_correction_term=false \
  --set 'pipeline.2.weight_bits_overrides={"patch_embed.proj": 8}' \
  --set pipeline.2.move_cache_to_cpu=true \
  --set pipeline.2.sensitivity.mode=cache \
  --set pipeline.3.lm_head_weight=uint8 \
  --set pipeline.3.embedding_weight=uint8 \
  --set pipeline.3.verbose=true \
  --set pipeline.3.reuse_gptq_qparams=false \
  --set calibration.seq_len=2048 \
  --set calibration.datasets.0.n_samples=128 \
  --set calibration.datasets.0.split=train \
  --set calibration.datasets.1.n_samples=128 \
  --set calibration.datasets.1.split=train \
  --set calibration.datasets.2.dataset=mmlu \
  --set calibration.datasets.2.split=auxiliary_train \
  --set calibration.datasets.2.n_samples=64 \
  --set calibration.datasets.3.dataset=textvqa \
  --set calibration.datasets.3.split=train \
  --set 'calibration.datasets.3.filter={"field":"image_classes","n_per_class":2,"max_classes":50,"distinct_images":true}' \
  --set export.enabled=true \
  --set evaluation.enabled=false \
```

**Evaluation:**

```
python -m tico.quantization.examples.evaluate \
 --config tico/quantization/examples/configs/qwen3_vl_eval_suite.yaml \
 --checkpoint=quantized_model.pt \
  --set evaluation.enabled=true \
  --set evaluation.max_seq_len=2048 \
  --set evaluation.vlm_tasks=[vqav2] \
  --set evaluation.coco=true \
  --set evaluation.n_samples=1000 \
  --set evaluation.mmlu.enabled=true \
  --set evaluation.hellaswag.enabled=true \
  --set evaluation.hellaswag.n_samples=1000 \
  --set evaluation.mmmu.enabled=true \
  --set evaluation.mmmu.dataset=MMMU/MMMU_Pro \
  --set evaluation.mmmu.subjects=[vision] \
  --set evaluation.mmmu.n_samples=1000 \
  --set evaluation.mmmu.verbose=true \
  --set evaluation.ppl.enabled=true \
  --set evaluation.ppl.split=test \
  --set evaluation.ppl.stride=512 \
  --set evaluation.llava_bench.enabled=true \
  --set evaluation.llava_bench.judge.model_id=unsloth/Llama-3.2-3B-Instruct \
  --set evaluation.llava_bench.n_samples=50 \
  --set evaluation.llava_bench.candidate_label="Qwen3-VL-4B-Instruct" \
  --set evaluation.llava_bench.max_seq_len=2048 \
  --set evaluation.videomme.enabled=true \
  --set evaluation.videomme.max_new_tokens=30 \
  --set evaluation.videomme.n_samples=1000 \
  --set evaluation.videomme.max_num_frames=32 \
  --set evaluation.videomme.verbose=true \
```
</details>

tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
```diff
@@ -463,6 +579,7 @@ def fasterquant(
             err1 = torch.zeros_like(w1)
             losses1 = torch.zeros_like(w1)
             hinv1 = hinv[i1:i2, i1:i2]
+            P1 = P[i1:i2, i1:i2] if P is not None else None
 
             for i in range(count):
                 w_col = w1[:, i]
```
mhs4670go: w_col is a view into w1. The preceding in-place update mutates the current column before the P correction is evaluated, so the correction uses the updated/quantized value instead of the original weight. 

Please either clone w_col before the update or combine both terms into a single RHS as in the GPTAQ reference implementation. A final-weight reference test with alpha > 0 would catch this regression.

```python
w_col = w1[:, i].clone()

gptq_update = cur_err.unsqueeze(1).matmul(
    hinv1[i, i:].unsqueeze(0)
)

if p1 is not None:
    p_update = w_col.unsqueeze(1).matmul(
        p1[i, i:].unsqueeze(0)
    )
    w1[:, i:] -= gptq_update - p_update
else:
    w1[:, i:] -= gptq_update
```

tico/quantization/algorithm/qwen3_vl_gptq/quantizer.py
mhs4670go, line 1: The current implementation collects statistics for every quantizable module in the stage before quantizing any of them.

GPTAQ performs sequential group calibration (q/k/v -> o -> gate/up -> down) and re-forwards the partially quantized layer between groups. Without that schedule, downstream projections do not observe the asymmetric input error introduced by earlier projections in the same layer. 

Or, that implementation isintentionally skipped for a future extention?

tico/quantization/algorithm/qwen3_vl_gptq/quantizer.py
```diff
@@ -443,6 +731,120 @@ def _apply_text_post_layer_processing(
...
+
+        # Save to in-memory disk cache for later persistence
+        if self.config.fp_inputs_cache_path is not None:
+            self._fp_inputs_disk_cache[stage_desc] = {
+                k: list(v) for k, v in result.items()
+            }
+        return result
+
+    @torch.no_grad()
+    def _collect_native_inputs_from_stage_cache(
```
mhs4670go, line 804: _collect_native_inputs_from_stage_cache() never stores the collected result in _fp_inputs_disk_cache, so vision/text block entries are absent from the persisted cache.

On a warm run, a missing entry is silently recomputed from the current model and then treated as the native input, which can make dXXT approximately zero. 

How about persisting every supported stage and make warm-cache lookup fail closed on any missing or incompatible entry. The test should exercise an actual save/load round trip without manually inserting the result into the cache dictionary.

tico/quantization/algorithm/qwen3_vl_gptq/quantizer.py
mhs4670go, line 238: The cache is persisted from finally, so a failed conversion can publish a partial cache as if it were complete. Please save only after successful completion, preferably through a temporary file plus atomic rename, and include a completeness/schema manifest that is validated on load.

tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
```diff
@@ -359,12 +381,93 @@ def add_batch(self, inp: torch.Tensor, out: torch.Tensor) -> None:
...
+
         self.nsamples += batch_size
-        inp = math.sqrt(2.0 / self.nsamples) * inp.float()
+
+        if self.normalize_H:
+            # Running average with sqrt(2/N) normalization
+            self.H *= (self.nsamples - batch_size) / self.nsamples
+            if self.dXXT is not None:
+                self.dXXT *= (self.nsamples - batch_size) / self.nsamples
+            inp = math.sqrt(2.0 / self.nsamples) * inp.double()
```
mhs4670go, line 448: This changes the accumulation path to FP64 even when gptq_v2=False, while H and dXXT are still allocated as FP32. As a result, we pay for an FP64 matrix multiplication and then cast the result back into an FP32 accumulator. This can be a significant time and temporary-memory regression for large Hessians, especially in projections with a large input dimension.

Could we instead cast to the accumulator dtype, for example inp.to(self.H.dtype), and do the same for native_inp? If FP64 accumulation is intentionally required, it would be better exposed as a separate explicit option, with H/dXXT allocated consistently and a performance justification.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -1708,3 +1847,161 @@ def get_mixed_calib_inputs(
...
+    class_freq: Dict[str, int] = {}
+    for ex in examples:
+        filter_classes = ex.get(filter_config.filter_field, [])
+        if not filter_classes:
+            continue
+        for cls in filter_classes:
+            cls_str = str(cls)
+            class_freq[cls_str] = class_freq.get(cls_str, 0) + 1
+
+    if not class_freq:
```
mhs4670go, line 1925: An active filter silently returns the entire dataset when the configured field is absent. Since the filtered path loads with n=-1 and ignores n_samples, a typo such as image_class instead of image_classes can unexpectedly turn a small calibration run into full-dataset preprocessing. Could this raise a ValueError containing the dataset name and configured field? A silent fallback makes configuration errors hard to detect and may consume a large amount of time and memory.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -1708,3 +1847,161 @@ def get_mixed_calib_inputs(
...
+    seen_image_ids: set = set()
+    skipped_dup_images = 0
+
+    for ex in examples:
+        filter_classes = ex.get(filter_config.filter_field, [])
+        if not filter_classes:
+            continue
+
+        # Check if any of this sample's classes is still under quota
+        under_quota = [
```
mhs4670go, line 1974: This does not guarantee the documented “at most n_per_class samples per class” behavior for multi-label examples.

For example, with n_per_class=1:

```
sample 1: [cat]
sample 2: [cat, dog]
```

both samples are selected because dog is still under quota, but cat appears in two selected samples. The internal class_counts["cat"] remains 1, so the reported count also differs from the actual selected-set distribution.

Please either:

1. enforce a strict cap by selecting a sample only when all of its target classes are below quota and incrementing every selected class, or
2. rename and document this as a coverage/credit heuristic rather than a per-class maximum.

The test should verify class occurrences in the selected examples, not only len(result).

tico/quantization/algorithm/qwen3_vl_gptq/quantizer.py
```diff
@@ -41,6 +44,46 @@
 from tico.quantization.quantizer_registry import register_quantizer
 
 
+class FPInputsCache:
```
mhs4670go, line 47: The current cache stores the full input tensor for every quantizable submodule, batch, and layer. For long-sequence calibration this can become extremely large, and several projections share exactly the same input—for example q/k/v and gate/up.

Before making this a persistent cache format, we could avoid duplicate tensors and store only one entry per shared-input group. It would also be safer to shard the cache by stage/layer and load one shard at a time rather than keeping one monolithic dictionary in memory.

tico/quantization/config/gptq.py
```diff
@@ -69,6 +69,26 @@ class GPTQConfig(BaseConfig):
     # use this option to stabilize GPTQ for deep models
     use_orig_model_inference: bool = False
 
+    # GPTQv2: uses FP inference for collecting inputs during quantization
```
mhs4670go, line 72: These fields are added to the generic GPTQConfig, but this PR implements their behavior only in Qwen3VLGPTQQuantizer. Other configs inheriting from GPTQConfig, such as Gemma4 GPTQ, can therefore accept and display gptq_v2=true without actually applying GPTQv2.

Could we either:

* move the GPTQv2-specific options to Qwen3VLGPTQConfig for now, or
* make every GPTQ quantizer explicitly support them or reject them during validation?

Silently accepting an unsupported algorithm flag would be misleading.

tico/quantization/config/gptq.py
```diff
@@ -69,6 +69,26 @@ class GPTQConfig(BaseConfig):
...
+
+    # GPTQv2: scaling factor for the asymmetric correction (P matrix)
+    # `alpha` is the correction strength for GPTQv2's input-error compensation.
+    # It scales the `P` matrix that adjusts weight updates to account for upstream quantization error in the activations.
+    # A value of `0` disables the correction (standard GPTQ), while values around `0.25` provide the best empirical results.
+    gptq_v2_alpha: float = 0.25
+
+    # Use running average for Hessian accumulation.
+    # When False, uses summation.
+    normalize_H: bool = False
```
mhs4670go, line 90: The previous Qwen3-VL GPTQ implementation always applied the running normalization, but the new generic default is False. This changes the default behavior even when GPTQv2 is disabled, while the reported benchmark explicitly uses normalize_H=true.

Is it intended?

tico/quantization/recipes/data/vlm.py
```diff
@@ -76,6 +76,52 @@ def _parse_dataset_spec(
     return dataset, config
 
 
+def _normalize_filter_config(filter_cfg: Any, *, context: str) -> dict[str, Any]:
```
mhs4670go, line 79: The docstring lists verbose as a supported filter option, and get_mixed_calib_inputs() reads it, but _normalize_filter_config() does not copy it into the normalized dictionary. As a result, a YAML value such as verbose: false is discarded and later falls back to True.

Please preserve the field here and add a normalization test for both True and False.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -32,6 +33,44 @@
...
+                    classes, selected samples, and per-class counts.  Defaults to
+                    ``False`` (silent).
+    """
+
+    n_per_class: int = 0
+    classes: Optional[List[str]] = None
+    max_classes: Optional[int] = None
+    distinct_images: bool = True
+    filter_field: str = "image_classes"
+    verbose: bool = True
```
mhs4670go, line 66: The attribute documentation says the default is False, but the dataclass value is True. 

Since verbose mode prints every selected question, False seems like the safer library default, with the final summary optionally controlled separately.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -1628,6 +1740,33 @@ def get_mixed_calib_inputs(
...
+                distinct_images=filter_dict.get("distinct_images", True),
+                filter_field=filter_dict.get("field", "image_classes"),
+                verbose=filter_dict.get("verbose", True),
+            )
+
+            print(
+                f"[info] Filtering '{dataset}' by {fc.filter_field} "
+                f"(n_per_class={fc.n_per_class}) in mixed mode"
+            )
+            class_inputs = get_calib_inputs(
```
mhs4670go, line 1760: This filtered branch does not forward allow_benchmark_overlap or allow_unregistered_dataset, while the normal dataset-loading branch does. This makes dataset policy behavior depend on whether a filter block is present.

Please forward both flags so that filtering changes only sample selection, not dataset-usage validation.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -1628,6 +1740,33 @@ def get_mixed_calib_inputs(
 
         is_text_only = DATASETS[dataset].get("is_text_only", False)
 
+        # --- Per-dataset filter block (e.g. TextVQA class filtering) ---
+        filter_dict: Optional[Dict[str, Any]] = config.get("filter")
+        if filter_dict and filter_dict.get("n_per_class", 0):
```
mhs4670go, line 1746: This condition uses truthiness, so a negative n_per_class enters the filtered branch even though CalibFilterConfig.is_active considers it disabled. The subsequent call also omits the configured n_samples, so the behavior can fall back to get_calib_inputs()’s default sample count rather than the requested value.

Please validate n_per_class >= 0 during normalization and use an explicit > 0 condition consistently.

Torrero:
**Fixed and re-evaluated GPTQv2 for Qwen3-vl with correct calibration datasets:**


| Config ID| PPL | mmlu | vqav2 | COCO (CIDEr/Bleu_4) | hellaswag | MMMU_Pro (vision) (under review)  | Llava-Bench Rel. | videomme (32 frames) |
|--|--|--|--|--|--|--|--|--|
| FP32                       | 9.46 | 0.735     | 0.895   | 0.335 / 0.079 | 0.642 | 0.305/ 0.18/0.231* |104.7|65|
| Qwen3_W4A16 <br> head:8bit<br>Spinquant<br>GPTQ_V2_alpha_0.2_mse_for_gptq_percdamp_0.01 <br>Calib_samples: <br>vqav2=128 split=train<br>wikitext2=128 split=train <br> mlu=64 split=auxiliary_train <br>filtered textvqa=70 split=train | 9.88   | 	0.705 | 0.882   | 0.323/0.083 | 0.621  |     0.215/0.184/0.195  | 104.15 |   65  |
|% degradation| 4.4%  | 3%  |  1.3%  |  3.5%/0%  | 2.1%  |  9%/ 0%/ 3.6%  | 0.5%  |  0%   |

* For MMMU_Pro (vision) evaluation used: [single-letter mode prompt](https://github.com/Samsung/TICO/pull/927) / https://github.com/Samsung/TICO/pull/929#issuecomment-5531617288 / https://github.com/Samsung/TICO/pull/929#issuecomment-5531617288

tico/quantization/algorithm/qwen3_vl_gptq/gptq.py
Torrero: Thank you for pointing out this Inconsistency. I fixed it and re-evaluated the model [comment](https://github.com/Samsung/TICO/pull/915#issuecomment-5562975141)

PR #916, opened 2026-08-18
[quantization] Support MMMU_Pro dataset for Qwen3 calibration

Torrero:
This commit supports MMMU_Pro dataset for Qwen3 calibration

Co-Authored-By: Cline

### DIFF
--- a/tico/quantization/evaluation/vlm_eval_utils.py
+++ b/tico/quantization/evaluation/vlm_eval_utils.py
@@ -242,6 +242,40 @@ def get_item_alpaca(ex: dict[str, Any]) -> dict[str, Any]:
     return {"text": text}
 
 
+def get_item_mmmu_calib(ex: dict[str, Any]) -> dict[str, Any]:
+    """
+    Adapt an MMMU/MMMU_Pro sample for VLM calibration.
+
+    MMMU samples may contain multiple images (image_1, image_2, ...).
+    Only single-image samples are used; multi-image samples are skipped
+    by returning ``None`` for the image field, which the caller filters out.
+
+    The returned schema is:
+
+    `{"image": image, "question": question, "golds": gold_answers}`
+
+    Args:
+        ex: Raw dataset example.
+
+    Returns:
+        A normalized calibration item.
+    """
+    # Skip multi-image samples
+    if ex.get("image_2") is not None:
+        return {"image": None, "question": "", "golds": []}
+
+    image = ex.get("image_1") or ex.get("image")
+    question = ex.get("question", "")
+    answer = ex.get("answer", "")
+    golds = [str(answer)] if answer else []
+
+    return {
+        "image": image,
+        "question": question,
+        "golds": golds,
+    }
+
+
 DATASETS: dict[str, dict[str, Any]] = {
     "vqav2": {
         "default_split": "validation",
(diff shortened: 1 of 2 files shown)


tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -283,6 +317,13 @@ def get_item_alpaca(ex: dict[str, Any]) -> dict[str, Any]:
         "candidates": ["tatsu-lab/alpaca"],
         "is_text_only": True,
     },
+    "mmmu_pro_vision": {
```
mhs4670go, line 320: The default split is set correctly. But, [the qwen adapter](https://github.com/Samsung/TICO/blob/0ef6d97b3aefd4805e083ddd31957cacb243c802/tico/quantization/recipes/adapters/qwen3_vl.py#L150) always assumes the default split is `testdev`. So, when the yaml file has contents like below, it would cause an error  because MMMU-pro dataset only has `test` split.

```yaml
calibratio:
  dataset: mmmu_pro_vision
  n_samples: 128
  seq_len: 2048
```

How about below changes? This would make the dataset registry decide the decision of the default split by itself.

```python
# qwen3_vl.py / gemma4.py
split=calib.get("split")

# recipes/data/vlm.py
def build_vlm_calibration_inputs(
    ...,
    split: str | None = None,
):
    ...

# vlm_eval_utils.py
def get_calib_inputs(
    ...,
    split: str | None = None,
):
    ...
```

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -364,9 +479,24 @@ def build_vlm_inputs(
         "images": image,
         "return_tensors": return_tensors,
     }
-    if max_seq_len is not None and max_seq_len > 0:
-        processor_kwargs["truncation"] = True
-        processor_kwargs["max_length"] = max_seq_len
+
+    if image is not None:
```
mhs4670go, line 483: This is right as you only focus on Qwen3 model. But, the proceesor of Gemma4 should be considered as well. It doesn't have some kwargs like `min_pixels` or `max_pixels`.

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -337,6 +378,79 @@ def build_prompt(processor, question: str) -> str:
...
+
+    # Convert tokens to pixels
+    # tokens = (H / vision_factor) × (W / vision_factor)
+    # pixels = H × W = tokens × vision_factor²
+    max_pixels = visual_budget * vision_factor * vision_factor
+
+    # Ensure min_pixels is at most max_pixels (default min_pixels=200,704
+    # can exceed the computed max_pixels for tight budgets).
+    # (comes from processor.image_processor)
+    min_pixels = min(256 * 28 * 28, max_pixels)
```
mhs4670go, line 446: How about getting the value from the processor?

tico/quantization/evaluation/vlm_eval_utils.py
```diff
@@ -337,6 +378,79 @@ def build_prompt(processor, question: str) -> str:
...
+        text_token_margin: Extra margin for text tokens (system prompt,
+            special tokens, etc.).
+
+    Returns:
+        A ``(max_pixels, min_pixels)`` tuple to pass to the processor.
+    """
+    vision_factor = _processor_vision_factor(processor)
+
+    # visual_budget = tokens available for visual tokens (excluding text margin)
+    visual_budget = max_seq_len - text_token_margin
```
mhs4670go, line 429: This pixel-budget path is Qwen-specific, but `build_vlm_inputs()` is also shared by the Gemma4 calibration and evaluation paths. Gemma4 uses `max_soft_tokens` and `pooling_kernel_size` rather than `max_pixels` and `min_pixels`, so the newly added kwargs are ignored while the previous truncation/max_length behavior is removed.

mhs4670go:
I'll merge this PR first and then, I'll post a PR for the suggestion fixes.

mhs4670go (approved):
LGTM

Issue #951, opened 2026-08-25
[quantization] Prevent benchmark data leakage between calibration and evaluation


mhs4670go:
## Background

This issue was identified while reviewing #950, but the root cause already exists on `main` and should be fixed independently of that PR.

Quantization calibration is data-dependent. PTQ observers and clipping ranges, SmoothQuant activation statistics, and GPTQ input/Hessian statistics are all fitted from the calibration samples. Therefore, using benchmark evaluation samples for calibration makes the resulting evaluation transductive rather than strictly held out, even when no gradients are computed. Including the gold answer in calibration text is a stronger form of target leakage.

## Problem on [main](https://github.com/Samsung/TICO/commit/c9011e6e7d228477dc02614e7cda80a0449ce80a)

The current recipe infrastructure does not distinguish between a dataset being used for calibration and the same dataset being used for evaluation.

- `tico/quantization/evaluation/vlm_eval_utils.py` has one shared `DATASETS` registry.
- Each registry entry has one `default_split`, regardless of whether the caller is calibrating or evaluating.
- `get_dataset()` has no `purpose` or `role` argument.
- `get_calib_inputs()` and `get_mixed_calib_inputs()` use this shared loader for calibration.
- VQA, COCO-style, and PPL evaluation paths also use the same registry and loader.
- MMLU, Video-MME, and MMMU evaluation are routed through separate helpers or external harnesses, so their resolved dataset/config/split is not visible to one common validator.
- `QuantizationRunner.run()` validates supported evaluation targets before loading the model, but it does not validate calibration dataset provenance or calibration/evaluation overlap.

This creates several unsafe configurations that are currently accepted.

| Source | Current behavior | Risk |
| --- | --- | --- |
| VQAv2 | The shared registry defaults to `validation`; the default VLM calibration dataset is also `vqav2`, and VQAv2 evaluation uses the same registry default. | Calibration and evaluation can use the exact same split. |
| VQAv2 `testdev` | Current Qwen3-VL and Gemma4 quantization examples explicitly use `testdev` for calibration. | The checkpoint is adapted to a benchmark evaluation split and may later be reported on that split. |
| TextVQA | The shared default is `validation` for both calibration and evaluation. | Exact split overlap is possible. |
| COCO Captioning | The shared default is `val` for both calibration and evaluation. | Exact split overlap is possible. |
[...]

PR #953, opened 2026-08-26
[quantization] Prevent benchmark data

mhs4670go:
This commit prevents benchmark data leakage between calibration and evaluation.

Related: #951

### DIFF
--- a/tico/quantization/recipes/data/dataset_usage.py
+++ b/tico/quantization/recipes/data/dataset_usage.py
@@ -1,0 +1,1087 @@
+# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Resolve semantic dataset roles and prevent benchmark leakage."""
+
+from __future__ import annotations
+
+import warnings
+
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass
+from typing import Any
+
+
+CALIBRATION_ROLE = "calibration"
+EVALUATION_ROLE = "evaluation"
+FEW_SHOT_ROLE = "few_shot"
+EVALUATION_CONTEXT_ROLE = "evaluation_context"
+
+_DATA_USE_ROLES = frozenset(
+    {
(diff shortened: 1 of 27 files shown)


tico/quantization/recipes/data/dataset_usage.py
```diff
@@ -0,0 +1,1087 @@
...
+        raise ValueError(
+            f"Unsupported dataset role {role!r}. Supported roles: "
+            f"{sorted(_DATA_USE_ROLES)}"
+        )
+    if not str(dataset).strip():
+        raise ValueError("Dataset identifier must not be empty.")
+    if not str(consumer).strip():
+        raise ValueError("Dataset consumer must not be empty.")
+
+    policy = get_dataset_policy(dataset)
```
Torrero, line 322: If I understand correctly, the current implementation allows any unregistered dataset with split to be used for calibration without any errors or warnings.
Maybe, when `policy is None` and the role is `calibration`, either raise an error requiring explicit registration, or emit a warning that safety checks are skipped for unregistered datasets.

tico/quantization/recipes/data/dataset_usage.py
```diff
@@ -0,0 +1,1087 @@
...
+            raise ValueError(
+                f"Dataset {dataset!r} has no registered default split for role "
+                f"{normalized_role!r}; specify the split explicitly."
+            )
+        resolved_split = policy.default_splits.get(normalized_role)
+        if resolved_split is None:
+            # Resolve the dataset's benchmark split so the central validator can
+            # report the semantic role violation and honor an explicit
+            # transductive override.
+            resolved_split = next(iter(policy.default_splits.values()), None)
```
Torrero: When policy has no default split (`default_splits={}`) the resolved split will be first arbitrary split without any warning.

tico/quantization/recipes/data/dataset_usage.py
```diff
@@ -0,0 +1,1087 @@
...
+    )
+
+
+DATASET_POLICIES: dict[str, DatasetPolicy] = {
+    "vqav2": _policy(
+        key="vqav2",
+        canonical_id="HuggingFaceM4/VQAv2",
+        aliases=("vqav2", "HuggingFaceM4/VQAv2", "lmms-lab/VQAv2"),
+        allowed_roles=(CALIBRATION_ROLE, EVALUATION_ROLE),
+        default_splits={CALIBRATION_ROLE: "train", EVALUATION_ROLE: "validation"},
```
Torrero, line 146: `lmms-lab/VQAv2` has no `train` split, only `test`, `testdev`, `validation`. It raises an error.

mhs4670go:
@Torrero Thank you for the detailed reviews! I've addressed below things. PTAL:)

- Reject unregistered calibration datasets by default and add an explicit opt-in for custom datasets.
- Remove fallback to another role’s default split; require an explicit split when the requested role has no default.
- Switch VQAv2 to lmms-lab/VQAv2-FewShot with the full config so train calibration works correctly.

Torrero (approved):
LGTM, thank you!

Issue #971, opened 2026-09-03
[quantization] Model-Agnostic GPTQ Quantizer


dvsav:
### Background

**GPTQ** is a weight-only quantization algorithm that processes weight matrices, using the Hessian of the input activations to minimize the output error introduced by quantization. GPTQ quantizes a model layer-by-layer in the order the layers (submodules) are called during inference run. A critical property of GPTQ is that quantizing a layer `i` introduces an error that propagates to all subsequent layers `i+1, i+2, ...`. This means the quantization of each layer depends on the quantization error accumulated in all upstream layers.

To capture this error propagation correctly, TICO's GPTQ quantizers use **progressive unrolling**: the model is executed layer-by-layer, and after quantizing each layer's weights, the model is re-run so that downstream layers receive activations that reflect the upstream quantization error. This ensures that each layer's Hessian is computed on realistic (post-quantization) activations, not on the original unquantized activations.

### The Problems

**1. Progressive unrolling requires model-specific knowledge**: the quantizer must know the model's architecture to determine which submodules to process, in what order, and how to feed the output of a quantized submodule forward as input for the next submodule's Hessian collection. This is currently implemented via hand-crafted quantizer classes (e.g., `Gemma4GPTQQuantizer`, `Qwen3VLGPTQQuantizer`) that:

- Know that they must quantize vision encoder submodules in a certain order and then quantize the submodules of language model.
- Know the exact names of blocks eligible to GPTQ quantization.
- Partially replicate the computation flow of specific models being quantized.

As a result, **every new model architecture requires a new quantizer/config class** — typically a thousand of lines of model-specific code. This is tedious, error-prone, and creates a significant maintenance burden. Supporting a new model family (e.g., Gemma, Mistral, Phi) requires understanding the model's architecture in detail and writing correct quantizer code, even though the GPTQ algorithm itself is model-agnostic.

**Consequences:**
- **High barrier to entry:** Adding GPTQ support for a new model requires deep knowledge of model architecture and GPTQ framework.
- **Code duplication:** Each model family re-implements the same progressive unrolling logic with minor architectural variations.
- **Fragility:** Model architecture changes (e.g., a new attention variant) may break the corresponding quantizer.

**2. Entire model or its large blocks are needlessly rerun during quantization:** when we need to collect the inputs for downstream submodules we run the entire model and hence already quantized upstream submodules run multiple times and compute exactly same outputs.

[...]

PR #972, opened 2026-09-10
[quantization] Implement Model-Agnostic GPTQ Quantizer

dvsav:
Issue: [\[quantization\] Model-Agnostic GPTQ Quantizer](https://github.com/Samsung/TICO/issues/971)

## What

This PR introduces a **universal (model-agnostic) GPTQ quantizer** that can quantize any PyTorch model without requiring model-specific wrapper classes. The implementation uses a novel **forward method replacement** strategy combined with a **state machine** approach to perform layer-by-layer GPTQ quantization.

Key features:
- Works with any PyTorch model containing `nn.Linear`, `nn.Conv1d/2d/3d`, or `nn.ConvTranspose2d` layers
- Automatically handles residual/skip connections
- Memory-efficient caching with automatic release of nested submodule outputs
- No model-specific code required - truly plug-and-play

## Why

Currently, TICO requires **model-specific GPTQ quantizers** (e.g., `Gemma4GPTQQuantizer`, `Qwen3VLGPTQQuantizer`) that must be hand-crafted for each model architecture. This approach has several limitations:

1. **High development cost**: Each new model family requires writing custom wrapper classes for every submodule type
2. **Maintenance burden**: Model architecture changes require updating corresponding quantizers
3. **Limited coverage**: Many models remain unsupported due to the effort required

The universal GPTQ quantizer addresses these issues by:
- Eliminating the need for model-specific wrapper classes
- Automatically discovering and quantizing all GPTQ-applicable layers
- Handling arbitrary model structures including residual connections and nested modules

### 1. Forward Method Replacement vs. Wrapper Classes
**Decision**: Replace `forward` methods instead of wrapping modules in `CacheWrapper` classes.

**Rationale**: 
- Cleaner implementation - no module hierarchy changes
- Easier to track state per-module via `gptq_data` attribute
- More flexible state management (COLLECT → CACHE → RETURN_CACHED → COMPUTE)

### 2. State Machine Approach
**Decision**: Use explicit 4-state machine (COLLECT, CACHE, RETURN_CACHED, COMPUTE)

**Rationale**:
- Clear separation of concerns at each stage
- Enables memory optimization (free children's cache when parent transitions)
- Prevents incorrect behavior (module cannot return cached output before quantization completes)

### 3. Per-Module Invocation Tracking
**Decision**: Track `invocation_idx` and `num_invocations` per module instead of global batch counter.

**Rationale**:
- Correctly handles modules called multiple times per forward pass
- Supports models where different modules have different call frequencies
- Enables proper cache indexing for nested submodules

### 4. Memory-Efficient Cache Release
**Decision**: Free children's cached outputs when parent finishes caching.

**Rationale**:
- When a parent block (e.g., `ResidualBlock`) transitions to RETURN_CACHED, its children are never called
- Children's cached outputs become dead weight
- Reduces memory from O(all layers) to O(one block depth) at any time

### 5. CPU Caching
**Decision**: Cache outputs on CPU (`move_to_cpu`) instead of GPU.

**Rationale**:
- GPU memory is precious for large models
- Calibration batches are processed sequentially - CPU↔GPU transfer overhead is acceptable
- Prevents OOM during caching phase

### New Files

| File | Description |
|------|-------------|
| `tico/quantization/algorithm/gptq/universal_gptq_quantizer.py` | Core implementation with `GPTQQuantizer` class, state machine, and caching logic |
| `test/quantization/algorithm/test_universal_gptq.py` | Comprehensive smoke tests with multiple model architectures |

### Key Components Added

**`universal_gptq_quantizer.py`:**
- `GPTQQuantizer` - Main quantizer class implementing `prepare()`/`convert()` interface
- `GPTQ_STATE` - 4-state enum (COLLECT, CACHE, RETURN_CACHED, COMPUTE)
- `GPTQ_Data` - Dataclass holding per-module quantization state
- `wrap_model()` / `unwrap_model()` - Forward method replacement utilities
- `gptq_quantize()` - Main quantization loop with frontier-based execution
- `finish_collection()` - Configures quantizer and runs `fasterquant()`
- `finish_caching()` - Transitions to RETURN_CACHED and frees children's cache
- Memory optimization: `gptq_data.cached_output.clear()` for children when parent finishes caching

**`test_universal_gptq.py`:**
- `FiveLinearModel` - Simple 5-layer sequential model
- `ResidualBlock` - Block with skip connection (Linear → ReLU → Linear + identity)
- `ConvBlock` - Convolutional block with skip connection (Conv2d → BN → ReLU → Conv2d → BN + identity)
- `MultiBlockModel` - Complex hybrid model with residual + conv blocks
- Test cases: basic quantization, weight update, output consistency, nested cache release

### Test Coverage

| Test | Model | What It Tests |
|------|-------|---------------|
| `test_universal_gptq_basic` | `FiveLinearModel` | Basic workflow, correct quantizer count, proper layer paths |
| `test_universal_gptq_weight_update` | `FiveLinearModel` | Weights are actually modified by GPTQ |
| `test_universal_gptq_output_consistency` | `FiveLinearModel` | Output fidelity (PEIR < 10% for 8-bit) |
| `test_universal_gptq_with_residual_blocks` | `MultiBlockModel` | Residual connections, nested submodules, mixed layer types |
| `test_universal_gptq_nested_submodule_cache_release` | `MultiBlockModel` (simplified) | Memory optimization, cache integrity (no NaN/Inf) |

### Running Tests

```bash
python -m pytest test/quantization/algorithm/test_universal_gptq.py -v

============================================================= test session starts =============================================================
platform linux -- Python 3.10.12, pytest-8.4.0, pluggy-1.6.0 -- /home/d.savchenkov/myenv/bin/python3
cachedir: .pytest_cache
configfile: pyproject.toml
plugins: anyio-4.12.0, mock-3.15.1, xdist-3.7.0, cov-6.2.1
collected 5 items                                                                                                                             

test/quantization/algorithm/test_universal_gptq.py::TestUniversalGPTQ::test_universal_gptq_basic PASSED                                 [ 20%]
test/quantization/algorithm/test_universal_gptq.py::TestUniversalGPTQ::test_universal_gptq_nested_submodule_cache_release PASSED        [ 40%]
test/quantization/algorithm/test_universal_gptq.py::TestUniversalGPTQ::test_universal_gptq_output_consistency PASSED                    [ 60%]
test/quantization/algorithm/test_universal_gptq.py::TestUniversalGPTQ::test_universal_gptq_weight_update PASSED                         [ 80%]
test/quantization/algorithm/test_universal_gptq.py::TestUniversalGPTQ::test_universal_gptq_with_residual_blocks PASSED                  [100%]

======================================================== 5 passed, 1 warning in 2.39s =========================================================
```

Tests verify:
- ✅ Correct number of quantizers created
- ✅ All expected layer paths are quantized
- ✅ Weights are updated (not unchanged)
- ✅ Output shapes match original model
- ✅ Relative error within acceptable bounds (< 15% for complex models)
- ✅ No NaN/Inf from broken caching
- ✅ Non-zero outputs (model actually computes)

## Future Work

- Testing and evaluation with supported LLMs/VLMs (Llama, Qwen, Gemma)


dvsav:
I've made a side-by-side comparison (see `tico/quantization/examples/compare_side_by_side.py`) of the outputs of all layers of TinyLlama for 3 cases:
- "Default" GPTQ quantizer (`tico.quantization.algorithm.gptq.quantizer.GPTQQuantizer`)
- "Model-agnostic" GPTQ quantizer (`tico.quantization.algorithm.universal_gptq.quantizer.UniversalGPTQQuantizer`)
- No GPTQ quantization (PTQ only)

The comparison shows the per-layer PEIR of quantized vs FP model (across the entire calibration dataset):
- [llama_default_gptq.log](https://github.com/user-attachments/files/32292188/llama_default_gptq.log)
- [llama_model_agnostic_gptq.log](https://github.com/user-attachments/files/32292189/llama_model_agnostic_gptq.log)
- [llama_no_gptq.log](https://github.com/user-attachments/files/32292190/llama_no_gptq.log)

The PEIR numbers are similar for the "Default" and "Model-agnostic" quantizers and both are generally better than PTQ only. For example:
```
=== DEFAULT ===
model.layers.2.self_attn.q_proj: mean|diff|=0.11720913, max|diff|=3.06908131, peir=15.206653%
model.layers.2.self_attn.k_proj: mean|diff|=0.48057956, max|diff|=4.59450340, peir=26.309054%
model.layers.2.self_attn.v_proj: mean|diff|=0.07255870, max|diff|=2.11191130, peir=22.437813%
model.layers.2.self_attn.o_proj: mean|diff|=0.01849241, max|diff|=0.39020842, peir=24.282093%

=== MODEL AGNOSTIC ===
model.layers.2.self_attn.q_proj: mean|diff|=0.11891412, max|diff|=3.13113022, peir=15.550750%
model.layers.2.self_attn.k_proj: mean|diff|=0.48048456, max|diff|=4.45088196, peir=26.501972%
model.layers.2.self_attn.v_proj: mean|diff|=0.07330883, max|diff|=2.04044032, peir=22.792178%
model.layers.2.self_attn.o_proj: mean|diff|=0.01886921, max|diff|=0.42189223, peir=24.417110%

=== PTQ ONLY ===
model.layers.2.self_attn.q_proj: mean|diff|=0.13874104, max|diff|=3.78267956, peir=16.396651%
model.layers.2.self_attn.k_proj: mean|diff|=0.47856715, max|diff|=4.50577974, peir=27.336685%
model.layers.2.self_attn.v_proj: mean|diff|=0.08777730, max|diff|=1.87823772, peir=24.508446%
model.layers.2.self_attn.o_proj: mean|diff|=0.02217917, max|diff|=0.41766059, peir=23.661203%
```

tico/quantization/algorithm/universal_gptq/quantizer.py
```diff
@@ -0,0 +1,814 @@
...
+    if suffix_matches:
+        return suffix_matches[-1]
+
+    return gptq_config.weight_bits
+
+
+def get_sensitivity(
+    sensitivity: dict[str, torch.Tensor] | None,
+    full_module_name: str,
+) -> torch.Tensor | None:
```
mhs4670go, line 650: The implementation might be missing.

dvsav: 👍  Thanks for catching that! Fixed.

tico/quantization/config/gptq.py
```diff
@@ -89,3 +89,63 @@ def validate(self) -> None:
             raise ValueError(f"groupsize must be -1 or positive. got {self.groupsize}")
         if not (0.0 < self.percdamp <= 1.0):
             raise ValueError(f"percdamp must be in (0, 1]. got {self.percdamp}")
+
+
+@dataclass
+class UniversalGPTQConfig(GPTQConfig):
```
mhs4670go, line 95: It seems that quantize_lm_head=False is currently ignored by the universal quantizer.

UniversalGPTQConfig inherits this option from GPTQConfig, but wrap_model() selects quantization targets based only on the module type. Therefore, an nn.Linear used as lm_head is still quantized even with the default quantize_lm_head=False.

This is especially important for models with tied embeddings, where lm_head.weight may share the same parameter with the input embedding.

Could we make the target-selection logic respect this option? It may also be worth explicitly detecting tied/shared target parameters, since modifying one target can otherwise modify another part of the model implicitly.

tico/quantization/algorithm/universal_gptq/quantizer.py
```diff
@@ -0,0 +1,822 @@
...
+    Returns:
+        The device where the model currently resides.
+    """
+    try:
+        return next(model.parameters()).device
+    except StopIteration:
+        return None
+
+
+def infer_object_device(obj: Any) -> torch.device | None:
```
mhs4670go, line 99: (optional) I think infer_object_device() is too strict for nested outputs containing non-tensor values.

For example, outputs such as:

```
(None, tensor)
{"loss": None, "hidden_states": tensor}
```

can fail because the device of the first element is used as the reference device. Empty containers also fail due to the len(obj) > 0 assertions.

Would it be safer to recursively collect devices only from tensor leaves, ignore non-tensor/empty values, and then verify that all discovered tensor devices are identical?

That would also make the behavior independent of the ordering of fields in a tuple or dictionary.

dvsav: 👍 Absolutely. Implemented.

tico/quantization/config/gptq.py
dvsav: 👍  Good point. I've added `ignored_module_patterns` parameter to `wrap_model` function that contains a collection of module names to be igrnored by GPTQ quantization. If `gptq_config.quantize_lm_head` is `False`, `"lm_head.*"` pattern is passed through that parameter and thus, **lm_head** gets ignored by the quantizer.

I can also add a new property `ignored_module_patterns` to `UniversalGPTQConfig` class if we may want to not quantize arbitrary submodules (by name), not just `lm_head`. What do you think?

mhs4670go: Thanks for the suggestion. For now, I’d prefer to keep the public configuration limited to the existing quantize_* options and defer adding ignored_module_patterns until we have a concrete use case.

A general pattern-based exclusion API would require us to define matching semantics, subtree behavior, and precedence with the existing options. I think that would add more interface and implementation complexity than we need for this PR.

Using an internal helper is fine if it simplifies the implementation, but I don’t think we need to expose it as a configurable option yet. Let’s focus on making the existing quantize_lm_head option work correctly and covering its behavior with tests. Thank you!

tico/quantization/algorithm/universal_gptq/quantizer.py
mhs4670go, line 1: Please add a license header.

test/quantization/algorithm/test_universal_gptq.py
```diff
@@ -0,0 +1,662 @@
+# Copyright (c) 2025 Samsung Electronics Co., Ltd. All Rights Reserved
```
mhs4670go:
```suggestion
# Copyright (c) 2026 Samsung Electronics Co., Ltd. All Rights Reserved
```

tico/quantization/algorithm/universal_gptq/quantizer.py
mhs4670go, line 1: I think the multiple-invocation case is not handled correctly.

invocation_idx tracks accesses to the output cache, but the cache itself does not distinguish different call sites of the same module within a single forward.

For example:

```python
def forward(self, x):
    x = self.shared(x)  # invocation 0
    x = self.shared(x)  # invocation 1
    return self.tail(x)
```

During COLLECT, the first call to shared raises StopForward, so invocation 1 is never reached.

The same happens during CACHE: only the output of invocation 0 is computed and cached before StopForward is raised.

Once shared transitions to RETURN_CACHED, both invocation 0 and invocation 1 return entries from that cache. However, the cache contains outputs collected from invocation 0 across calibration batches, not outputs for the two distinct invocations.

For example, if shared(x) = 2x and the input is [1, 2], tail should receive [4, 8], but the current replay can feed it the cached output of the first invocation instead.

Of course I don’t think we necessarily need to support reused modules in the first version. It would be fine to define the supported execution contract as one invocation per quantizable module per forward and fail explicitly when that assumption is violated.

Otherwise, supporting this case seems to require cache identity that includes the call site/invocation within each batch, rather than a single per-module invocation_idx.

test/quantization/algorithm/test_universal_gptq.py
dvsav: 👍 Fixed

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: 👍 Yes, that's a very apt observation. I guess the problem can also be fixed by redesigning the cache structure from [batch_idx --> output] mapping to [input --> output] mapping. If the required input is found in the cache, we return the respective output, otherwise we compute the output and cache both input and output. The cached input will serve as the call site/invocation identity you've mentioned. Yet, that may double memory consumption, so I'm going to try to explore the issue in greater depth.

mhs4670go: @dvsav 

Thanks for looking into this. Input-based caching could address the output replay issue, but I think we should distinguish that from the calibration and quantization scheduling issue.

For example:

```python
def forward(self, x):
    x = self.shared(x)  # First invocation
    x = self.shared(x)  # Second invocation, using the same weight
    return self.tail(x)
```

With the current COLLECT behavior, the first invocation raises StopForward, so the second invocation is never reached during collection. finish_collection() then quantizes the shared weight using only the first invocation’s inputs.

Computing the second invocation on a later cache miss could fix the output passed to tail, but it would not retroactively include that invocation’s inputs in the Hessian used for the already-completed quantization. Supporting reused modules therefore requires a collection and quantization policy, not just a different cache key.

That being said, again, supporting this can be TODO for now as tackling this issue seems to take times and have low priority. 

For this PR, I would prefer to explicitly require each intercepted module to be called at most once per forward and fail clearly when that assumption is violated, rather than expand the cache design now. The validation needs to observe the complete execution trace, or use equivalent execution tracking; a check only inside COLLECT would miss the second call because of early termination.

Reused-module support could then be added separately, with explicit policies for collecting inputs across all uses, deciding when to quantize the shared weight, and invalidating cached outputs when necessary.

tico/quantization/algorithm/universal_gptq/quantizer.py
```diff
@@ -0,0 +1,838 @@
...
+        }
+        if devices:
+            assert (
+                len(devices) == 1
+            ), f"The object is assumed to be located on a single device. Got {devices}."
+            return next(iter(devices))
+        else:
+            return None
+
+    if isinstance(obj, Iterable):
```
mhs4670go, line 142: (optional) A non-empty string is iterable, and iterating over a one-character string produces that same string again. We can treat strings as non-tensor leaf values before entering the `Iterable` branch.

```python
if isinstance(obj, (str, bytes, bytearray)):
    return None
```

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: ## Can a module invoked multiple times be GPTQ quantized?

GPTQ algorithm assumes that the module currently being quantized receives its inputs from already quantized modules.
With a multiply-invoked module the module can receive inputs that depend on the previous invocation of that same module like in the example above. Let's consider another example with anotehr module invocation in the middle:
```python
def forward(self, x):
    x = self.multiply_invoked_module(x)  # invocation 0
    x = self.other_module(x)
    x = self.multiply_invoked_module(x)  # invocation 1
    return self.tail(x)
```
If we quantize the weights of `multiply_invoked_module` after `invocation 0` the `self.other_module` will receive the input from the already quantized `multiply_invoked_module` (which complies with GPTQ algorithm). But in that case the `invocation 1` of `multiply_invoked_module` will not participate in GPTQ quantization at all.

Otherwise, if we don't quantize the weights of `multiply_invoked_module` after `invocation 0` and wait until `invocation 1` collects its inputs, then `self.other_module` will be invoked with inputs from yet unquantized `self.multiply_invoked_module` (invocation 0) which violates the GPTQ algorithm assumption "each subsequent module receives inputs from already quantized modules".

This looks like a vicious circle inherent to GPTQ.
I can see 3 options here:
1. Detect multiply-invoked module beforehand and exclude them from GPTQ quantization.
2. Quantize the module after its **1st invocation**, accept that subsequent invocations reuse the same quantized weights without independent Hessian accumulation.
3. Treat each invocation site as a **separate quantization target** with independent weights and GPTQ solvers.

Option 1 seems to be the simplest.
Option 2 looks implementable.
Option 3 looks rather nontrivial and hard to implement.

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: 👍 Fixed

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: Implemented a solution to the "multi-call modules" issue. Here's the summary.

# Support for Multiply-Invoked Modules in `UniversalGPTQQuantizer`

## Problem

The original UniversalGPTQQuantizer implementation assumed each module is invoked exactly once per forward pass. This assumption breaks for modules that are called multiple times within a single forward, such as:

- **Shared weights**: `y = self.M(x); z = self.M(x)`
- **Circular dependencies**: `x = self.M(x); x = self.other(x); x = self.M(x)`
- **Recurrent structures**: Any module appearing multiple times in the computation graph

Such modules violate GPTQ's core assumption that "each module receives inputs from already-quantized predecessors," creating a circular dependency that cannot be resolved by layer-by-layer quantization.

## Solution

### 1. Detection and Exclusion of Multiply-Invoked Modules

**New function: `find_multiply_invoked_modules()`**

Before GPTQ quantization begins, we detect modules that are invoked multiple times per batch by:
- Temporarily wrapping each module's `forward` method to count invocations
- Running the model on calibration data
- Marking modules with `invocation_count > 1` in any batch
- Restoring original forward methods

Modules identified as multiply-invoked are **excluded from GPTQ** by passing them via the new `ignored_modules` parameter to `wrap_model()`. These modules will be quantized by the subsequent PTQ stage instead.

```python
ignored_modules=find_multiply_invoked_modules(
    model,
    args_dataset=args_dataset,
    kwargs_dataset=kwargs_dataset,
    show_progress=gptq_config.show_progress,
)
```

### 2. Merged CACHE/RETURN_CACHED States

**State machine simplification: 4 states → 3 states**

The original implementation had separate `CACHE` and `RETURN_CACHED` states, transitioning after the first calibration batch. This doesn't work for multiply-invoked modules because we don't know the total invocation count in advance.

**New behavior:** The `CACHE` state now handles both roles dynamically:

```python
case GPTQ_STATE.CACHE:
    if gptq_data.invocation_idx < len(gptq_data.cached_output):
        # Return cached output (previously RETURN_CACHED behavior)
        result = gptq_data.cached_output[gptq_data.invocation_idx]
        gptq_data.invocation_idx += 1
        return result
    else:
        # Compute and cache new output (previously CACHE behavior)
        out = gptq_data.old_forward(*args, **kwargs)
        gptq_data.cached_output.append(out)
        gptq_data.invocation_idx += 1
        raise StopForward(module)
```

This allows modules to:
- Return cached outputs for known invocation indices
- Dynamically extend the cache for new invocation sites
- Handle any number of invocations without prior knowledge

### 3. Invocation Counter Reset

**New function: `reset_invocation_counter()`**

After each calibration batch, we reset `invocation_idx` for all modules to ensure cached output indices align with the next batch's invocation sequence:

```python
def reset_invocation_counter(module: nn.Module) -> None:
    """Reset invocation_idx for all submodules after each batch."""
    for m in module.modules():
        gptq_data: GPTQ_Data = get_gptq_data(m)
        gptq_data.invocation_idx = 0
```

This is called in `gptq_quantize()` after processing each batch:

```python
for args, kwargs in zip(...):
    frontier_submodules = run_model(...)
    reset_invocation_counter(model)  # ← Reset for next batch
```

### 4. Cache Optimization Retained

**Memory efficiency: Children's cached outputs are cleared**

When a parent module finishes caching for a call site, its children's cached outputs are cleared (memory optimization). However, children **do not transition to `COMPUTE`** because they may still need to cache outputs for subsequent invocations from the parent.

```python
def finish_caching(module: nn.Module) -> None:
    """Complete caching for a call site, clear children's cache."""
    gptq_data: GPTQ_Data = get_gptq_data(module)
    
    # Free children's cached outputs (memory optimization)
    for child in module.children():
        child_gptq_data = get_gptq_data(child)
        child_gptq_data.cached_output.clear()
        # Note: child stays in CACHE state, not COMPUTE
```

The `COMPUTE` state is retained but currently unused—reserved for potential future optimizations.

## Changes Summary

| File | Changes |
|------|---------|
| `tico/quantization/algorithm/universal_gptq/quantizer.py` | - Added `find_multiply_invoked_modules()` detection<br>- Added `reset_invocation_counter()`<br>- Merged `RETURN_CACHED` into `CACHE` state<br>- Added `ignored_modules` parameter to `wrap_model()`<br>- Updated state machine documentation |
| `test/quantization/algorithm/test_universal_gptq.py` | - 5 new tests for multiply-invoked module edge cases |
| `.gitignore` | Added local development directories |

## Test Coverage

Five new tests verify correct handling of multiply-invoked modules:

1. **`test_universal_gptq_module_called_twice_same_input`** — Non-problematic double calls (same input)
2. **`test_universal_gptq_module_called_twice_chained`** — Circular dependency (chained calls)
3. **`test_universal_gptq_module_called_once_per_batch`** — Single-call modules (correctly NOT excluded)
4. **`test_universal_gptq_nested_multiply_invoked_modules`** — Nested multiply-invoked structures
5. **`test_universal_gptq_multiply_invoked_output_correctness`** — Output correctness with known transforms

All tests pass with the new implementation.

## Impact

- **Backward compatible**: Existing models with single-invocation modules work unchanged
- **Enables new models**: Models with shared weights, recurrent structures, or circular dependencies can now be quantized
- **Graceful degradation**: Multiply-invoked modules fall back to PTQ quantization (slightly less optimal but functional)

tico/quantization/algorithm/universal_gptq/quantizer.py
```diff
@@ -0,0 +1,956 @@
...
+        kwargs_dataset: List of cached keyword arguments from calibration.
+
+    Raises:
+        AssertionError: If the calibration dataset is empty or mismatched.
+    """
+    assert len(args_dataset) > 0, "Empty calibration dataset"
+    assert len(args_dataset) == len(kwargs_dataset), "Dataset length mismatch"
+
+    if not has_gptq_data(model):
+        wrap_model(
```
mhs4670go: Seems that the detected modules are not actually excluded from GPTQ for normal nested models.

`ignored_modules` is passed to the top-level `wrap_model()` call, but it is not forwarded by the recursive call:

```python
wrap_model(
    child,
    full_model_name=full_child_name,
    ignored_module_patterns=ignored_module_patterns,
)

As a result, a multiply-invoked child `nn.Linear` is wrapped again with the default empty `ignored_modules` collection and remains in COLLECT.

File: tico/quantization/algorithm/universal_gptq/quantizer.py [hunk 7]
```diff
@@ -0,0 +1,956 @@
...
+        invocation_idx: Current invocation index during replay.
+        weight_device: Device where the module's parameters reside.
+        out_device: Device where module output should reside.
+    """
+
+    full_module_name: str
+    old_forward: Callable
+    gptq: GPTQ | None
+    quantizer: Quantizer | None
+    cached_output: list[Any]
```
mhs4670go: I think there is still a correctness issue in the new cache/replay logic: the current implementation does not preserve calibration-batch identity.

There seem to be two related consequences:

1. cached_output conflates the batch axis with the invocation/call-site axis.
2. frontier_submodules loses information about which calibration batches actually reached each frontier module.

### 1. Cache identity needs both batch and invocation

Consider a multiply-invoked module:

```python
def forward(self, x):
    y = self.shared(x)  # invocation 0
    z = self.shared(x)  # invocation 1
    return self.tail(torch.cat([y, z], dim=-1))
```

Suppose:

```
shared(x) = 2x
batch 0: x = 1
batch 1: x = 10
```

The correct outputs of shared are:

```
batch 0:
    invocation 0 -> 2
    invocation 1 -> 2
batch 1:
    invocation 0 -> 20
    invocation 1 -> 20
```

Therefore, the cache logically needs to distinguish at least:

```
(batch_idx, invocation_idx)
```

However, the current implementation uses a single flat list:

```
cached_output[invocation_idx]
```

During an early caching pass, the cache can effectively become:

```
cached_output[0] = batch 0, invocation 0 -> 2
cached_output[1] = batch 1, invocation 0 -> 20
```

When batch 0 is replayed and shared is invoked twice, the two calls can then consume:

```
invocation 0 -> cached_output[0] -> 2
invocation 1 -> cached_output[1] -> 20
```

so tail receives:

```
[2, 20]
```

instead of the correct:

```
[2, 2]
```

Also, despite the comment saying that invocation_idx is reset after each calibration batch, the current code calls:

```
frontier_submodules = run_model(...)
reset_invocation_counter(model)
```

while run_model() itself iterates over the entire calibration dataset. So the reset happens once per replay pass, not once per calibration batch.

Simply moving the reset inside the batch loop would not be sufficient either, because then every batch could start reading from the same cached_output[0]. I think the cache itself needs explicit batch identity, conceptually something like:

```
cached_output[batch_idx][invocation_idx]
```

or an equivalent representation.

### 2. Per-batch frontier information is also lost

The same issue affects the decision of when a module has collected enough Hessian statistics to be quantized.

Suppose there are three calibration batches and, during one replay iteration, they stop at different modules:

```
batch 0 -> reaches B -> StopForward(B)
batch 1 -> reaches A -> StopForward(A)
batch 2 -> reaches A -> StopForward(A)
```

This can happen with the dynamic cache behavior when one batch gets a cache hit and proceeds further while another batch gets a cache miss and stops earlier.

run_model() currently collapses this into:

```
frontier_submodules = {A, B}
```

The information that only batch 0 reached B is lost.

Then gptq_quantize() processes every module in the set:

```python
for frontier_submodule in frontier_submodules:
    if gptq_data.state == GPTQ_STATE.COLLECT:
        finish_collection(frontier_submodule, gptq_config)
```

so finish_collection(B) may be called even though B has accumulated Hessian statistics from only one of the three calibration batches.

Conceptually, instead of:

```
H_B = H(batch 0) + H(batch 1) + H(batch 2)
```

we could quantize B with only:

```
H_B = H(batch 0)
```

because batches 1 and 2 never reached it.

The problem is not that frontier_submodules is a set by itself. The problem is that a set[nn.Module] no longer contains enough information once different calibration batches can make different amounts of progress.

Previously, if all batches stopped at the same frontier:

```
batch 0 -> A
batch 1 -> A
batch 2 -> A
```

then:

```
frontier_submodules = {A}
```

implicitly meant that all calibration batches had contributed to A.

With the new dynamic cache behavior, that assumption no longer seems valid.

I think replay now needs to preserve batch identity in both places, for example conceptually:

```
cache[batch_idx][invocation_idx]
```

and:

```
frontier_by_batch = {
    0: B,
    1: A,
    2: A,
}
```

or equivalently:

```
reached_batches = {
    A: {1, 2},
    B: {0},
}
```

A module should only be considered ready for finish_collection() once all required calibration batches have reached and contributed to it.

So I think these are two manifestations of the same underlying issue: batch identity is currently lost during replay. Fixing only invocation_idx or only the frontier set would not be sufficient; cache indexing and frontier/completion tracking need to use a consistent per-batch execution context.

tico/quantization/algorithm/universal_gptq/quantizer.py
```diff
@@ -0,0 +1,956 @@
...
+            AssertionError: If prepare() is called twice without convert().
+        """
+        assert len(self._cache_args) == 0, "prepare() called twice without convert()"
+        # Substitute the model's forward method to cache model inputs
+        def new_forward(model: nn.Module, *args, **kwargs) -> Any:
+            """
+            Stores this batch's inputs and kwargs, then raises StopForward to stop computation.
+            """
+            self._cache_args.append(tuple(move_to_cpu(arg) for arg in args))
+            self._cache_kwargs.append({k: move_to_cpu(v) for k, v in kwargs.items()})
```
mhs4670go: `prepare()` stores all calibration tensors on CPU via `move_to_cpu()`. However, `find_multiply_invoked_modules()` directly executes:

```python
model(*args, **kwargs)
```

without moving the cached arguments back to the model device.

The normal run_model() path does perform a device transfer before execution, but the new detection path does not. Therefore, a typical CUDA model will receive CPU calibration tensors during this preliminary trace and fail with a device mismatch. Am I missing something?

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: Let me first respond to **1. Cache identity needs both batch and invocation**.
Below is the summary of changes in the last commit that addresses the issue.

## Summary of Changes

### 1. **Cache Structure: From 1D to 2D**

**Before:**
```python
cached_output: list[Any]  # cached_output[invocation_idx]
```

**After:**
```python
cached_output: list[list[Any]]  # cached_output[batch_idx][invocation_idx]
```

This is the core fix: each batch now has its own dedicated list of cached outputs, preventing cross-batch contamination.

### 2. **New `batch_idx` Field in GPTQ_Data**

```python
@dataclass
class GPTQ_Data:
    cached_output: list[list[Any]]
    invocation_idx: int
    batch_idx: int  # ← NEW: tracks current batch during replay
```

### 3. **Updated CACHE State Logic**

The `wrap_model` function's CACHE state now:

```python
case GPTQ_STATE.CACHE:
    # Create new batch entry if needed
    if gptq_data.batch_idx >= n_cached_batches:
        gptq_data.cached_output.append([])
    
    cached_batch_out = gptq_data.cached_output[gptq_data.batch_idx]
    
    # Return cached output if available for THIS batch
    if gptq_data.invocation_idx < len(cached_batch_out):
        result = cached_batch_out[gptq_data.invocation_idx]
        gptq_data.invocation_idx += 1
        return result
    else:
        # Compute and cache for THIS batch
        out = gptq_data.old_forward(*args, **kwargs)
        cached_batch_out.append(move_to_cpu(out))
        gptq_data.invocation_idx += 1
        raise StopForward(module)
```

### 4. **Batch Counter Management**

**New functions:**
- `reset_batch_counter(module)`: Resets `batch_idx` to 0 for all submodules
- `increment_batch_counter(module)`: Increments `batch_idx` for all submodules

**Updated `run_model()` flow:**
```python
for args, kwargs in tqdm(...):  # For each batch
    try:
        model(*args, **kwargs)
    except StopForward as stop_fwd:
        frontier_submodules.add(stop_fwd.module)
    finally:
        reset_invocation_counter(model)    # ← Reset per-batch
        increment_batch_counter(model)     # ← Increment per-batch

reset_batch_counter(model)  # ← Reset after all batches
```

---

## How It Solves the Example

**Scenario from reviewer:**
```python
def forward(self, x):
    y = self.shared(x)  # invocation 0
    z = self.shared(x)  # invocation 1
    return self.tail(torch.cat([y, z], dim=-1))

# Batch 0: x = 1 → shared(x) = 2
# Batch 1: x = 10 → shared(x) = 20
```

**Before (buggy):**
```
cached_output = [2, 20]  # Mixed batches!

Replay batch 0:
  invocation 0 → cached_output[0] = 2  ✓
  invocation 1 → cached_output[1] = 20 ✗ (should be 2!)
```

**After (fixed):**
```
cached_output = [
    [2, 2],    # batch 0: both invocations → 2
    [20, 20],  # batch 1: both invocations → 20
]

Replay batch 0:
  batch_idx = 0, invocation_idx = 0 → cached_output[0][0] = 2  ✓
  batch_idx = 0, invocation_idx = 1 → cached_output[0][1] = 2  ✓

Replay batch 1:
  batch_idx = 1, invocation_idx = 0 → cached_output[1][0] = 20  ✓
  batch_idx = 1, invocation_idx = 1 → cached_output[1][1] = 20  ✓
```

---

## Test Validation

All 10 existing tests pass, including the 5 new tests specifically for multiply-invoked modules:

```
Ran 10 tests in 0.545s
OK
```

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: 👍 Fixed

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: 👍 Fixed

tico/quantization/algorithm/universal_gptq/quantizer.py
dvsav: Now, to the **2. Per-batch frontier information is also lost** section.

Let me try to imagine several computation graph patterns and analyze hypothetical issues with them.

**1. Two sequential modules**
```
... --> (A) --> (B) --> ...
```

Obviously, no batch can reach `(B)` before `(A)` gets fully cached (for all batches).
*All* batches reach `(A)` first, then *all* batches reach `(B)`.
No issues.

**2. Data-dependent computation path**
```
             .------------------------------> (A) --> ...
            /
           / condition(input) == True
          /
input ---+
          \
           \ condition(input) == False
            \
             `------------------------------> (B) --> ...
```
It depends on specific batch whether we reach `(A)` or `(B)` first:
```
batch 0 -> reaches B -> StopForward(B)
batch 1 -> reaches A -> StopForward(A)
```
Yet, *all* batches that reach `(A)` are collected by A's Hessian in a single dataset replay.
And similarly *all* batches that reach `(B)` are collected by B's Hessian in the same dataset replay.
Albeit the number of batches reaching `(A)` and `(B)` can be different and smaller than the total number of batches in the dataset.
No issues here.

**3. Multi-call module in a single batch**

```
... --> (A) --> (B) --> (A) --> ...
```
All batches reach `(A)` first, but the number of actual inputs to `(A)` is twice as large as the number of batches because `(A)` is invoked twice per a single batch.
If we quantize `(A)` first, we won't collect the other half of A's inputs.
If we quantize `(B)` first, we'll need to let `(B)` collect the outputs of yet unquantized `(A)`, which violates the spirit of GPTQ algorithm.
That's a vicious circle. But we can detect multi-call modules like `(A)` in advance by replaying unquantized model and counting the number of each module invocations per batch.
Then we can exclude those multi-call modules from GPTQ quantization altogether.

**4. Data-dependent computation path + multi-call module across different batches**

```
             .------------------------------> (A) --> ...
            /
           / condition(input) == True
          /
input ---+
          \
           \ condition(input) == False
            \
             `------------------------------> (B) --> (A) --> ...
```
If we quantize `(A)` first, we'll collect only one half of the A's inputs.
So, we need to collect the inputs for `(B)` first. But we don't have the above computation graph at hand, so we generally don't know which module to quantize first.
We want to quantize the module that has collected all batches that can potentially reach it.
We may assume that **the number of batches reaching each module stays the same for quantized and unquantized model**.
Then we may count the number of batches reaching each module in a single unquantized model replay.
Then, during the quantization process, after each model replay we may quantize only the modules that have collected the required number of batches (there should be at least 1 such module after each replay, otherwise there's a circular dependency - see the next case).

**5. Data-dependent computation path + multi-call modules across different batches**

```
             .------------------------------> (A) --> (B) --> ...
            /
           / condition(input) == True
          /
input ---+
          \
           \ condition(input) == False
            \
             `------------------------------> (B) --> (A) --> ...
```

If we quantize `(A)` first, we'll collect only one half of the A's inputs.
If we quantize `(B)` first, we'll collect only one half of the B's inputs.
We can't collect the required number of inputs for either module without compromising the requirement of GPTQ: "each module should collect the inputs from the already quantized modules".
Here we may first quantize the module that has collected the most inputs.

mhs4670go (approved):
LGTM. Thank you!

dvsav:
Qwen3-VL quantization completed, except for the side-by-side comparison of quantized and unquantized models that failed because of out-of-memory exception (NVIDIA RTX A5000 RAM 23 GiB).

[qwen3_vl_model_agnostic_gptq_no_cache.log](https://github.com/user-attachments/files/32596302/qwen3_vl_model_agnostic_gptq_no_cache.log)

